PR #27, opened 2021-01-28
Add is_trace_preserving feature

### LINKED ISSUE #11: Feature: Is trace preserving
Given a channel specified by either its Choi matrix or its Kraus representation, determine if the channel is trace-preserving.

For the definition of what constitutes a channel to be trace-preserving, refer to Section 2.2.1 of https://cs.uwaterloo.ca/~watrous/TQI/TQI.pdf

This would involve creating `channel_props/is_trace_preserving.py` and adding the logic there. Refer to other files in the same directory, namely `is_completely_positive.py`, `is_herm_preserving.py`, etc. for ensuring consistency amongst the other related functions. 

The function prototype would follow this form:

```
def is_trace_preserving(
    phi: Union[np.ndarray, List[List[np.ndarray]]],
    rtol: float = 1e-05,
    atol: float = 1e-08,
) -> bool:
...
```

This task would also require adding test coverage for the function along with ensuring it is included in the docs. Refer to the style guide and code contributing guidelines for more information.

georgios-ts:
## Description
closes #11 .
Add `is_trace_preserving` function. Tests and docs included.

## Status
-  [x] Ready to go

### DIFF
--- a/toqito/channel_props/is_trace_preserving.py
+++ b/toqito/channel_props/is_trace_preserving.py
@@ -1,4 +1,4 @@
-"""Is channel Hermiticity-preserving."""
+"""Is channel Trace-preserving."""
 from typing import List, Union
 
 import numpy as np
--- a/toqito/channel_props/is_trace_preserving.py
+++ b/toqito/channel_props/is_trace_preserving.py
@@ -1,0 +1,98 @@
+"""Is channel Hermiticity-preserving."""
+from typing import List, Union
+
+import numpy as np
+
+from toqito.matrix_props import is_identity
+from toqito.channels import partial_trace
+
+
+def is_trace_preserving(
+    phi: Union[np.ndarray, List[List[np.ndarray]]],
+    rtol: float = 1e-05,
+    atol: float = 1e-08,
+    dim: Union[List[int], np.ndarray] = None,
+) -> bool:
+    r"""
+    Determine whether the given channel is Trace-preserving [WatH18]_.
+
+    A map :math:`\Phi \in \text{T} \left(\mathcal{X}, \mathcal{Y} \right)` is
+    *Trace-preserving* if it holds that
+
+    .. math::
+        \text{Tr} \left( \Phi(X) \right) = \text{Tr}\left( X \right)
+
+    for every operator :math:`X \in \text{L}(\mathcal{X})`.
+
+    Given the corresponding Choi matrix of the channel, a neccessary and sufficient condition is
+
+    .. math::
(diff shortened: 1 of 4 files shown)


toqito/channel_props/is_trace_preserving.py
```diff
@@ -0,0 +1,98 @@
...
+    if isinstance(phi, list):
+        phi_l = [A for A, _ in phi]
+        phi_r = [B for _, B in phi]
+
+        k_l = np.concatenate(phi_l, axis=0)
+        k_r = np.concatenate(phi_r, axis=0)
+
+        mat = k_l.conj().T @ k_r
+    else:
+        mat = partial_trace(input_mat=phi, sys=2, dim=dim)
```
vprusso, line 97: I wonder if we should expect an argument for `sys` as we do for `dim` as input to the `is_trace_preserving` function.

toqito/channel_props/is_trace_preserving.py
```diff
@@ -0,0 +1,98 @@
...
+    :param phi: The channel provided as either a Choi matrix or a list of Kraus operators.
+    :param rtol: The relative tolerance parameter (default 1e-05).
+    :param atol: The absolute tolerance parameter (default 1e-08).
+    :param dim: Dimension of the subsystems. If :code:`None`, all dimensions are assumed to be
+                equal.
+    :return: True if the channel is Trace-preserving, and False otherwise.
+    """
+    # If the variable `phi` is provided as a list, we assume this is a list
+    # of Kraus operators.
+    if isinstance(phi, list):
```
vprusso, line 88: Instead of converting this way, we could make use of the existing `kraus_to_choi` function found in `channel_ops/kraus_to_choi.py`.

toqito/channel_props/is_trace_preserving.py
```diff
@@ -0,0 +1,98 @@
+"""Is channel Trace-preserving."""
```
vprusso, line 1: The word `Trace` is not a proper noun, so I would suggest `Trace-preserving` should be changed to `trace-preserving`.

vprusso (changes requested):

toqito/channel_props/is_trace_preserving.py
georgios-ts: I was thinking this way might be faster. Should i use `kraus_to_choi` for simplicity instead?

vprusso:
Awesome. Thanks yet again for contributing and improving the toqito project! :)

PR #57, opened 2021-05-09
Affiliation fix.

arfon:
EOM


vprusso:
Thank you, @arfon! Merged. Let me know if there is anything else, and thanks again!

PR #64, opened 2021-05-20
Fix implementation of `random_povm`

theRoughCode:
## Description
I noticed a bug in the code of `random_povm` in which running `random_povm(dim, num_inputs, num_outputs)`, where `dim != num_inputs` and `dim != num_outputs` throws an error. It turns out that this is a one-line fix.

I also touched up the docs to state more clearly what the output should look like.

## Status
-  [x] Ready to go


vprusso:
Nice catch, @theRoughCode! Thank you for making the fix and for your attention to detail!

Issue #44, opened 2021-03-05
Feature: Convert description of a binary constraint game to a nonlocal game


vprusso:
Binary constraint games (sometimes called linear system games or binary constraint system games) are a subset of the class of nonlocal games that are related to binary constraint systems. We shall refer to these as BCS games for short. The formulation of these games can be found in [arXiv:1209.2729](https://arxiv.org/abs/1209.2729).

BCS games have served as useful tools for the study of nonlocal games and have also been used as a mechanism to disprove one variant of [Tsirelson's problem](https://uwaterloo.ca/institute-for-quantum-computing/news/solving-tsirelsons-problem) as shown in [arXiv:1703.08618](https://arxiv.org/abs/1703.08618). 

This task pertains to writing a function that takes a description of a BCS game and converting that to a description of a nonlocal game. The purpose of this task is to allow us to calculate various quantities of the game in question (including the classical, quantum, and non-signaling values). 

A BCS game is specified by a list of constraints that define the game. Specifically, it is defined by a list of `n` 2-dimensional binary `numpy` arrays where the entry of the array is equal to 1 if and only if the values of the variables for that constraint are satisfied. For concreteness, consult the example provided in [arXiv:1209.2729](https://arxiv.org/pdf/1209.2729.pdf) on the first page.

A nonlocal game is specified by variables `prob_mat` and `pred_mat`, where

```
        :param prob_mat: A matrix whose (x, y)-entry gives the probability
                        that the referee will give Alice the value `x` and Bob
                        the value `y`.
        :param pred_mat: A four-dimensional matrix whose (a,b,x,y)-entry gives
                         the outcome for answers "a" and "b" given questions
                         "x" and "y".
```

The goal of this task is to convert the specification of a BCS game and convert it to represent the `prob_mat` and `pred_mat` variables to define a nonlocal game.

theRoughCode:
Hey @vprusso, I'm interested in tackling this feature. However, I'm not entirely clear about the definition of a BCS game as defined as `numpy` arrays. in the first part of the definition, we define a BCS game as defined by a list of constraints. My initial intuition was to represent this as an `m x (n+1)` array where the `(i,j)`-th element is 1 if `v_j` is in `c_i`, for `0 <= j < n`, and `(i, n)` is the result of `c_i`.

However, the second half of the definition above seemed to use a solution to the constraints as the definition:
> where the entry of the array is equal to 1 if and only if the values of the variables for that constraint are satisfied

[...]

PR #65, opened 2021-05-20
Fix implementation of computing the quantum value lower bound of a non-local game

theRoughCode:
## Description
When testing my implementation of #62 by trying to calculate the quantum value lower bound of the CHSH game, I got the following error:
```python
ValueError: Invalid dimensions (4, 2). Must be a square matrix.

toqito\nonlocal_games\nonlocal_game.py:361: in quantum_value_lower_bound
    alice_povms, lower_bound = self.__optimize_alice(bob_povms)
toqito\nonlocal_games\nonlocal_game.py:392: in __optimize_alice
    alice_povms[x_ques, a_ans] = cvxpy.Variable(
..\venv\lib\site-packages\cvxpy\expressions\variable.py:81: in __init__
    super(Variable, self).__init__(shape, **kwargs)
```

This is because the current SDP algorithm requires a square matrix  if `PSD` is true, but in the case of the BCS representation of the CHSH game, `num_output_alice = 4` and `num_inputs_alice = 2`.  More info can be found [here](https://github.com/vprusso/toqito/issues/44#issuecomment-843391727).

This feature fixes this by following what was done in [QETLAB](https://github.com/nathanieljohnston/QETLAB/blob/702e6230a62c2241452775ceb88e63630bfd679e/NonlocalGameLB.m) more closely and optimizing over the shared `d`-dimensional quantum state `tau`. We also allow the user to specify what `d` is in the arguments to `NonlocalGame.quantum_value_lower_bound`.

## Status
-  [ ] Ready to go (depends on the fix in #64 to be merged in first)

### DIFF
--- a/tests/test_nonlocal_games/test_nonlocal_game.py
+++ b/tests/test_nonlocal_games/test_nonlocal_game.py
@@ -52,6 +52,10 @@ class TestNonlocalGame(unittest.TestCase):
         # the true quantum value.
         self.assertLessEqual(np.isclose(res, np.cos(np.pi / 8) ** 2, rtol=1e-02), True)
 
+        # Even with 2 qubits each, the lower bound remains the same
+        res = chsh.quantum_value_lower_bound(4)
+        self.assertLessEqual(np.isclose(res, np.cos(np.pi / 8) ** 2, rtol=1e-02), True)
+
     def test_chsh_game_classical_value(self):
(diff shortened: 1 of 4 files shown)


vprusso:
Thanks for this PR as well, @theRoughCode. As #64 has been merged in, per your request, I'll be merging this one in now. Thanks again for the PR!

Issue #28, opened 2021-01-28
Feature: Matsumoto fidelity


vprusso:
The Matusmoto fidelity was introduced in arXiv:1006.0302. In arXiv:2006.06918, it was shown that it is possible to compute this quantity by means of a semidefinite program. Refer to equation 1.11 in the aforementioned preprint. 

The functionality for this feature should be included under `state_metrics/matumoto_fidelity.py`. To solve the semidefinite program, it will most likely make sense to make use of the `cvxpy` module that is used elsewhere in `toqito` to solve similar semidefinite programs. For instance, consult the `state_metrics/fidelity.py` file to see how this function operates, how it invokes the SDP solver, etc. 

Ensure proper testing is in place and 100% test coverage is obtained by creating a unit test file in `tests/test_state_metrics/test_matusmoto_fidelity.py`. Ensure to add `toqito.state_metrics.matusmoto_fidelity` under "Distance Metrics for Quantum States" in `/docs/states.rst`

dfacoet:
I would like to do this. 

It should be pretty similar to `fidelity`, the main differences being that the variable in the SDP is Hermitian, and that the direct calculation can only be used if at least one of the density matrices is non-singular.
Does this sound right?

vprusso:
Hi @dfacoet. That's great to hear, thank you for taking an interest in this feature!

And yes, I believe your initial assessment of this task seems pretty spot on. For what it's worth, it may also be worth comparing with another implementation of this metric as implemented by [QETLAB](http://www.qetlab.com/MatsumotoFidelity). If you have access to MATLAB, you can check your implementation against the QETLAB one, and even if you don't, I think the code itself is well-written and should serve as a guidepost for how to implement it. 

Hope that helps, and please do not hesitate to ping me if you have any questions. Thanks again for your interest, and looking forward to seeing what you come up with!

vprusso:
Closing as this has been implemented. Thank you, @dfacoet!

PR #71, opened 2021-06-04
Adding matsumoto_fidelity feature

dfacoet:
## Description
This implements `matsumoto_fidelity` as described in #28, including documentation and tests.

The implementation is similar to that of `fidelity`, the main difference being that the direct calculation requires at least one of the density matrices to be invertible.
To work around this I adapted the QETLAB implementation: try to invert the matrix with largest determinant; if an error is raised, add a small number `eps` to it. Initially I tried to use machine epsilon, but it created stability problems with the square root. After some trial and error, I found `1e-8` to be a reasonable value (this is also what QETLAB uses).
This works because for singular matrices the limit of zero `eps` is the correct definition of the Matsumoto fidelity (eq. 1.6 in https://arxiv.org/abs/2006.06918). The QETLAB documentation states that the result is accurate to `1e-4`, which is consistent with what I see.

A possible improvement could be to check whether the matrices are singular (perhaps with `toqito.matrix_props.is_positive_definite`) and use the `cvxpy` solution if they are. This could be given as an option, to avoid introducing a `cvxpy` dependence.

Regarding the tests, I reproduced those of `fidelity`. One thing to note is that they are all for either equal states, or classical mixtures. In particular, `matsumoto_fidelity` and `fidelity` have the same behaviour for all the tests.

## Questions
-  [ ] Should we use `cvxpy` if both matrices are singular?
-  [x] Should we add more tests?

### DIFF
--- a/docs/states.rst
+++ b/docs/states.rst
@@ -1,3 +1,4 @@
+
 States
 =====================
 
@@ -24,6 +25,7 @@ Distance Metrics for Quantum States
     toqito.state_metrics.trace_distance
     toqito.state_metrics.trace_norm
     toqito.state_metrics.bures_distance
+    toqito.state_metrics.matsumoto_fidelity
 
 Optimizations over Quantum States
 -----------------------------------------------
--- a/tests/test_state_metrics/test_matsumoto_fidelity.py
+++ b/tests/test_state_metrics/test_matsumoto_fidelity.py
@@ -1,4 +1,4 @@
-"""Tests for fidelity."""
+"""Tests for matsumoto_fidelity."""
 import cvxpy
 import numpy as np
 
--- a/tests/test_state_metrics/test_matsumoto_fidelity.py
+++ b/tests/test_state_metrics/test_matsumoto_fidelity.py
@@ -1,0 +1,61 @@
+"""Tests for fidelity."""
+import cvxpy
+import numpy as np
+
+
+from toqito.state_metrics import matsumoto_fidelity
+from toqito.states import basis
+
+
+def test_matsumoto_fidelity_default():
+    """Test Matsumoto fidelity default arguments."""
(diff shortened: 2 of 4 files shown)


vprusso:
Hi @dfacoet. Overall, looks pretty great!

A few comments: I also noticed that building the docs by navigating to the `toqito/docs` directory and running

```
make clean html
```

yields the following warning:

```
toqito.state_metrics.matsumoto_fidelity:58: WARNING: Explicit markup ends without a blank line; unexpected unindent.
```

W.r.t. your questions:

> Should we add more tests?

Looks like you achieved 100% coverage, which is great. More tests are always good, but at a minimum obtaining the 100% coverage is a great metric. It also looks like you have a pretty good set of tests, but if you notice any other edge cases, by all means, please do add them, but for now, it looks good. 

> Should we use `cvxpy` if both matrices are singular?

I could be missing something, but I would imagine so. Is there a reason why one would not want to do this?

vprusso:
@dfacoet just following up to see where you are in the updates to this PR. I think it's 99% ready to go, but let me know if you have any questions of course as well. Cheers!

dfacoet:
Hi @vprusso. Thank you for the review!

> yields the following warning:

This should be fixed by ae521707535e62a477a8a89ccbeaf0438ae26711.

> I could be missing something, but I would imagine so. Is there a reason why one would not want to do this

The way it is done at the moment (similar to QETLAB) is a bit blunt but reasonably accurate (~10^4) and avoids `cvxpy`, unless explicitly requested (if the input is a `cvxpy` variable).
I thought it could be desirable to know which method will be used, depending only on the input type, but I don't know if it is, which is why I asked.

A finer implementation could include an additional parameter such that if both determinants are smaller than the parameter, `cvxpy` is used. But I could be overthinking this for no good reason, and we can either leave it as it is, or use `cvxpy` whenever matrix inversion fails.

vprusso:
Hi @dfacoet. Gotcha, those are all good points. I think for the sake of the task at hand though, I'd say this feature looks good to me. I'll go ahead and merge it in. Thanks again for your contribution!

PR #75, opened 2021-06-10
Fix link in `depolarizing` docstring

dfacoet:
Fixed the link to https://en.wikipedia.org/wiki/Quantum_depolarizing_channel


vprusso:
Nice catch! :)

PR #93, opened 2022-06-06
Fix dimension handling in `symmetric_extension_hierarchy`.

georgios-ts:
## Description
Fixes a bug where `symmetric_extension_hierarchy` would fail:
```python
from toqito.states import werner
from toqito.state_opt import symmetric_extension_hierarchy

n = 5
states = [werner(n, -1.0), werner(n, 1.0)]
res = symmetric_extension_hierarchy(states=states, probs=None, level=1)
```
The source of the problem was a miscalculation of the dimensions of the two sub-systems that states live.
In addition, this PR introduces an extra optional arg `dim` which allow users to define the dimensions of the
sub-systems since we assume they're equal by default.

## Status
-  [x] Ready to go

### DIFF
--- a/tests/test_state_opt/test_symmetric_extension_hierarchy.py
+++ b/tests/test_state_opt/test_symmetric_extension_hierarchy.py
@@ -4,7 +4,7 @@ import pytest
 
 from toqito.perms import swap
 from toqito.state_opt import symmetric_extension_hierarchy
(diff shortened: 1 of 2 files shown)


vprusso:
Great to see you around here again, @georgios-ts ! Thank you for the fix--that's fantastic!

PR #96, opened 2022-06-10
is_separable function dev

meandmytram:
## Description
Hey @vprusso, sorry it took me a while, but here's my contribution to your implementation of `is_separable` function. The only thing is yet to figure out is on the lines of 152-156 of the `is_separable.py` file. I read the paper referenced there and it seems like one needs to solve an optimization problem there regarding the decomposition of the density matrix, but I haven't figured out how to do that.

## Criteria implemented

- [x] PPT criterion for prod(dims)<=6
- [ ] Operational criterion (not covered in #96)
- [x] Realignment criterion
- [x] Symmetric extension search criterion

### DIFF
--- a/docs/states.rst
+++ b/docs/states.rst
@@ -77,6 +77,7 @@ Properties of Quantum States
     toqito.state_props.negativity
     toqito.state_props.purity
     toqito.state_props.schmidt_rank
+    toqito.state_props.is_separable
     toqito.state_props.von_neumann_entropy
 
 Quantum States
--- a/tests/test_state_props/test_is_separable.py
+++ b/tests/test_state_props/test_is_separable.py
@@ -1,7 +1,7 @@
 """Test is_separable."""
 import numpy as np
 
-from toqito.state_props import is_separable
+from toqito.state_props.is_separable import is_separable
 from toqito.channels import partial_trace
 from toqito.matrix_props import is_density
 from toqito.states import basis, bell, isotropic, tile
@@ -46,14 +46,14 @@ def test_ppt_small_dimensions():
 
 
 def test_ppt_low_rank():
-    """Determined to be separable via sufficiency of the PPT criterion for low-rank operators."""
-    m = 5
+    """Determined to be separable via the operational criterion for low-rank operators."""
+    m = 6
     n = m
     rho = random_density_matrix(m)
     u, s, v_h = np.linalg.svd(rho)
     rho_cut = u[:, :m-1] @ np.diag(s[:m-1]) @ v_h[:m-1]
     rho_cut = rho_cut / np.trace(rho_cut)
-    pt_state_alice = partial_trace(rho_cut, 2, dim=m)
+    pt_state_alice = partial_trace(rho_cut, 2, dim=[3, 2])
 
     np.testing.assert_equal(is_density(rho_cut), True)
     np.testing.assert_equal(is_density(pt_state_alice), True)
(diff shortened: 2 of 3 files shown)


vprusso:
Hi @meandmytram 

Thank you for your time and effort on this function--that's great! 

It looks like some of the other separability tests that QETLAB includes are not presently covered in the diff. For instance, the tests beginning here:
https://github.com/nathanieljohnston/QETLAB/blob/21e3115f5676d8650cd1e94107a63772720e7591/IsSeparable.m#L251

Granted, not *all* of the same tests need to necessarily be included--although I think some of them should be. Any of the ones that are not included should have a "TODO" stub to track any future progress. 

One important thing to mention as well is that if all of these separability tests fail, the final thing that needs to be checked is whether or not a symmetric extension exists for some level of the symmetric extension hierarchy. 

An example of this check is at the end of the `IsSeparable.m` file in QETLAB (https://github.com/nathanieljohnston/QETLAB/blob/21e3115f5676d8650cd1e94107a63772720e7591/IsSeparable.m#L374)

Furthermore, `toqito` does have a function to determine if an operator does have a symmetric extension which can be used for this "final check" if all of the other separability tests fail to detect anything. 

`has_symmetric_extension`: https://github.com/vprusso/toqito/blob/master/toqito/state_props/has_symmetric_extension.py
`symmetric_extension_hierarchy`: 

Hopefully that makes sense, but of course, if I can offer any additional input, do let me know!

meandmytram:
I have realized I got confused with the operational criterion from https://arxiv.org/pdf/quant-ph/0002089.pdf in case of a low-rank operator. I added a procedure to construct a density operator corresponding to the low-rank criterion via SVD, but we have to figure out how to implement the criterion from page 8 which requires searching for cofficients in a matrix decomposition.

meandmytram:
As for the tests, can you please point out the particular tests we would like to have @vprusso ?

vprusso:
> As for the tests, can you please point out the particular tests we would like to have @vprusso ?

I think so long as you've got the symmetric extension catch at the end, the others are mostly just "nice-to-haves". The only suggestion I have here is to track the ones that you won't be implementing and create a task that clarifies which tests are left out. 

> I have realized I got confused with the operational criterion from https://arxiv.org/pdf/quant-ph/0002089.pdf in case of a low-rank operator. I added a procedure to construct a density operator corresponding to the low-rank criterion via SVD, but we have to figure out how to implement the criterion from page 8 which requires searching for cofficients in a matrix decomposition. This would also cover a particular 3x3 state (by the way, I wasn't able to find in the literature whether the rank=4 is really needed).

I'm not *super* familiar with this particular test so my understanding of this one is mostly based on the implementation of the one in QETLAB. Again, if this test is preventing you from wrapping the task up, so long as you track it missing I think that should be okay. 

Does all of that make sense? Of course, if you have any other questions, please don't hesitate to let me know.

meandmytram:
Got it. I think now we can merge the pull request since I have marked the low-rank (aka the decomposition) criterion as a TODO.

vprusso:
Sounds great, @meandmytram. 

Could you also create a task that mentions the outstanding separability tests that are not presently covered in this PR and put in a `TODO-X` stub where "X" is the task number associated with the outstanding tasks?

You'll also want to ensure that the `docs/states.rst` file is adapted to automatically generate the docs from the comments you have by adding the line `toqito.state_props.is_separable` under the "Properties of Quantum States" heading. 

Let me know if you have any further questions, and thanks again!

meandmytram:
Done!

PR #98, opened 2022-06-12
Bug fixes in `partial_transpose` and `sk_operator_norm`.

georgios-ts:
## Description
This commit fixes some Matlab vs Python differences that were missed when porting the code from QETLAB.


vprusso:
Nice catch, @georgios-ts ! Your attention to detail here is next level. Thank you as always for the contribution and the excellent fix!

PR #143, opened 2023-04-19
Parallel repetition for XOR games

juliusw352:
## Description
Adds parallel repetition for XOR games by bridging over to the generic NLG version of the function.

## Questions
-  Would it be better to only create a new NLG object if `reps>1` and use the already existing XOR function otherwise? It might yield some small performance benefits, but I'm not sure.

## Status
-  [ ] Ready to go

## Other comments
My apologies for the long silence – I was quite busy the last few weeks. I also built a separate `classical_value` function to avoid having to create a `NonlocalGame` object, but it is incredibly slow and doesn't seem to like it when `q_0, q_1 > 2`.

This solution seems a bit underwhelming, but it works, and seems to run fairly quickly as well.

### DIFF
--- a/tests/test_nonlocal_games/test_xor_game.py
+++ b/tests/test_nonlocal_games/test_xor_game.py
@@ -126,12 +126,12 @@ class TestXORGame(unittest.TestCase):
 
 	def test_classical_parallel_rep(self):
 		"""Tests for classical value parallel reps."""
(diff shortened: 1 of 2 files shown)


vprusso:
Awesome, thank you as always for the contribution, @juliusw352 !

PR #154, opened 2023-05-27
Update CI/CD

### LINKED ISSUE #146: CI/CD: Setup automated lint + build system
Previously, `toqito` made use of the previously free and supported Travis system for automated test, build, and lint pipelines. 

At present, there is not an automated CI/CD solution in place. I don't have any particularly strong opinions on what technology stack is used here, so I'm all for discussing this at greater length to hone in on an ideal solution. It would be great for tests/linting/etc. to run automatically along with more in-depth reports on important codebase metrics (i.e. code-coverage, etc).

In addition to setting this up, the pipeline would need to successfully run the tests and be linted properly As the tests are currently passing (but not in an automated manner) and the code is fairly well-linted, there should not be too many additional tweaks here.

ryanprior:
Closes #146

## Description
This PR adds automated CI/CD system, replacing the obsolete Travis-based pipeline with an up to date GitHub Actions pipeline. After merge, these pipelines will be available to review in an "Actions" tab in the repo's main navigation.

You can review a run of the pipeline here and verify it's working as you'd expect: https://github.com/ryanprior/toqito/actions/runs/5097437677/jobs/9163961190

## Todos
  - [x] Remove Travis-based pipeline config `.travis.yml`
  - [X] Add equivalent GitHub Action pipeline
  - [X] Allow linting and doc style warnings without failing the CI, to allow pipeline to pass
  - [X] Update `mypy` dependency to 1.3.0, to make precommit happy on Python 3.11

## Future work
Out of scope for this PR but good ideas for follow-ups:
- Invalidate the current codecov API key, as it's exposed in the config file (both before and after this PR.) Create a new one and add it as a repository secret, then transition the CI to pull the value from secrets.
- Fix all failing skipped tests and un-skip them
- Fix all linter and doc style warnings, then reconfigure the pipeline to fail on the same, preventing new ones from getting committed
- Upload the built artifacts to a repository automatically on pipeline success

## Status
-  [X] Ready to go


.github/workflows/build-test-actions.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on: [push]
+
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # Add versions here to expand support matrix
+        python-version: ["3.9"]
```
vprusso, line 11: It's great to check Python 3.9, however I wonder if we should also include 3.10 and 3.11 (especially since 3.11 is pinned in the `pyproject.toml` file.

vprusso (commented):
I should also add if you have any ideas or suggestions for how to further leverage some of the GitHub Actions tools to improve the linting/testing/etc. workflow here, I'm all ears! I think what you've put together is a great start though in any case!

tests/test_matrix_props/test_sk_norm.py
```diff
@@ -44,7 +44,7 @@ def test_sk_norm_rank_1():
     np.testing.assert_equal(np.allclose(lower_bound, expected), True)
     np.testing.assert_equal(np.allclose(upper_bound, expected), True)
 
-
+@pytest.mark.skip(reason="failure not relevant to CI/CD test; re-enable after #146 closed")
```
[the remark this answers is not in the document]
ryanprior: That's great news! I don't see that merged into master, what branch is that fixed in?

.github/workflows/build-test-actions.yml
ryanprior: I pushed a commit adding these versions to the matrix. Let's watch and see how it goes 👀

tests/test_matrix_props/test_sk_norm.py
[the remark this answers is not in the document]
vprusso: Ah, I should have linked it. Here it is:
https://github.com/vprusso/toqito/pull/156

.github/workflows/build-test-actions.yml
ryanprior: Looks like Python 3.10 has at least one problem, which you can see here: https://github.com/ryanprior/toqito/actions/runs/5117092849/jobs/9199863846

So this validates the usefulness of adding more versions to the support matrix, such that we find these problems. Do you prefer to merge this PR now with a wider matrix and failing pipeline though, or merge a green pipeline first and update the matrix later after build issues have been fixed?

tests/test_matrix_props/test_sk_norm.py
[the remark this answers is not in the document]
ryanprior: Great, I've removed the commit that skips the test and pulled that change into my tree. Yay @georgios-ts!

ryanprior:
> I should also add if you have any ideas or suggestions for how to further leverage some of the GitHub Actions tools to improve the linting/testing/etc. workflow here, I'm all ears! I think what you've put together is a great start though in any case!

Once the current linting and style warnings are fixed, I would prioritize type checking using mypy or pyright, and look into automated publishing of docs and packages on release. I can write issues for those if it's helpful.

vprusso:
> Once the current linting and style warnings are fixed, I would prioritize type checking using mypy or pyright, and look into automated publishing of docs and packages on release. I can write issues for those if it's helpful.

Those are all great ideas. Of course, for the purposes of the bounty and this issue, I think they are all beyond the scope of this task. That's not to say I wouldn't appreciate having the issues on the board or even having a few of them picked off, so by all means, if you're keen on doing either or both of those, I would be much obliged!

.github/workflows/build-test-actions.yml
vprusso: Hmm, I wonder if this is a `mypy` issue?
https://github.com/pre-commit/pre-commit/issues/2850

Preferably, I'd prefer to resolve this prior to merging if possible. Reading through the issue, it looks like we might be able to resolve the 3.11 failure by pinning to a specific `mypy` as the OP mentioned.

ryanprior: Great link. [Quick bump to `mypy`](https://github.com/vprusso/toqito/commit/e2d0a92c0aa1523075bf74fbde71181344aec1b0) and we are [all green](https://github.com/ryanprior/toqito/actions/runs/5122486308).

Issue #148, opened 2023-05-05
Feature: Completely bounded trace norm


vprusso:
Compute the "completely bounded trace norm" (CBnorm) of a quantum channel. 

The definition of the CBnorm can be found in Section 3.3.2 of John Watrous's [textbook](https://cs.uwaterloo.ca/~watrous/TQI/TQI.pdf).

A reference implementation in QETLAB can be found [here](https://github.com/nathanieljohnston/QETLAB/blob/master/CBNorm.m).

The functionality for this should be created in `channel_metrics/completely_bounded_trace_norm.py` with corresponding unit tests found in `tests/test_channel_metrics/test_completely_bounded_trace_norm.py`. Be sure to also update the docs in `/docs/channels.rst` under "Metrics of Quantum Channels".

Anish-lab98:
Could you please assign me for this topic? Thank you.

vprusso:
Absolutely @Anish-lab98. You have been assigned! Thank you for your interest and I am looking forward to seeing what you put together!

vprusso:
Tagging @AmanieOxana to assign them.

AmanieOxana:
Hi

PR #157, opened 2023-05-28
cb norm

AmanieOxana:
## Description
Added cb norm function and test. 
I tried the SDP on a unitary, the values are not very far apart, but not equal, help required. resolves https://github.com/vprusso/toqito/issues/148

The cb trace norm is also known as the diamond norm, therefore renamed former `diamond_norm.py`  to `diamond_norm_distance.py` . Btw `np.linalg.norm` is also applicable for matrices.

## Todos
  -  [ ] Check SDP implementation

## Questions
-  [ ] Is it possible for hermitian preserving channel to use y0= y1 during the SDP algorithm?

## Status
-  [ ] Help required to find out why unitary example  I get 1.847765622725633  where I expect 1.414213562373095

### DIFF
--- a/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
+++ b/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
@@ -9,8 +9,8 @@ def test_diamond_norm_unitary():
     U = 1 / np.sqrt(2) * np.array([[1, 1], [-1, 1]]) # Hadamard gate
     phi = np.array([[np.eye(2), np.eye(2)], [U, -U]])
     phi = phi.reshape((4, 4))
-    lam = np.linalg.eig(U)
-    diameter = np.abs(lam[0][0] - lam[0][1])
+    lam, eigv = np.linalg.eig(U)
+    diameter = np.abs(lam[0] - lam[1])
     np.testing.assert_equal(np.isclose(diamond_norm(phi), diameter, atol=1e-3), True)
 def test_diamond_norm_non_square():
     """Non-square inputs for diamond norm."""
--- a/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
+++ b/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
@@ -1,22 +1,30 @@
 import numpy as np
-from toqito.channels import dephasing
-def test_diamond_norm_quantum_channel():
+from toqito.channels.dephasing import dephasing
+from toqito.channel_metrics.completely_bounded_trace_norm import completely_bounded_trace_norm
+
+
+def test_cb_trace_norm_quantum_channel():
     """The diamond norm of a quantum channel is 1"""
     phi = dephasing(1)
-    np.testing.assert_equal(np.isclose(diamond_norm(phi), 1, atol=1e-3), True)
-def test_diamond_norm_unitary():
+    np.testing.assert_equal(completely_bounded_trace_norm(phi), 1)
+
+
+def test_cb_trace_norm_unitary():
     """The diamond norm of phi = id- U id U* is the diameter of the smallest circle that contains the eigenvalues of U"""
     U = 1 / np.sqrt(2) * np.array([[1, 1], [-1, 1]]) # Hadamard gate
     phi = np.array([[np.eye(2), np.eye(2)], [U, -U]])
     phi = phi.reshape((4, 4))
     lam, eigv = np.linalg.eig(U)
     diameter = np.abs(lam[0] - lam[1])
-    np.testing.assert_equal(np.isclose(diamond_norm(phi), diameter, atol=1e-3), True)
-def test_diamond_norm_non_square():
+    np.testing.assert_equal(np.isclose(completely_bounded_trace_norm(phi), diameter, atol=1e-3), True)
+
+
+def test_cb_trace_norm_non_square():
     """Non-square inputs for diamond norm."""
     with np.testing.assert_raises(ValueError):
         phi = np.array([[1, 2, 3], [4, 5, 6]])
-        diamond_norm(phi)
+        completely_bounded_trace_norm(phi)
+
 
 if __name__ == "__main__":
     np.testing.run_module_suite()
--- a/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
+++ b/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
@@ -3,6 +3,7 @@ import numpy as np
 
 from toqito.channels.dephasing import dephasing
 from toqito.channel_metrics.completely_bounded_trace_norm import completely_bounded_trace_norm
+from toqito.channel_ops.kraus_to_choi import kraus_to_choi
 
 
 def test_cb_trace_norm_quantum_channel():
--- a/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
+++ b/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
@@ -2,8 +2,8 @@
 import numpy as np
 
 from toqito.channels.dephasing import dephasing
-from toqito.channel_metrics.completely_bounded_trace_norm import completely_bounded_trace_norm
-from toqito.channel_ops.kraus_to_choi import kraus_to_choi
+from toqito.channel_metrics import completely_bounded_trace_norm
+from toqito.channel_ops import kraus_to_choi
 
 
 def test_cb_trace_norm_quantum_channel():
--- a/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
+++ b/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
@@ -1,0 +1,22 @@
+import numpy as np
+from toqito.channels import dephasing
+def test_diamond_norm_unitary():
+    """The diamond norm of a quantum channel is 1"""
+    phi = dephasing(1)
+    np.testing.assert_equal(np.isclose(diamond_norm(phi), 1, atol=1e-3), True)
+def test_diamond_norm_unitary():
+    """The diamond norm of phi = id- U id U* is the diameter of the smallest circle that contains the eigenvalues of U"""
+    U = 1 / np.sqrt(2) * np.array([[1, 1], [-1, 1]]) # Hadamard gate
+    phi = np.array([[np.eye(2), np.eye(2)], [U, -U]])
+    phi = phi.reshape((4, 4))
+    lam = np.linalg.eig(U)
+    diameter = np.abs(lam[0][0] - lam[0][1])
+    np.testing.assert_equal(np.isclose(diamond_norm(phi), diameter, atol=1e-3), True)
+def test_diamond_norm_non_square():
+    """Non-square inputs for diamond norm."""
+    with np.testing.assert_raises(ValueError):
+        phi = np.array([[1, 2, 3], [4, 5, 6]])
+        diamond_norm(phi)
+
+if __name__ == "__main__":
+    np.testing.run_module_suite()
--- a/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
+++ b/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
@@ -1,3 +1,4 @@
+"""Tests for completely_bounded_trace_norm"""
 import numpy as np
 from toqito.channels.dephasing import dephasing
 from toqito.channel_metrics.completely_bounded_trace_norm import completely_bounded_trace_norm
@@ -5,7 +6,7 @@ from toqito.channel_metrics.completely_bounded_trace_norm import completely_boun
 
 def test_cb_trace_norm_quantum_channel():
     """The diamond norm of a quantum channel is 1"""
-    phi = dephasing(1)
+    phi = dephasing(2)
     np.testing.assert_equal(completely_bounded_trace_norm(phi), 1)
 
 
@@ -20,7 +21,7 @@ def test_cb_trace_norm_unitary():
 
 
 def test_cb_trace_norm_non_square():
-    """Non-square inputs for diamond norm."""
+    """Non-square inputs for cb trace norm."""
     with np.testing.assert_raises(ValueError):
         phi = np.array([[1, 2, 3], [4, 5, 6]])
         completely_bounded_trace_norm(phi)
--- a/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
+++ b/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
@@ -14,7 +14,7 @@ def test_cb_trace_norm_quantum_channel():
 def test_cb_trace_norm_unitaries_channel():
     """The diamond norm of phi = id- U id U* is the diameter of the smallest circle that contains the eigenvalues of U"""
     U = 1/np.sqrt(2) * np.array([[1, 1], [-1, 1]]) # Hadamard gate
-    phi = np.block([[np.eye(2), np.eye(2)],[U, -U]])
+    phi = kraus_to_choi([[np.eye(2), np.eye(2)],[U, -U]])
     lam, eigv = np.linalg.eig(U)
     dist = np.abs(lam[:, None] - lam[None, :]) # all to all distance
     diameter =  np.max(dist)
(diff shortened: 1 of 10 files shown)


AmanieOxana:
resolves #148

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+from toqito.channels import dephasing
+def test_diamond_norm_quantum_channel():
```
vprusso, line 3: We should place in spacing between the functions and between the import statements and the functions. Refer to the other test files to see how this is done.

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,22 @@
...
+    """The diamond norm of a quantum channel is 1"""
+    phi = dephasing(1)
+    np.testing.assert_equal(np.isclose(diamond_norm(phi), 1, atol=1e-3), True)
```
vprusso, line 6: Since this function will return `1` in the event that the channel is quantum, I wonder if we need the `isclose` statement.

toqito/channel_metrics/completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,78 @@
+"Compute the completely bounded trace norm of a quantum channel"
+import numpy as np
+from toqito.channel_ops import apply_channel, dual_channel, choi_to_kraus
+from toqto.channel_props import is_quantum_channel, is_herm_perserving, is_completely_positive, is_unitary
+from toqito.state_metrics import trace_norm
+import cvxpy as cp
+
+def diamond_norm(phi: np.ndarray)-> float:
```
vprusso, line 8: Should the function name here be `completely_bounded_trace_norm`?

toqito/channel_metrics/diamond_norm_distance.py
```diff
@@ -3,7 +3,7 @@
 import numpy as np
 
 
-def diamond_norm(choi_1: np.ndarray, choi_2: np.ndarray) -> float:
+def diamond_norm_distance(choi_1: np.ndarray, choi_2: np.ndarray) -> float:
```
vprusso, line 6: Since the `diamond_norm` function is already in the `channel_metrics/` folder, it's already implied that this is a distance metric. I would opt to keep the name `diamond_norm` over `diamond_norm_distance` (same with the file name as well, etc.)

toqito/channel_metrics/completely_bounded_spectral_norm.py
```diff
@@ -0,0 +1,24 @@
+"Compute the completely bounded spectral norm of a quantum channel"
+from toqito.channel_ops import dual_channel
+from toqito.channel_metrics import completely_bounded_trace_norm
+def cb_spectral_norm(phi):
```
vprusso, line 4: To maintain convention with the rest of the project, I would opt to rename the function to `completely_bounded_spectral_norm`. It is a bit longer, but more explicit, and if the user wishes to condense the name on import, they can do that if they wish.

toqito/channel_metrics/completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,78 @@
...
+                                    + cp.atoms.norm_inf(cp.partial_trace(y1, dims=(dim, dim), axis=1)))
+            problem = cp.Problem(objective, constraints)
+            problem.solve()
+        else:
+            y1 = cp.Variable([dim_Lx, dim_Lx], complex=True)
+            constraints += [y1 == y1.H]
+            constraints += [y1 >> 0]
+            A = cp.bmat([[y0, -phi], [-np.conj(phi).T, y1]])
+            constraints += [  A >> 0  ]
+            objective = cp.Minimize( cp.atoms.norm_inf(cp.partial_trace(y0, dims= (dim,dim), axis = 1))
```
vprusso, line 71: I should really configure some linting pre-commit hooks to catch these things, but as a minor nitpick, `dims=(dim,dim)` instead of `dims = (dim, dim)`.

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,22 @@
...
+    lam, eigv = np.linalg.eig(U)
+    diameter = np.abs(lam[0] - lam[1])
+    np.testing.assert_equal(np.isclose(diamond_norm(phi), diameter, atol=1e-3), True)
+def test_diamond_norm_non_square():
+    """Non-square inputs for diamond norm."""
+    with np.testing.assert_raises(ValueError):
+        phi = np.array([[1, 2, 3], [4, 5, 6]])
+        diamond_norm(phi)
+
+if __name__ == "__main__":
```
vprusso, line 21: It seems odd to me that the `test_completely_bounded_trace_norm` file is testing the diamond norm functionality, right? We probably should also add explicit tests for the `completely_bounded_spectral_norm` as well.

AmanieOxana: I wanted to call `completely_bounded_trace_norm` diamond norm, as in the Wikipedia article, which is easy to confuse with the diamond norm distance functionality in toqito. But sticking to `completely_bounded_trace_norm` is maybe easier in implementation.

AmanieOxana:
I wanted to reproduce the unitary channel test from [here](https://qetlab.com/DiamondNorm#Can_be_used_within_CVX). 
`U = 1 / np.sqrt(2) * np.array([[1, 1], [-1, 1]])  # Hadamard gate
phi = np.array([[np.eye(2), np.eye(2)], [U, -U]])
phi = phi.reshape((4, 4))
`
But `is_unitary(phi)` gives `False`. Does the reshaping change the map?

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
vprusso: Ah, I see where the issue here might be:

- Indeed, the completely bounded trace norm is equivalent to the diamond norm (you are completely correct)
- The `diamond_norm` function as is present on `toqito` computes the SDP for the diamond norm. 
- My mistake in writing the title of this issue is that I should have specified that this is for the "completely bounded norm" (not just the "completely bounded trace norm").

There shouldn't be too much to change on your end though as the completely bounded norm of a channel is equal to the diamond norm of the dual map.

Does that make sense? Let me know if anything I've said above isn't clear, and I'll be happy to clarify. Thank you again!

vprusso:
> here

Hmm, it might be the case that you are invoking the incorrect `is_unitary` function. For instance, there is an `is_unitary` that is applicable to matrices under `matrix_props` and then another `is_unitary` function that is applicable to quantum channels under `channel_props`. Are you invoking the correct one?

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
AmanieOxana: According to [sec. 4](http://theoryofcomputing.org/articles/v005a011) the SDP for the distance between two channels is simpler than the SDP for the completely bounded trace norm of channels (sec. 3). Have you implemented the SDP of sec. 4 in  `diamond_norm.py` @vprusso ? 
I have not implemented SDP from sec.3, but the [QETLAB](https://github.com/nathanieljohnston/QETLAB/blob/master/DiamondNorm.m) SDP, which is also cited in Wikipedia. Yet, i have not seen a mathematical derivaton of this formulation.

vprusso: Hi @AmanieOxana 

Right, the `diamond_norm.py` file in `toqito` (should be) implementing the diamond norm as specified in this QETLAB file:
https://github.com/nathanieljohnston/QETLAB/blob/master/DiamondNorm.m

And, ideally, the `completely_bounded_norm.py` file in `toqito` for this feature would implement the same functionality as in this file on QETLAB:
https://github.com/nathanieljohnston/QETLAB/blob/master/CBNorm.m

Does that clear things up or have I made things more confusing?

vprusso: It is a new day and I have a cup of coffee now. I believe (somehow) I missed what you are saying here (which you were saying perfectly clearly before, I just missed it). 

I agree with everything you've written above. Thank you for sticking with me and your patience :)

AmanieOxana:
Hi @vprusso, I corrected the norm in the SDP and a factor 1/2, now the tests work. 
Still open is the is_unitary case and potentially a hermitian_preserving case. I still miss a mathematical derivation of this SDP, to decide the latter.

AmanieOxana:
Hi @vprusso, I have found a reference for the SDP. The case where the cb trace norm is the spectral diameter, could hold for mixed unitaries, but I have no specifications for this theorem at hand, that I just found in the qetlab wiki, so I left it out. For the hermitian preserving case, the only simplification I see is that the choi matrix is hermitian, but this does not make the sdp significantly faster. 
Would you revise the current version of the implementation? If you have ideas for further enhancements, I am happy to hear about them.

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,32 @@
...
+def test_cb_trace_norm_quantum_channel():
+    """The diamond norm of a quantum channel is 1"""
+    phi = dephasing(2)
+    np.testing.assert_equal(completely_bounded_trace_norm(phi), 1)
+
+
+def test_cb_trace_norm_unitaries_channel():
+    """The diamond norm of phi = id- U id U* is the diameter of the smallest circle that contains the eigenvalues of U"""
+    U = 1/np.sqrt(2) * np.array([[1, 1], [-1, 1]]) # Hadamard gate
+    phi = np.block([[np.eye(2), np.eye(2)],[U, -U]])
```
georgios-ts, line 17: If I may jump in: 
The example in [QETLAB](https://qetlab.com/DiamondNorm#A_difference_of_unitaries_channel) specifies the channel as a cell of left and right Kraus operators. Your implementation of `completely_bounded_trace_norm` works only for Choi inputs. This means you'll first need to obtain a Choi representation of the channel (which is not as simple as stacking the matrices together).

`toqito` already provides this functionality:
```suggestion
    phi = kraus_to_choi([[np.eye(2), np.eye(2)],[U, -U]])
```
and of coure you have to import it first.
```python
from toqito.channel_ops import kraus_to_choi
```

Making this change will let you replace `1/4` factor in https://github.com/vprusso/toqito/blob/c39ac220126d8094c3ed356af1b2edd4b1605cca/toqito/channel_metrics/completely_bounded_trace_norm.py#L63 with `1/2` which is what QETLAB uses and it's in agreement with the SDP formulation found in literature.

AmanieOxana: amazing @georgios-ts, thanks so much for solving it! I was not at all sure why it needs this extra 1/2 resulting in a 1/4, but great that it is not at all the case, that makes me understand it.

vprusso: Outstanding, and thank you, @georgios-ts for the clear and concise solution, and thank you @AmanieOxana for making the change!

AmanieOxana: @vprusso do you have reviewed the code? Do you have further ideas?

vprusso (commented):
Hi @AmanieOxana , 

Thanks for the ping. Overall, looks great! A few nit-pick comments, but excellent overall!

vprusso (commented):
Hi @AmanieOxana , 

Thanks for the ping. Overall, looks great! A few nit-pick comments, but excellent overall!

AmanieOxana:
Thanks for the review @vprusso. I have changed accordingly.

toqito/channel_metrics/completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,66 @@
...
+    .. [JohQET]: Nathaniel Johnston. QETLAB:
+    A MATLAB toolbox for quantum entanglement, version 0.9.
+    https://github.com/nathanieljohnston/QETLAB/blob/master/DiamondNorm.m
+    http://www.qetlab.com, January 12, 2016. doi:10.5281/zenodo.44637
+
+    :raises ValueError: If matrix is not square.
+    :param phi: superoperator as choi matrix
+    :return: The completely bounded trace norm of the channel
+    """
+    dim_Lx, dim_Ly = phi.shape
```
vprusso, line 33: Nitpick: Having capital letters in variable names goes against PEP-8 ([link](https://peps.python.org/pep-0008/#function-and-variable-names))

Perhaps we can rename `dim_Lx` and `dim_Ly` to `dim_lx` and `dim_ly`, respectively? Or do you disagree?

toqito/channel_metrics/completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if dim_Lx != dim_Ly:
```
vprusso, line 35: The way you are doing this here is completely fine, although `toqito` does have a `matrix_props.is_square` function that can also be invoked. Your call as to whether or not you'd like to use that here or not.

toqito/channel_metrics/completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,66 @@
...
+    # SDP
+    y0 = cp.Variable([dim_Lx, dim_Lx], complex=True)
+    constraints = [y0 == y0.H]
+    constraints += [y0 >> 0]
+
+    y1 = cp.Variable([dim_Lx, dim_Lx], complex=True)
+    constraints += [y1 == y1.H]
+    constraints += [y1 >> 0]
+
+    A = cp.bmat([[y0, -phi], [-phi.conj().T, y1]])
```
vprusso, line 56: One more pedantic PEP-8 suggestion of changing `A` to `a_var` (or something like that).

vprusso (approved):
Left a few minor nit-pick comments (I promise these are the last :) )

Approved and good to go once those are in! Great job, @AmanieOxana !

vprusso (approved):
Left a few minor nit-pick comments (I promise these are the last :) )

Approved and good to go once those are in! Great job, @AmanieOxana !

vprusso (approved):
Left a few minor nit-pick comments (I promise these are the last :) )

Approved and good to go once those are in! Great job, @AmanieOxana !

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
```diff
@@ -0,0 +1,33 @@
+"""Tests for completely_bounded_trace_norm."""
+import numpy as np
+
+from toqito.channels.dephasing import dephasing
+from toqito.channel_metrics.completely_bounded_trace_norm import completely_bounded_trace_norm
```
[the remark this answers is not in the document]
AmanieOxana: @vprusso I made this alteration, but this import seems not to work, as not the function but a module is imported. Should I change it to the prior version, or is sth wrong with the init?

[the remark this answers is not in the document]
vprusso: Looks like the `__init__.py` is incorrect. For clarity, the `__init__.py` will look like this:

```py
from toqito.channel_metrics.channel_fidelity import channel_fidelity
from toqito.channel_metrics.diamond_norm import diamond_norm
from toqito.channel_metrics.completely_bounded_spectral_norm import completely_bounded_spectral_norm
from toqito.channel_metrics.completely_bounded_trace_norm import completely_bounded_trace_norm
```

Then, if you want to refer to these functions in another file, you can write the more condensed:

```py
from toqito.channel_metrics import X
```

instead of the longer statement

```py
from toqito.channel_metrics.X import X
```

Does that make sense?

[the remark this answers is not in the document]
AmanieOxana: changed it accordingly @vprusso , I though the init could be made also more concise, but should be fine

vprusso:
Thanks for the quick update, @AmanieOxana . Let me know if I can provide any further help and feel free to let me know when you think this is ready to merge!

vprusso:
> Sure, unless you have no further ideas for simplifications when the channel is unitary or hermitian @vprusso, I think it is ready.

Excellent! Well, in that case, I will go ahead and put this through! Thank you again, @AmanieOxana, for your contribution to the `toqito` project and I have that this was a fun experience for you! It was a privilege on my side, and I hope to see you around here again!

Issue #145, opened 2023-05-05
Refactor: Partial trace and partial transpose


vprusso:
## The problem:
The present implementation of `partial_trace` and `partial_transpose` is not particularly favorable for a few different reasons:

- Counterintuitively, the first index is "1" and opposed to "0". The violates the conventions used elsewhere and is a result of these functions originally being converted over from MATLAB

- The functions themselves are quite convoluted, and hence, more likely to induce error. While the test suite for these functions is fairly thorough, it is still not ideal. 

## The solution:
The solution is fairly straightforward:

- Replace the `partial_transpose` and `partial_trace` functions with the corresponding equivalent functions provided from the [picos module](https://picos-api.gitlab.io/picos/index.html) (that is, `picos.partial_transpose` and `picos.partial_trace`, respectively. 

- The other places in the code that make use of either `partial_trace` or `partial_tranpose` will necessarily need to be updated along with any mention of usage in the docs/tutorials.

ErikQQY:
Hi there! I am interested in this issue, it would be great if you can assign this issue to me!😄

vprusso:
Thank you for your interest @ErikQQY ! You have been assigned!

ErikQQY:
@vprusso I just refactored ```partial_trace``` and ```partial_transpose```, but as for those input matrix is ```Variable``` matrix generated from cvxpy, ```partial_trace``` and ```partial_transpose``` in picos can't handle them😢, how do we handle these usage?

ErikQQY:
For example:
https://github.com/vprusso/toqito/blob/835fcd3e2b0c44f10ba7877710c80edbd49e3860/toqito/state_opt/optimal_clone.py#L176

https://github.com/vprusso/toqito/blob/835fcd3e2b0c44f10ba7877710c80edbd49e3860/toqito/nonlocal_games/quantum_hedging.py#L131

https://github.com/vprusso/toqito/blob/835fcd3e2b0c44f10ba7877710c80edbd49e3860/toqito/state_opt/ppt_distinguishability.py#L180

vprusso:
> @vprusso I just refactored `partial_trace` and `partial_transpose`, but as for those input matrix is `Variable` matrix generated from cvxpy, `partial_trace` and `partial_transpose` in picos can't handle them😢, how do we handle these usage?

Hi @ErikQQY . Thanks for the work thus far and for the question. Two options come to mind:

[...]

PR #158, opened 2023-05-29
Refactor partial_trace and partial_transpose

ErikQQY:
## Description
Fix: #145
This PR refactored ```partial_trace``` and ```partial_transpose```.

## Status
-  [x] Ready to go

### DIFF
--- a/tests/test_channels/test_partial_trace.py
+++ b/tests/test_channels/test_partial_trace.py
@@ -1180,14 +1180,14 @@ def test_partial_trace_64_by_64_subsystems_4_4_4():
 
 def test_partial_trace_invalid_sys_arg():
     """The `sys` argument must be either a list or int."""
-    with np.testing.assert_raises(IndexError):
+    with np.testing.assert_raises(ValueError):
         rho = np.array([[1 / 2, 0, 0, 1 / 2], [0, 0, 0, 0], [0, 0, 0, 0], [1 / 2, 0, 0, 1 / 2]])
         partial_trace(rho, "invalid_input")
 
 
 def test_partial_trace_non_square_matrix_dim_2():
     """Matrix must be square for partial trace."""
-    with np.testing.assert_raises(TypeError):
+    with np.testing.assert_raises(ValueError):
         rho = np.array([[1, 2, 3, 4], [5, 6, 7, 8], [9, 10, 11, 12]])
         partial_trace(rho, [1], [2])
 
--- a/tests/test_channels/test_partial_trace.py
+++ b/tests/test_channels/test_partial_trace.py
@@ -40,6 +40,17 @@ def test_partial_trace_sys():
     np.testing.assert_equal(np.all(bool_mat), True)
 
 
+def test_partial_trace_sys_int():
+    """Supply `sys` argument as int."""
+    test_input_mat = np.arange(1, 17).reshape(4, 4)
+
+    expected_res = np.array([[12, 14], [20, 22]])
+
+    res = partial_trace(test_input_mat, 0)
+
+    bool_mat = np.isclose(expected_res, res)
+    np.testing.assert_equal(np.all(bool_mat), True)
+
 def test_partial_trace_sys_int_dim_int():
     """
     Default second subsystem.
--- a/tests/test_channels/test_partial_trace.py
+++ b/tests/test_channels/test_partial_trace.py
@@ -34,7 +34,7 @@ def test_partial_trace_sys():
 
     expected_res = np.array([[12, 14], [20, 22]])
 
-    res = partial_trace(test_input_mat, 1)
+    res = partial_trace(test_input_mat, [0])
 
     bool_mat = np.isclose(expected_res, res)
     np.testing.assert_equal(np.all(bool_mat), True)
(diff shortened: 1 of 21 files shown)


ErikQQY:
@vprusso I think this PR is ready to go!!

tests/test_channels/test_partial_transpose.py
```diff
@@ -708,26 +708,12 @@ def test_partial_transpose_cvxpy():
     np.testing.assert_equal(isinstance(x_pt, Vstack), True)
 
 
-def test_partial_transpose_non_square():
```
vprusso, line 711: Since we are going back to using the original partial transpose function (with the indices adapted as you've done of course), we may want to include the "non-square" test (with the `sys` argument changed) like so:

```python
def test_partial_transpose_non_square():
    """Test partial transpose on non square matrices ."""
    rho = np.kron(np.eye(2, 3), np.ones((2, 3)))
    rho = np.kron(rho, np.eye(2, 3))

    dim = np.array([[2, 2, 2], [3, 3, 3]])

    res = partial_transpose(rho, sys=1, dim=dim)

    expected = np.kron(np.eye(2, 3), np.ones((3, 2)))
    expected = np.kron(expected, np.eye(2, 3))
    np.testing.assert_equal(np.allclose(res, expected), True)
```

tests/test_channels/test_partial_transpose.py
```diff
@@ -13,19 +13,19 @@ def test_partial_transpose_bipartite():
     rho = np.arange(16).reshape(4, 4)
 
     # Partial transpose of first subsystem:
-    res = partial_transpose(rho, [1], [2, 2])
+    res = partial_transpose(rho, [0], [2, 2])
```
vprusso, line 16: We should also ensure that we include a test to cover the case when `sys` is provided as an `int` (instead of a list with a single `int` inside of it). Something like:

```python
def test_partial_transpose_sys_int():
    """Partial transpose `sys` argument is provided as `int`."""
    test_input_mat = np.arange(1, 17).reshape(4, 4)

    expected_res = np.array([[1, 2, 9, 10], [5, 6, 13, 14], [3, 4, 11, 12], [7, 8, 15, 16]])

    res = partial_transpose(test_input_mat, sys=0)

    bool_mat = np.isclose(res, expected_res)
```

tests/test_channels/test_partial_trace.py
```diff
@@ -34,7 +34,7 @@ def test_partial_trace_sys():
 
     expected_res = np.array([[12, 14], [20, 22]])
 
-    res = partial_trace(test_input_mat, 1)
+    res = partial_trace(test_input_mat, [0])
```
vprusso, line 37: Similarly to the other comment about the `partial_transpose`, we should have a test for when the user supplies an `int` argument for `sys`. Something like:

```python
def test_partial_trace_sys_int():
    """Supply `sys` argument as int."""
    test_input_mat = np.arange(1, 17).reshape(4, 4)

    expected_res = np.array([[12, 14], [20, 22]])

    res = partial_trace(test_input_mat, 0)

    bool_mat = np.isclose(expected_res, res)
```

tests/test_state_props/test_is_separable.py
```diff
@@ -53,10 +53,10 @@ def test_ppt_low_rank():
     u, s, v_h = np.linalg.svd(rho)
     rho_cut = u[:, :m-1] @ np.diag(s[:m-1]) @ v_h[:m-1]
     rho_cut = rho_cut / np.trace(rho_cut)
-    pt_state_alice = partial_trace(rho_cut, 2, dim=[3, 2])
+    pt_state_alice = partial_trace(rho_cut, [1], dim=[3, 2])
```
vprusso, line 56: This line yields the following warning:

```
tests/test_state_props/test_is_separable.py::test_ppt_low_rank
  /Users/vincent.russo/Projects/research/toqito/tests/test_state_props/test_is_separable.py:56: DeprecationWarning: Argument 'dim' to partial_trace is deprecated: Use 'dimensions' instead.
    pt_state_alice = partial_trace(rho_cut, [1], dim=[3, 2])

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

Basically, we just need to alter the `dim=` part to `dimensions=`.

tests/test_state_props/test_is_separable.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
 
 from toqito.state_props.is_separable import is_separable
-from toqito.channels import partial_trace
+from picos import partial_trace
```
vprusso, line 5: Hmm, I think what we want to do here is the same that you did for the `partial_transpose` function.

Namely, we want to invoke the `partial_trace` function from `toqito` (instead of from `picos`) with the change that the index offset is `0` instead of `1` (as you did for the `partial_transpose`)

Does that make sense?

vprusso (commented):
Looking good! 

Adding a few minor comments. Let me know if you have any questions and we can hash it out!

tests/test_state_props/test_is_separable.py
ErikQQY: Yeah, I just changed that

vprusso:
Awesome, thanks for making the changes! I think this diff looks good to go on my end. If you agree, I'll go ahead and merge, but wanted to check with you beforehand. @ErikQQY

ErikQQY:
Let’s go!!

vprusso:
Awesome! Thank you again @ErikQQY for your contribution!

PR #160, opened 2023-05-31
Add `channel_dim`.

### LINKED ISSUE #32: Feature: Channel dimensions
Given a quantum channel, compute the input, output, and environment dimensions of the quantum channel. 

The channel should be provided as input as either a Choi matrix or a `List` or Kraus operators. 

In spirit, the approach for computing these dimensions align with the QETLAB function `superoperator_dims` found [here](http://www.qetlab.com/Superoperator_dims).

For reference, here is some code written that partially implements the solution:

```python
def channel_dim(phi, allow_rect=True, dim=None):
    if isinstance(phi, list):
        pass

    else:
        r, c = phi.shape
        da = np.array([np.round(np.sqrt(r)), np.round(np.sqrt(c))])
        db = da

        if dim is None:
            dim = np.array([da.conj().T, db.conj().T])

        dim = expand_dim(dim)

        if r != c and not allow_rect:
            raise ValueError("ERR")
        if dim[0, 0] * dim[0, 1] != r or dim[1, 0] * dim[1, 1] != c:
            raise ValueError("ERR")

        # TODO calc environ dim

    # Finally put DIM back into DA and DB
    if allow_rect:
        da = np.array([dim[0, 0], dim[1, 0]])
        db = np.array([dim[0, 1], dim[1, 1]])
    else:
        da = dim[0, 0]
        db = dim[0, 1]

    return da, db
    
def expand_dim(dim):
    sz = dim.shape
    if np.max(sz) == 1:
        return np.array([[dim, dim], [dim, dim]])
    elif np.min(sz) == 1:
        return np.array([[dim[:].conj().T], [dim[:].conj().T]])
    elif np.max(sz) == 2:
        return dim
    else:
        raise ValueError("Error")    
```

georgios-ts:
## Description
This commit adds a helper function to compute the input, output and environment dimensions of a channel. It will be used in subsequent PRs to fix places where we try to retrieve the dimensions of the channel.

Closes #32

## Status
-  [x] Ready to go

### DIFF
--- a/tests/test_helper/test_channel_dim.py
+++ b/tests/test_helper/test_channel_dim.py
@@ -1,0 +1,105 @@
+"""Test channel dimensions."""
+import numpy as np
+import pytest
+
+from toqito.helper import channel_dim
+from toqito.matrices import pauli
+from toqito.perms import swap_operator
+
+
+def test_channel_dim_kraus():
+    """
+    Test channel dim of the superoperator PHI defined by:
+
+    Phi(X) = [[1,5],[1,0],[0,2]] X [[0,1][2,3][4,5]].conj().T -
+    [[1,0],[0,0],[0,1]] X [[0,0][1,1],[0,0]].conj().T
+    """
+    kraus_1 = np.array([[1, 5], [1, 0], [0, 2]])
+    kraus_2 = np.array([[0, 1], [2, 3], [4, 5]])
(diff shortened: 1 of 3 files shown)


vprusso (approved):
Fantastic work as always! Thank you so much for your continued efforts and contribution to `toqito`! It's always wonderful to see you here!

PR #163, opened 2023-06-02
Minor fix in `choi_to_kraus`.

georgios-ts:
## Description
This commit fixes an oversight in `choi_to_kraus` where we failed to build correctly the right Kraus operators for non-positive hermitian-preserving channels. (we forgot to discard small eigenvalues!!)

## Status
-  [x] Ready to go


georgios-ts:
@vprusso 
I'm afraid we merged #162 a bit early! There was a bug I was late to found out and hopefully this PR fixes it.

vprusso:
My mistake! Thank you for the due diligence in patching that up, and my apologies again for missing that!

PR #164, opened 2023-06-02
Fix `is_unitary` for channels.

georgios-ts:
## Description
Previously the check in place was not sufficient to determine if a channel is unitary. We should assert that the channel is described by one Kraus operator and this matrix is unitary.

## Status
-  [x] Ready to go


vprusso:
As mentioned in the last thread, I sincerely appreciate your professionalism and due diligence--thank you, and again my apologies for merging prematurely!

PR #165, opened 2023-06-02
SDPs Fidelity of Separability

purva-thakre:
## Description
Fixes #141 

Second try to fix this issue because I can never manage to rebase properly. :sweat_smile:

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] Update docstring with more information about the SDP
  - [x] Clarify these are only for pure states
  - [x] Examples in the docstring
  - [x] Create a new issue & assign to myself for mixed states ([link to the issue](https://github.com/vprusso/toqito/issues/166#issue-1738859267))

## Questions
-  N/A

## Status
-  [x] Ready to go

PR #172, opened 2023-06-04
Move Trace Norm

### LINKED ISSUE #167: Refactor: Location of trace norm should be changed
At present the `trace_norm.py` file is located in the `state_metrics/` directory. Generally, the other metrics that are contained within the `state_metrics/` directory are some type of distance measure between one state and another. The trace norm function breaks this pattern and is really just a measure of the state itself (instead of being a measure of a state against another state). 

It may be more appropriate, therefore, to move the `trace_norm` functionality inside of the `matrix_props/` directory. This would also seem to make somewhat more categorical sense, as `sk_norm.py` is contained in `matrix_props/` as well.

purva-thakre:
## Description
Fixes #167 

## Todos
Moved `trace_norm`, fixed import paths & updated docs. 

## Questions
N/A

## Status
-  [x] Ready to go


purva-thakre:
Same as what was said in #165, `autodocsummary` is adding more files than expected. 

Here, only wanted `docs/_autosummary/toqito.matrix_props.trace_norm.rst`.

vprusso:
Hi @purva-thakre. Thanks for tackling this one!

It might be worthwhile to mention that when https://github.com/vprusso/toqito/pull/157 goes in, I imagine some parts of the PR may have to be (slightly) updated. Just a heads up!

docs/_autosummary/toqito.matrix_ops.inner_product.rst
```diff
@@ -1,6 +0,0 @@
-toqito.matrix\_ops.inner_product
```
vprusso, line 1: Hmm, I'm not sure we want to delete the files that are already present here. 

I think you are correct though in expecting that these *should* be updated as part of the docs build when new features go in. However, I think these ones (which are presently on `master` shouldn't be removed.

toqito/state_metrics/helstrom_holevo.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
 
 from toqito.matrix_props import is_density
-from toqito.state_metrics import trace_norm
+from toqito.matrix_props.trace_norm import trace_norm
```
vprusso, line 5: Just a note that if `from toqito.matrix_props.trace_norm import trace_norm` is in the `__init__.py` file (as you did), then you can condense this a bit to:

```suggestion
from toqito.matrix_props import trace_norm
```

toqito/state_metrics/trace_distance.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
 
 from toqito.matrix_props import is_density
-from toqito.state_metrics import trace_norm
+from toqito.matrix_props.trace_norm import trace_norm
```
vprusso, line 5: Similar comment here:

```suggestion
from toqito.matrix_props import trace_norm
```

toqito/state_props/is_separable.py
```diff
@@ -7,7 +7,7 @@
 from toqito.matrix_props import is_positive_semidefinite
 from toqito.state_props import is_ppt, in_separable_ball
 from toqito.state_props.has_symmetric_extension import has_symmetric_extension
-from toqito.state_metrics import trace_norm
+from toqito.matrix_props.trace_norm import trace_norm
```
vprusso, line 10: Similar comment here:

```suggestion
from toqito.matrix_props import trace_norm
```

docs/_autosummary/toqito.matrix_ops.inner_product.rst
purva-thakre: Sure. I deleted those because `make html` was editing them. 

image

I will copy-paste what's in the current files.

vprusso: Sure, you could probably also just do `git checkout <NAME-OF-FILE>` as well to replace them with what is presently on `master`. Either way works!

purva-thakre:
> This looks great, @purva-thakre! Just a heads up that there might be a change or two when you rebase from the changes in this MR: #157
> 
> Let me know if I can help out with that if need be!

@vprusso Thanks for the offer! I always forget I need to force push after a rebase & then panic. This PR should be good now.

vprusso:
Excellent! Looks good to me! 🚀 

Thanks again, @purva-thakre !

Issue #33, opened 2021-01-28
Feature: Is UPB (unextendible product basis)


vprusso:
Given a collection of vectors, determine if the set constitutes an unextendible product basis (UPB). 

According to [arXiv:quant-ph/9808030](https://arxiv.org/abs/quant-ph/9808030):

> An unextendible product basis (UPB) for a multipartite quantum system is an incomplete orthogonal product basis whose complementary subspace contains no product state. 

In spirit, the functionality of being able to determine whether a collection of vectors form a UPB can be derived from the QETLAB package, specifically the function "[IsUPB](http://www.qetlab.com/IsUPB)". 

The functionality for this should be created in `state_props/is_unextendible_product_basis.py` with corresponding unit tests found in `tests/test_state_props/test_is_unextendible_product_basis.py`. Be sure to also update the docs in `/docs/states.rst` under "Properties of Quantum States" with `toqito.state_props.is_unextendible_product_basis.py`. 

Please ensure proper credit is provided to QETLAB as serving as the inspiration for including this function.

vprusso:
Hi Sumit, 

That is awesome! Thank you for taking an interest in this issue and I'm sincerely looking forward to what you put together for this issue. Of course, do take advantage of the GitHub thread here for any questions, discussions, etc. I'm here to help and am very much looking forward to seeing this in `toqito`!

smtsjhr:
Hey Vincent,

 :rocket:
Just following up to say I think i have a working implementation of the main functionality! Consider me committed to making a PR for this issue eventually.. I still need to polish things up, write tests, and do the docs. 

:question:
I will likely ask for some more clarification and guidance soon, including a potential suggestion to move a helper function I defined into an appropriate place like `toqito/helpers` or perhaps `toqito/perms`. And also if I should write separate unit tests for the helper function in addition to testing the main functionality. 

:family:
To any one else interested from UnitaryHack or not, I am still open to collaborate in whatever capacity. Bounty is all yours if we get it, I am just happy to contribute!

vprusso:
> Just following up to say I think i have a working implementation of the main functionality! Consider me committed to making a PR for this issue eventually.. I still need to polish things up, write tests, and do the docs.

Woo-hoo, that is fantastic--great job! Looking forward to seeing what you've put together!

[...]

PR #175, opened 2023-06-10
Is unextendible product basis

smtsjhr:
## Description
Adds a new feature: "Is UPB (unextendible product basis)" requested in #33 .

The implementation closely follows the one given in QETLAB: [IsUPB](https://qetlab.com/IsUPB).
The function decides `True` or `False` if the input is a UPB or not, and when `False` also returns a witness state which is a product state orthogonal to all states of the non-UPB. Note that the existence of a witness state is in general not unique, and that our implementation returns just one particular reproducible example. 

The main function has the following signature:
`is_unextendible_product_basis(local_states_list: list[np.ndarray]): -> (isUPB, witness)`

([View the DOCSTRING here on this comment](https://github.com/vprusso/toqito/pull/175#issuecomment-1588115488))
(Note: Some minor formatting changes were made to conform to GitHub markdown, but otherwise this is the docstring verbatim)

**Input:**

The input `local_states_list` specifies the supposed UPB as a list of 2D arrays of type `numpy.ndarray`, whose length corresponds to the number of parties with each party associated to its corresponding list index.  Each array corresponds to that respective parties local states comprising the UPB, with the number of rows of each array corresponding to the local dimension of that party. The number of columns of each array must be identical and corresponds to the number of product states in the UPB.
In this way, the k-th product state of the UPB is constructed by taking the tensor product of the local states given by the k-th column of each array.

* (1?) The form this input is specified following the same convention made in the QETLAB implementation. This form is just fine in my opinion, but we can consider other forms.  Same goes for the form of the output `witness` state when it exists. For example, instead of represented the input local states as column vectors of the arrays, we can do row vectors. If anything would maybe amount to like 2 - 4 less lines of code i think!
*  (1.1?) Another important point to consider is representing the input product states in this factored "local form" VS "global form". Although a bit contrived, the "local form" we require as input in this implementation guarantees that the states conform to the parties product structure. Otherwise, if we allow input to be expresssed in a more general "global form" what guarantees that the input actually has the appropriate product structure? I need to think a bit more about how the implementation would need to be modified... We could check for the product structure within the function's implementation, but that would require hard separability tests in general. Perhaps we can consider this optional functionality and give the user the option of specifying the input states in "local form" vs "global form". 

**Return:**

The function returns a Tuple of the form `(isUPB, witness)`.
Here `isUPB` is a boolean which is `True` if the input is a UPB, and `False` otherwise.
If `isUPB` is `True`, then `witness` is `None`.
If  `isUPB` is `False`, then `witness` is a list containing a local state for each party. The product state resulting from taking the tensor product of all the local states in `witness` is orthogonal to all the  nonUPB input states.

**Comments for helper function:**

The main function `is_unextendible_product_basis()` makes use of a helper function named `item_partitions()` also contained in the module `toqito/state_props/is_unextendible_product_basis.py`, which is based off the  QETLAB function [`vec_partitions()`](https://qetlab.com/vec_partitions).
* (2?) Should this helper function `item_partitions()` be moved into its own file from the main module `is_unextendible_product_basis.py`, and  with a separate test file to test it?

In regards to the functions complexity (not too sure about this) i think non poly time and space comes from searching through a stored list of permissible partitions, which is constructed with a single call of the helper function `item_partitions()`. The relevant problem size parameter in this context is "the number of product states of the UPB" in question, which corresponds to the "number of items" we partition. In general this list would be huge for a relatively small number of items, say 10 - 30. However in practice,  the main function `is_unextendible_product_basis()` calls the helper `item_partitions()` with partition size constraints which typically makes the list of partitions smaller (not sure by how much though).
* (3?) Perhaps there is a way to improve space performance here. Already, in the implementation of `item_partitions()` I took advantage of Python's `itertools.combinations` to use an iterator for the loop over relevant item combinations. Does it make sense, or is it even possible, to also implement `item_partitions()` as some type of custom Python "Generator" or "Iterator" object? (Please excuse my ignorance here!). Regarding time complexity,  it seems necessary to iterate through the permissible partitions. So I am not sure how to speed this up unless we come up with a more clever way of eliminating permissible partitions.

**Testing Methodology:**

Currently all new tests are in a single file  `tests/test_state_props/test_is_unextendible_product_basis.py`. ~~Tests for the helper function `item_paritions()` still need to be written in either the same or different file. ~~

My testing of the function `is_unextendible_product_basis()` is quite extensive but can maybe still be improved. In particular, there are several tests for verifying a `ValueError` is raised upon invalid input. 
* (4?) The input error handling implemented in the function `is_unextendible_product_basis()` seems excessive and can maybe be consolidated. 

To test UPB properties I used two main cases of UPBs: Tiles and GenShifts described further below (and in [this paper](https://arxiv.org/abs/quant-ph/9908070v3)).
These two UPBs are constructed in the  functions `tiles_local_state_list()` and `genshifts_local_state_list()` defined in the same test file `test_is_unextendible_product_basis.py`.
* (5?) Is it worth moving the functions into their own module files in `toqito/states`, and writing docs and unit tests for them? _EDIT:_ I noticed there is already a file `toqito/states/tile.py',  but that function returns the states in global form and this UPB implementation requires the local state factors as input. So maybe we can add optional functionality to return the states in a desired form: either"local" or "global". 

Tiles is a 5 state UPB of two parties each holding a qutrit. A unit test verifies this case.
If any number of product states from Tiles is removed, the set is no longer a UPB.
Unit tests are done to check Tiles for each of these cases with 1, 2, 3, and 4 states removed.
For each case, the generated witness is tested to be orthogonal to each input product state of the nonUPB.
The TIles UPB (with states removed) are good test cases, because they cover all the control flow branches of the function's implementation. This includes the positive case when `is_unextendible_product_basis()`returns `(True, None)`, and two instances of the negative case when `(False, witness)` returned. These two negative cases correspond to whether or not there are any permissible partitions to search through, and the returned `witness` is constructed differently depending on the case.

GenShifts is a family of UPBs of a m-party system (with odd m >=3 ) each consisting of a qubit, that has a m+1 product states in its UPB. Unit tests verify this in the case of 3, 5, and 7 parties. Note that there is freedom in specifying the exact form of the local states involved in this UPB, but ours follows the same choice used in the QETLAB implementation. (I can add details about this representation in the functions TODO docstring.)
The GenShifts UPB family is a good test case, because it allows to test our implementation with variable number of parties (not just 2 parties as in the Tiles UPB).

**Files added:**
1) `toqito/state_props/is_unextendible_product_basis.py`
2) `tests/test_state_props/test_is_unextendible_product_basis.py`

**Files modified:**
1) `docs/states.rst`

## Todos
(A) Things that still need to be done:
  -  [x] Add more thorough docstrings with theory and code examples.
           (I am having some issues viewing rendered docs, so I am concerned of typos)
  -  [x] Add inline code commentary for sake of explanation
  -  [x] Write unit tests for helper function `item_partitions()`
          (These tests are all in the same file: `test_is_unextendible_product_basis.py` )
          (Hopefully sufficient, but maybe even excessive, but whatever ... they are easy to delete!)

(B) Things that maybe should be done:
  -  [ ] (1?) Discuss if we are satisfied with the form of functions input and output.
  -  [ ] (2?) Move the helper function `item_partitions` into its own file from the main module `is_unextendible_product_basis.py`, ~~and write tests/docs for it~~.
  -  [ ] (3?) Improve algorithm performance by improving performance of `item_partitions()`.
  -  [ ] (4?) Improve input error handling of `is_unextendible_product_basis()`.
  -  [ ] (5?) Move the functions `tiles_local_state_list()` and `genshifts_local_state_list()` in  `test_is_unextendible_product_basis.py` that construct the Tiles and GenShifts UPBs into the folder `toqito/states`, and ~~write unit tests/docs for them~~.

## Questions
-  [ ] Pease advise regarding Todos (B).

## Status
-  [x] Ready to go

### DIFF
--- a/docs/states.rst
+++ b/docs/states.rst
@@ -73,6 +73,7 @@ Properties of Quantum States
     toqito.state_props.is_ppt
     toqito.state_props.is_product
     toqito.state_props.is_pure
+    toqito.state_props.is_unextendible_product_basis
     toqito.state_props.l1_norm_coherence
     toqito.state_props.log_negativity
     toqito.state_props.negativity
--- a/tests/test_state_props/test_is_unextendible_product_basis.py
+++ b/tests/test_state_props/test_is_unextendible_product_basis.py
@@ -1,0 +1,142 @@
+"""Test is_unextendible_product_basis."""
+
+import numpy as np
+import pytest
+
+from toqito.state_props.is_unextendible_product_basis import (
+    is_unextendible_product_basis,
+)
+from toqito.matrix_ops import tensor
+from toqito.matrix_ops import inner_product
+
+
+def Tiles():
+    """Constructs Tiles UPB."""
+    tiles_A = np.zeros([3, 5])
+    tiles_A[:, 0] = [1, 0, 0]
+    tiles_A[:, 1] = [1 / np.sqrt(2), -1 / np.sqrt(2), 0]
+    tiles_A[:, 2] = [0, 0, 1]
+    tiles_A[:, 3] = [0, 1 / np.sqrt(2), -1 / np.sqrt(2)]
+    tiles_A[:, 4] = [1 / np.sqrt(3), 1 / np.sqrt(3), 1 / np.sqrt(3)]
+
+    tiles_B = np.zeros([3, 5])
+    tiles_B[:, 0] = [1 / np.sqrt(2), -1 / np.sqrt(2), 0]
+    tiles_B[:, 1] = [0, 0, 1]
+    tiles_B[:, 2] = [0, 1 / np.sqrt(2), -1 / np.sqrt(2)]
+    tiles_B[:, 3] = [1, 0, 0]
+    tiles_B[:, 4] = [1 / np.sqrt(3), 1 / np.sqrt(3), 1 / np.sqrt(3)]
+
+    return [tiles_A, tiles_B]
+
+
+def GenShifts(parties: int):
+    """Constructs GenShifts UPB for odd number of parties."""
+    if (parties % 2 == 0) or parties == 1:
+        raise ValueError("Input must be an odd int greater than 1.")
+
+    k = int((parties + 1) / 2)
+    num_states = 2 * k
+    upb = [np.zeros([2, num_states]) for _ in range(parties)]
+    angles = [(i / k) * np.pi / 2 for i in range(2 * k - 1, k, -1)]
+    local_states = [[np.cos(angle), np.sin(angle)] for angle in angles]
+    orth_angles = [(i / k) * np.pi / 2 for i in range(1, k)]
+    orth_local_states = [[np.cos(angle), np.sin(angle)] for angle in orth_angles]
+
+    state_list = []
+    state_list.append([0, 1])
+    state_list.extend(local_states)
+    state_list.extend(orth_local_states)
+
+    for party in upb:
+        party[:, 0] = [1, 0]
+
+    for state in range(1, num_states):
+        for i, party in enumerate(upb):
+            party[:, state] = state_list[i]
+        state_list = state_list[-1:] + state_list[:-1]
+
+    return upb
+
+
+def test_is_unextendible_product_basis_input_empty_list():
+    """Empty list as input."""
+    with np.testing.assert_raises(ValueError):
+        empty_list = []
+        is_unextendible_product_basis(empty_list)
+
+
+def test_is_unextendible_product_basis_input_not_numpy_arrays():
+    """List elements are not type numpy.ndarray."""
+    with np.testing.assert_raises(ValueError):
+        list_of_listarrays = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]]
+        is_unextendible_product_basis(list_of_listarrays)
+
+
+def test_is_unextendible_product_basis_input_arrays_not_two_dimensional():
+    """Arrays are not two-dimensional."""
+    with np.testing.assert_raises(ValueError):
+        array_1D = np.array([1, 2, 3, 4])
+        input = [array_1D, array_1D]
+        is_unextendible_product_basis(input)
+
+
+def test_is_unextendible_product_basis_input_arrays_not_same_num_columns():
+    """Arrays do not have the same number of columns"""
+    with np.testing.assert_raises(ValueError):
+        array_1 = np.array([[1, 2], [3, 4]])
+        array_2 = np.array([[1], [3]])
+        input = [array_1, array_2]
+        is_unextendible_product_basis(input)
+
+
+def test_is_unextendable_product_basis_tiles():
+    """Verify if Tiles UPB is a UPB"""
+    res = is_unextendible_product_basis(Tiles())
+    expected_res = [True, None]
+    np.testing.assert_array_equal(res, expected_res)
+
+
+@pytest.mark.parametrize("num_states", [1, 2, 3, 4])
+def test_is_unextendable_product_basis_tiles_remove_states_false(num_states):
+    """Check if Tiles UPB fails to be a UPB when 1, 2, 3, and 4 states are removed."""
+    tiles = Tiles()
+    res = is_unextendible_product_basis(
+        [tiles[0][0:, 0:num_states], tiles[1][:, 0:num_states]]
+    )
+    expected_res = False
+    np.testing.assert_equal(res[0], expected_res)
+
+
+@pytest.mark.parametrize("num_states", [1, 2, 3, 4])
+def test_is_unextendable_product_basis_tiles_remove_states_orthogonal_witness(
+    num_states,
+):
+    """Check if witness is orthogonal to Tiles UPB when 1, 2, 3, and 4 states are removed."""
+    tiles = Tiles()
+    witness = is_unextendible_product_basis(
+        [tiles[0][:, 0:num_states], tiles[1][:, 0:num_states]]
+    )[1]
+    witness_product = tensor(witness[0], witness[1])
+    # Perhaps we can use `toqito.state_props.is_mutually_orthogonal` to test orthogonallity
+    inner_product_list = []
+    for i in range(num_states):
+        UPB_state_product = tensor(
+            np.array([tiles[0][:, i]]).T, np.array([tiles[1][:, i]]).T
+        )
+        ip = inner_product(witness_product[:, 0], UPB_state_product[:, 0])
+        inner_product_list.append(ip)
+    res = inner_product_list
+    expected_res = [0] * num_states
+    np.testing.assert_array_almost_equal(res, expected_res)
+
+
+@pytest.mark.parametrize("num_parties", [3, 5, 7])
+def test_is_unextendable_product_basis_tiles_GenShifts(num_parties):
+    """Verify if GenShifts UPB is a UPB for 3, 5, and, 7 parties"""
+    res = is_unextendible_product_basis(GenShifts(num_parties))
+    expected_res = [True, None]
+    np.testing.assert_array_equal(res, expected_res)
+
+
+if __name__ == "__main__":
+    np.testing.run_module_suite()
--- a/tests/test_state_props/test_is_unextendible_product_basis.py
+++ b/tests/test_state_props/test_is_unextendible_product_basis.py
@@ -95,41 +95,51 @@ def test_is_unextendable_product_basis_tiles():
     expected_res = [True, None]
     np.testing.assert_array_equal(res, expected_res)
 
-
+# Test parameter ranges over number of states removed out of 5 original Tiles UPB states.
 @pytest.mark.parametrize("num_states", [1, 2, 3, 4])
 def test_is_unextendable_product_basis_tiles_remove_states_false(num_states):
     """Check if Tiles UPB fails to be a UPB when 1, 2, 3, and 4 states are removed."""
     tiles = Tiles()
-    res = is_unextendible_product_basis(
-        [tiles[0][0:, 0:num_states], tiles[1][:, 0:num_states]]
-    )
+    # Remove some number of states from the Tiles UPB.
+    tiles_with_states_removed  = [tiles[0][0:, 0:num_states], tiles[1][:, 0:num_states]]
+    res = is_unextendible_product_basis(tiles_with_states_removed)
+    # We expect these to not be a valid UPB.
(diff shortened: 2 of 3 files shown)


smtsjhr:
### Docstring for function `is_unextendible_product_basis()`
([View the source code here](https://github.com/smtsjhr/toqito/blob/0fcf6e4abc2648b993d22f15dc039e7dd5316b1c/toqito/state_props/is_unextendible_product_basis.py#L132))
```
def is_unextendible_product_basis(local_states_list: list[np.ndarray]) -> tuple[bool, list[np.ndarray] | None]:
```

Determine if a collection of pure states is an Unextendible Product Basis (UBP) [UPB99]_.

A UPB is a set of mutually orthogonal product states spanning a proper subspace of a multipartite system whose complementary subspace contains no product state.
    
Let 
```math
 \mathcal{H} = \bigotimes_{i=1}^{m} \mathcal{H}_{i}
```
 be an $m$-party system, where $\mathcal{H}_{i} = \mathbb{C}^{d_i}$ is a complex Euclidean space of dimension $d_i$ of the $i$-th party. Consider a collection $\mathcal{C}=\{|\psi_j\rangle\}_j$ of product states of $\mathcal{H}$ expressed as
```math
\begin{align*}
|\psi_j\rangle = \bigotimes_{i=1}^{m} |\varphi_{i,j}\rangle,
\end{align*}
```
where $|\varphi_{i,j}\rangle \in \mathcal{H}_i$ are local states belonging to the $i$-th party.  Then then collection $\mathcal{C}$ is a UPB if the following hold.


(i) The states are mutually orthogonal:

```math
\langle \psi_j\|\psi_k\rangle = 0 \text{ if} j \neq k.
```

(ii) The states span a proper subspace:
```math
\mathcal{H}_\mathcal{C} = span(|\psi_j\rangle))  \subset \mathcal{H}.
````

(iii) The complementary subspace contains no product state:
```math
|\psi\rangle \in \mathcal{H} - \mathcal{H}_\mathcal{C} \implies |\psi\rangle \text{ is entangled.}
```

### Examples
==========

For example, the Tiles UPB is a set of $5$ states in the bipartite qutrit space $\mathbb{C}^3\otimes \mathbb{C}^3$ given by:
```math
\begin{align*}
|\psi_0\rangle &= \tfrac{1}{\sqrt{2}}|0\rangle \otimes \left(|0\rangle - |1\rangle \right), \\
|\psi_1\rangle &= \tfrac{1}{\sqrt{2}}\left(|0\rangle - |1\rangle \right) \otimes |2\rangle, \\
|\psi_2\rangle &= \tfrac{1}{\sqrt{2}}|2\rangle \otimes \left(|1\rangle - |2\rangle \right), \\
|\psi_3\rangle &= \tfrac{1}{\sqrt{2}}\left(|1\rangle - |2\rangle \right) \otimes |0\rangle, \\
|\psi_4\rangle &= \tfrac{1}{3}\left(|0\rangle + |1\rangle + |2\rangle \right) \otimes \left(|0\rangle + |1\rangle + |2\rangle \right).
\end{align*}
```

The product states of the Tiles UPB can be factored into two sets, $\mathcal{C}_0$ and $\mathcal{C}_1$, of local qutrit states over the biparition $\mathbb{C}^3\otimes \mathbb{C}^3$ as follows.

The first party has states 
```math
\mathcal{C}_0 = \{|\varphi_{0,j}\rangle\}_j
```
in $\mathbb{C}^3$:
```math
\begin{align*}
|\varphi_{0, 0}\rangle &= |0\rangle , \\
|\varphi_{0, 1}\rangle &= \tfrac{1}{\sqrt{2}}\left(|0\rangle - |1\rangle \right), \\
|\varphi_{0, 2}\rangle &= |2\rangle , \\
|\varphi_{0, 3}\rangle &= \tfrac{1}{\sqrt{2}}\left(|1\rangle - |2\rangle \right), \\
|\varphi_{0, 4}\rangle &= \tfrac{1}{\sqrt{3}}\left(|0\rangle + |1\rangle + |2\rangle \right).
\end{align*}
```
The second party has states 
```math
\mathcal{C}_1 = \{|\varphi_{1,j}\rangle\}_j 
```
in $\mathbb{C}^3$:
```math
\begin{align*}
|\varphi_{1, 0}\rangle &= \tfrac{1}{\sqrt{2}}\left(|0\rangle - |1\rangle \right), \\
|\varphi_{1, 1}\rangle &= |2\rangle,  \\
|\varphi_{1, 1}\rangle &= \tfrac{1}{\sqrt{2}}\left(|1\rangle - |2\rangle \right), \\
|\varphi_{1, 3}\rangle &= |0\rangle, \\
|\varphi_{1, 4}\rangle &= \tfrac{1}{\sqrt{3}}\left(|0\rangle + |1\rangle + |2\rangle \right).
\end{align*}
```


When using `toqito` to determine if a collection of $n$ product states $\mathcal{C}=\{|\psi_j\rangle\}_j$ of an $m$-party system is a UPB, the states in question must be provided as an ordered list of $m$ local collections 
```math
\mathcal{C}_i=\{|\varphi_{i,j}\rangle\}_j
```
 comprising each of the product state's local factors (as shown above for the Tiles UPB). 

The particular data structure used here to represent the local collections  of states are two dimensional arrays of type `numpy.ndarray`, where the $j$-th column of each array defines the local state $|\varphi_{i,j}\rangle$ of the respective party. In this way, each array has the same number of columns corresponding to the number of product states $n$. Moreover, the array representing the local collection $\mathcal{C}_i$ of the $i$-th party must have $d_i$ many rows corresponding to the local dimension $d_i$ of the $i$-th party.

Below we verify that the Tiles states are indeed a UPB.

```
>>> # Array for the first party's local states of Tiles UPB.
>>> C_0 = np.array( [[1,  1/np.sqrt(2), 0,             0, 1/np.sqrt(3)],
>>>                 [0, -1/np.sqrt(2), 0,  1/np.sqrt(2), 1/np.sqrt(3)],
>>>                 [0,             0, 1, -1/np.sqrt(2), 1/np.sqrt(3)]]
>>>                 )
>>> # Array for the second party's local states of Tiles UPB.
>>> C_1 = np.array( [[ 1/np.sqrt(2), 0,             0, 1,  1/np.sqrt(3)],
>>>                 [-1/np.sqrt(2), 0,  1/np.sqrt(2), 0, 1/np.sqrt(3)],
>>>                 [            0, 1, -1/np.sqrt(2), 0,  1/np.sqrt(3)]]
>>>                 )
>>> # Construct list of local states of all parties.
>>> tiles = [C_0, C_1]
>>> # Verify Tiles is a UPB.
>>> isUPB, witness = is_unextendible_product_basis(tiles)
(True, None)
```
If any number of states from the Tiles UPB are removed, then the remaining collection of states is no longer a UPB. Here we demonstrate this by removing the $5$-th state from the Tiles UPB.

```
>>> # Remove the 5th state from the Tiles UPB
>>> tiles_state_removed = [C_0[:, 0:4], C_1[:, 0:4]]
>>> # Verify Tiles with a state removed is not a UPB.
>>> isUPB, witness = is_unextendible_product_basis(tiles_state_removed)
(False, [array([[0.],
                [0.],
                [1.]]),
         array([[1.11022302e-16],
                [7.07106781e-01],
                [7.07106781e-01]])] )
```
When it exists, the witness state returned is given as an ordered list of column vectors defining a local state for the respective party. The product state resulting from tensoring the local states together is orthogonal to all product states of the input collection. 

In the example above the output witness is (within numerical precision) given by the two local states $|\phi_0\rangle = |2\rangle$ and $|\phi_1\rangle = \frac{1}{\sqrt{2}}(|1\rangle + |2\rangle)$ that produce the product state

```math
|\phi\rangle = \tfrac{1}{\sqrt{2}} |2\rangle \otimes \left(|1\rangle + |2\rangle \right),
```
which is orthogonal to all input states considered.

Note that the when a witness state exists it in general not unique, and the implementation used here only returns one particular reproducible state.

This function was adapted from QETLAB [Joh16]_.

### References
==========

..  [Ben99] Bennett, Charles H., et al.
    "Unextendible product bases and bound entanglement."
    Physical Review Letters 82.26 (1999): 5385.
    https://arxiv.org/abs/quant-ph/9808030
.. [Joh16] Nathaniel Johnston.
    "QETLAB: A MATLAB toolbox for quantum entanglement"
    http://www.qetlab.com

:raises ValueError: If list elements are not of type `numpy.ndarray` with two dimensions and the same number of columns.
:param local_states_list: The list of states to check.
:return: `(True, None)` if states form an a UPB and `(False, wittness)` otherwise.

smtsjhr:
Hi @vprusso ,

My last commit [(3dc65c2)](https://github.com/vprusso/toqito/pull/175/commits/3dc65c22afe64d509834dfefb5b150ed46ba35e4) adds docstrings and tests for the two functions I introduced to generate the Tiles and GenShifts UPBs as a list of local states. I think this commit essentially completes (with documentation and testing) all functionality I introduced pertaining to helper functions and mocking up test examples. Now if desired, it should be easy to refactor via copy/paste and minor modifications to function names and module imports in relevant files!

These two functions are called `tiles_local_state_list()` and `genshifts_local_state_list()`, which are still located in the test file `tests/test_state_props/test_is_unextendible_product_basis.py`.

In this same test file, I have also added unit tests for these two functions. For each function there are two tests (note that for the GenShifts case we use pytest to mark.parametrize multiple tests ranging over the number of parties): 

1.  (Orthogonallity) Checks the global product states of the UPB are mutually orthogonal.
2. (Incompleteness) Checks the global product states do not span the whole global space.

To explain this testing methodolgy, I am simply trying to verify that these states form a "incomplete product basis" as defined by the (Orthogonallity) and (Incompleteness) conditions stated above. We do not actually test for the full UPB property here ... thats what the main purpose of the function `is_unextendible_product_basis()` is for. 

The Incompleteness test is done by using the function `numpy.matrix_rank()` to check the rank of the vector span. If it happens to be the case that the number of product states (of a supposed UPB) are known to be less than the global dimension a priori, then this test might be trivial since the rank of a set of states can never be larger than the number of states.


: )

vprusso:
Thank you for the update, @smtsjhr !

Sincerely appreciate the information you provided on this issue and I hope to have some proper time set aside to factor your contribution into `toqito`. Thank you very much again for the updates!

vprusso:
Closing this the UPB feature was cherry-picked and added in as of https://github.com/vprusso/toqito/pull/328.

Thanks again, @smtsjhr !

PR #176, opened 2023-06-11
Diagnall dominant

elads888:
## Description
modified PR to be up to date, and changed code according to comments.
Hope this is good.

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [ ] Ready to go


tests/test_matrix_props/test_is_diagonally_dominant.py
```diff
@@ -0,0 +1,26 @@
+"""Test is_diagonally_dominant."""
+import numpy as np
+
+from toqito.matrix_props import is_diagonally_dominant
+
+
+def test_is_diagonally_dominant():
```
vprusso, line 7: Two more tests might be good to add here:

When matrix is not square:
```suggestion
def test_is_not_square():
    """Test that non-square matrix returns False."""
    mat = np.array([[1, 2, 3], [4, 5, 6]])
    np.testing.assert_equal(is_diagonally_dominant(mat), False)
```

Cover all possible cases. The tests right now miss the testing branch when `is_strict=False`, which we should cover. Something like the following might be good to ensure that we have proper testing coverage. 
```suggestion
 def test_diagonally_dominant():
    """Check that strict diagonally dominance.
    
    Matrix examples from: https://en.wikipedia.org/wiki/Diagonally_dominant_matrix
    """

    # Diagonally dominant (but not strict)
    mat = np.array([[3, -2, 1], [1, 3, 2], [-1, 2, 4]])
    np.testing.assert_equal(is_diagonally_dominant(mat, is_strict=False), True)
        
    # Non-diagonally dominant
    mat = np.array([[-2, 2, 1], [1, 3, 2], [1, -2, 0]])
    np.testing.assert_equal(is_diagonally_dominant(mat), False)
        
    # Strictly diagonally dominant
    mat = np.array([[-4, 2, 1], [1, 6, 2], [1, -2, 5]])
    np.testing.assert_equal(is_diagonally_dominant(mat, is_strict=True), True)
```

vprusso (changes requested):
Just one minor comment about test coverage, but overall looks really good! Once that's addressed, I think we can move this over the line. Let me know if you have any questions!

elads888:
Hi Vincenct!
Added the new tests, Hope we can move it over the line.

PR #181, opened 2023-06-25
Add bures angle

### LINKED ISSUE #171: Feature: Bures angle
Calculate the Bures angle is a measure of distance between two quantum states [[1](https://en.wikipedia.org/wiki/Bures_metric)].

The Bures angle can be computed as follows:

```py
def bures_angle(rho: np.ndarray, sigma: np.ndarray) -> float:
    """Returns the Bures angle between two density operators."""
    return np.real(np.arccos(np.sqrt(fidelity(rho, sigma))))
```

As this is a metric, the location for this should be in `state_metrics/bures_angle.py`. Ensure that proper tests in `tests/test_state_metrics/test_bures_angle.py` and documentation in `docs/state_metrics.rst` are included.  

[1]: [Wikipedia: Bures angle](https://en.wikipedia.org/wiki/Bures_metric)

epelaaez:
## Description
Add bures angle state metric to close #171 

## Todos
  -  [x] Add bures angle in state metrics
  -  [x] Add test for bures angle
  -  [x] Add bures angle to documentation

## Status
-  [x] Ready to go

### DIFF
--- a/docs/_autosummary/toqito.state_metrics.bures_angle.rst
+++ b/docs/_autosummary/toqito.state_metrics.bures_angle.rst
@@ -1,0 +1,6 @@
+toqito.state\_metrics.bures\_angle
+=====================================
+
+.. currentmodule:: toqito.state_metrics
+
+.. autofunction:: bures_angle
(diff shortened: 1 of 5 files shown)


vprusso (approved):
This is a great contribution, @epelaaez! Thank you for taking the time to contribute to the toqito project, and I hope to see you here again soon! Cheers.

Issue #180, opened 2023-06-24
DevOps: Integrate linting within GitHub Actions pipeline


vprusso:
The GitHub Actions (as specified [here](https://github.com/vprusso/toqito/actions)) should be configured to include a linting step that is run automatically. 

In addition, when configuring and running the linter locally, any warnings or errors should be addressed prior to incorporating it within the pipeline.

FrankXenarthra:
@vprusso I'd like to follow up on this. I checked with black which could already reformat automatically 35 of the python files. As a separate commit I could setup black in Github Actions. Do you agree with that approach? Many thanks for the clarification!

vprusso:
Thank you, @FrankXenarthra ! And yes, that approach seems like the right one to me. By all means, have at it and let me know if I can provide any support. Thank you again for your contribution and interest in the project!

purva-thakre:
Just wanted to comment that an auto-formatter like `black` will re-format something requested by a linter like `flake8`. 

If you use `black` on all files in the project and accept all of these changes, it might be incompatible with `flake8`. I did try using it in this manner a few months ago.

FrankXenarthra:
@vprusso Thank you for the friendly welcome. I created this PR https://github.com/vprusso/toqito/pull/188 Could you have a look if the suggestions all make sense or if I should adjust anything? Many thanks!

purva-thakre:
> Edit: Thank you @purva-thakre . I've just seen your comment. I will have to check how to combine those two. Apart from flake8, what are the other dependencies?

Off the top of my head, I don't remember them all right now. I figured out these incompatibilities because I have been thinking of how to create automated Actions workflows for this repo.

vprusso:
There should be a way to add additional content in the `pyproject.toml` file to deviate from the `black` defaults if they conflict with the `flake8` specifications. 

I don't believe we are doing anything super custom with `flake8`, so I imagine the bulk of how `black` jives with `flake8` should be (for the most part) just fine. 

Does that seem to align with what you both are aware of / have seen @purva-thakre  / @FrankXenarthra ?

Thanks again, and do let me know if there are any additional questions or anything I can help out with. Cheers!

FrankXenarthra:
@vprusso @purva-thakre  According to https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html the rule E203 is not PEP8 compilant so it should be ignored to be compatible to black.

I added the following to the pyproject.toml but flake8 ignores it:

[...]

PR #189, opened 2023-09-01
Ensure compatibility of black and flake8 (#180)

FrankXenarthra:
The config file for flake8has been added. The rule E203 is now excluded to ensure compatibility of black and flake8.


vprusso:
Awesome, thank you for providing this one as well!

PR #215, opened 2023-10-31
Test Python 3.12

purva-thakre:
## Description
Provide a brief description of the PR's purpose here.

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [ ] Ready to go

### DIFF
--- a/.github/workflows/build-test-actions.yml
+++ b/.github/workflows/build-test-actions.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         # Add versions here to expand support matrix
-        python-version: ["3.10", "3.11"]
+        python-version: ["3.10", "3.11", "3.12"]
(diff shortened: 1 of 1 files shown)


purva-thakre:
Closing this for now. Won't be able to check this until numpy allows 3.12.

image

vprusso:
As Python 3.12 is rather new, it might take a few weeks for some of the packages like `numpy` etc. to be fully supported. I think this is a worthwhile thing to do, but perhaps we can put this on the backburner for the time being (instead of closing the issue). Any thoughts, @purva-thakre ?

purva-thakre:
@vprusso I don't think I had an issue for this. I tried it out on a whim because I noticed a newer version of python was released. But I can make an issue for the future!

PR #223, opened 2023-11-09
Fix test failure

### LINKED ISSUE #221: Mastumoto Fidelity Test Failure
`test_state_metrics/test_matsumoto_fidelity.py::test_matsumoto_fidelity_default` was commented out in https://github.com/vprusso/toqito/commit/c2d8d8961107c1f91197d0616cc9e353238ba5a6 because it failed. The value is supposed to be `0.99` but instead ends up being calculated as `0.89`. 

This failure happened in this repo after the package versions were updated in #212. Similar failure was observed when I used `pip install -e.` in my local virtual environment. If the older version of `pyproject.toml` was used to install `toqito` by `poetry`, there were no failures. 

This issue is to fix the failure and increase test coverage. 

Two possibilities: 

1. Something to do with changes in `cvxpy`: If the input is a `cvxpy` variable, SDP will return the fidelity value. But `rho` in the failing test is not `isinstance(rho, cvxpy.atoms.affine.vstack.Vstack)`.
2. Something to do with changes in `scipy`: `scipy` version was changed in #212. If the input is not a `cvxpy` variable, linear algebra functions from `scipy` are used to return the calculated fidelity.

purva-thakre:
## Description
Fixes #221 

## Todos
  -  [ ] N/A

## Questions
-  [ ] N/A

## Status
-  [x] Ready to go

### DIFF
--- a/tests/test_state_metrics/test_matsumoto_fidelity.py
+++ b/tests/test_state_metrics/test_matsumoto_fidelity.py
@@ -5,13 +5,13 @@ from toqito.state_metrics import matsumoto_fidelity
 from toqito.states import basis
 
 
-#def test_matsumoto_fidelity_default():
-    #"""Test Matsumoto fidelity default arguments."""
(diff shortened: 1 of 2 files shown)


vprusso (approved):
LGTM!

PR #234, opened 2023-11-14
Update pytest coverage for CB trace norm

### LINKED ISSUE #233: Add tests for toqito/channel_metrics/completely_bounded_trace_norm.py: 48-49
Requires a test where the input is not a quantum channel but is CP. 

https://github.com/vprusso/toqito/blob/a929bb0a8130e9efb0e903fe1f690942d29068de/toqito/channel_metrics/completely_bounded_trace_norm.py#L47-L49

purva-thakre:
## Description
Fixes #233 

## Todos
  -  [x] N/A

## Questions
-  [x] N/A
## Status
-  [x] Ready to go

### DIFF
--- a/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
+++ b/tests/test_channel_metrics/test_completely_bounded_trace_norm.py
@@ -1,5 +1,6 @@
 """Tests for completely_bounded_trace_norm."""
 import numpy as np
+import pytest
 
 from toqito.channel_metrics import completely_bounded_trace_norm
 from toqito.channel_ops import kraus_to_choi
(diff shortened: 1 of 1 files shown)


tests/test_channel_metrics/test_completely_bounded_trace_norm.py
```diff
@@ -9,27 +10,31 @@
...
     """The diamond norm of a quantum channel is 1."""
     phi = dephasing(2)
-    np.testing.assert_equal(completely_bounded_trace_norm(phi), 1)
+    assert completely_bounded_trace_norm(phi) == 1
+
+
+def test_cb_trace_norm_CP():
+    """Test for the diamond norm of a CP map."""
+    non_normalized_depolarizing_array = np.eye(4)
+    assert completely_bounded_trace_norm(non_normalized_depolarizing_array) == 4.0
```
purva-thakre, line 20: I did the lazy thing and used an un-normalized quantum channel. Let me know if I should think of other `np.array` examples.

vprusso: Honestly, I think the "lazy" thing here is the right thing to do. Nice!

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
```diff
@@ -9,27 +10,31 @@
...
 
 
 def test_cb_trace_norm_unitaries_channel():
-    """The diamond norm of phi = id- U id U* is the diameter of the smallest circle that contains the eigenvalues of U."""
+    """The diamond norm of phi = id- U id U* is the diameter of the smallest circle
```
vprusso, line 23: I wonder if the linter is over-aggressively making new lines here when it shouldn't be?

vprusso (approved):
A few minor things, but overall LGTM! I like how the tests are starting to become simpler and are also more specific with the message matching on error output. Nice work!

tests/test_channel_metrics/test_completely_bounded_trace_norm.py
purva-thakre: Probably due to us having different configuration files with incompatible line lengths.

PR #246, opened 2023-11-15
Pytest Coverage: kp_norm

### LINKED ISSUE #238: Add tests for toqito/helper/kp_norm.py: 20
There is no test file for `toqito/helper/kp_norm.py` in `tests/test_helper`. But even then, only L20 is flagged as uncovered by tests instead of both L20 and L23. 

https://github.com/vprusso/toqito/blob/f47dea627bbfcd4d2f232c88d83da1d4e3539b60/toqito/helper/kp_norm.py#L19-L23

### LINKED ISSUE #245: Refactor: The kp_norm function should live in matrix_props
Right now the `kp_norm` file lives in `helper/` but it should live in `matrix_props/` where other such norm calculations reside as well.

purva-thakre:
## Description
Fixes #245 

Fixes #238 

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [x] Ready to go


purva-thakre:
Pylint has flagged some issues after some changes in 2fd4daed47abedc755a522c19e19b4d7465c2449. This PR could be used to correct these. 

```
************* Module tests.test_state_props.test_schmidt_rank
tests/test_state_props/test_schmidt_rank.py:44:4: W0621: Redefining name 'e_0' from outer scope (line 9) (redefined-outer-name)
tests/test_state_props/test_schmidt_rank.py:44:9: W0621: Redefining name 'e_1' from outer scope (line 9) (redefined-outer-name)
tests/test_state_props/test_schmidt_rank.py:58:4: W0621: Redefining name 'e_0' from outer scope (line 9) (redefined-outer-name)
tests/test_state_props/test_schmidt_rank.py:58:9: W0621: Redefining name 'e_1' from outer scope (line 9) (redefined-outer-name)
```

vprusso:
> Pylint has flagged some issues after some changes in [2fd4dae](https://github.com/vprusso/toqito/commit/2fd4daed47abedc755a522c19e19b4d7465c2449). This PR could be used to correct these.
> 
> ```
> ************* Module tests.test_state_props.test_schmidt_rank
> tests/test_state_props/test_schmidt_rank.py:44:4: W0621: Redefining name 'e_0' from outer scope (line 9) (redefined-outer-name)
> tests/test_state_props/test_schmidt_rank.py:44:9: W0621: Redefining name 'e_1' from outer scope (line 9) (redefined-outer-name)
> tests/test_state_props/test_schmidt_rank.py:58:4: W0621: Redefining name 'e_0' from outer scope (line 9) (redefined-outer-name)
> tests/test_state_props/test_schmidt_rank.py:58:9: W0621: Redefining name 'e_1' from outer scope (line 9) (redefined-outer-name)
> ```

Whoops, that is completely my fault. Thank you for flagging and catching that, @purva-thakre !

tests/test_matrix_props/test_kp_norm.py
```diff
@@ -0,0 +1,36 @@
...
+
+import re
+
+import numpy as np
+import pytest
+
+from toqito.matrix_props import kp_norm, trace_norm
+from toqito.states import bell
+
+
```
vprusso, line 12: We could make this a bit more concise using `pytest`:

```suggestion
@pytest.mark.parametrize("vector, k, p, norm_to_compare", [
    # When (k=1, p= Inf)the kp_norm(vector) is the same as the trace norm (the 1-norm).
    (bell(0), 1, np.inf, 1),
    # When p=2 and k is greater than or equal to one of the input matrice's dimensions, the value calculated is the frobenius norm.
    (np.random.rand(5, 4), 4, 2, 2),
])
def test_kp_norm(vector, k, p, norm)
    calculated_kp_norm = kp_norm(vector, k, p)
    assert calculated_kp_norm == np.linalg.norm(vector, norm_to_compare)
```

(Will need to import `pytest` and am not totally sure if this will work. If not, keeping the original is fine).

tests/test_state_props/test_schmidt_rank.py
```diff
@@ -6,7 +6,7 @@
 from toqito.states import basis, bell
 
 
-e_0, e_1 = np.array([[1], [0]]), np.array([[0], [1]])
+bas_0, bas_1 = np.array([[1], [0]]), np.array([[0], [1]])
```
vprusso, line 9: We usually use `e_0` and `e_1` elsewhere in the code to denote these standard basis vectors. Any reason to change that convention here? I'm fine with either, just curious.

vprusso (approved):
Minor comments, but LGTM! MWR! 🚀

tests/test_state_props/test_schmidt_rank.py
purva-thakre: Pylint flagged them. So, I changed them.

vprusso: Huh, interesting. Curious that it doesn't complain about this convention elsewhere. Is it maybe something that's only occurring locally? If not, I wonder if it's worthwhile to set something in the settings as I think I disagree with the linter and think that `e_0` and `e_1` are okay.

purva-thakre: >Is it maybe something that's only occurring locally?

Nope. https://github.com/vprusso/toqito/actions/runs/6872626793/job/18691377546#step:6:14

>If not, I wonder if it's worthwhile to set something in the settings 

Not sure. I'll check.

vprusso: Ah, looking at the error, yeah I think I know why. I think we just want the definition of `e_0` and `e_1` at the top of the file (and then remove the definitions of `e_0` and `e_1` elsewhere in the file. That should fix this issue!

purva-thakre: Stange. My first attempt at fixing this was doing what you just said I try. At the time, tests were failing. But now, they work. :shrug:

vprusso: If I had a nickel for every time something like that happened to me, I'd be a wealthy man 💰 

Glad it works now!

PR #249, opened 2023-11-17
Removed redundant and not needed iden file.

vprusso:
- Removed the `iden.py` file from `matrices/`.
- This was really just an ugly placeholder for what is already provided by both `sympy` and `numpy`. 
- Removed corresponding tests and document generation content for `iden` as well.


purva-thakre (approved):
LGTM! Feel free to merge.

PR #257, opened 2023-11-18
Restructuring tests.

### LINKED ISSUE #256: Restructure tests to be in same directory as functions
Right now we have:

```
tests
└── test_action
    ├── __init__.py
    └── test_action.py
toqito
├── action
│   ├── __init__.py
│   └── action.py
```

however, it would probably be better if we had the following:

```
toqito
├── action
│   ├── __init__.py
│   └── action.py
└── test
    ├── __init__.py         # also make test a package
    └── test_antigravity.py
```

vprusso:
Closes #256 

- As discussed, tests are now contained in a folder within the module in question. 
- Had to change "random/" to "rand/" to avoid naming conflicts with built-in `random` module.


vprusso:
Pretty sure the failure of the build here is because the GitHub action script is pointing to the old test structure. I believe if we update the paths to look in the respective `tests/` folders, that should fix it. 
https://github.com/vprusso/toqito/blob/master/.github/workflows/build-test-actions.yml

vprusso:
Okay, think we are good. Plus there's a boost in test coverage which is nice.

purva-thakre (approved):
That was quick! LGTM. :)

PR #272, opened 2023-11-22
#259-Generate set of mutually unbiased bases

### LINKED ISSUE #259: Feature: Generate a set of d-dimensional mutually unbiased bases (for prime d)
This feature would allow the user to generate a set of d-dimensional [mutually unbiased bases](https://en.wikipedia.org/wiki/Mutually_unbiased_bases). Note that this procedure only works when the dimension "d" is prime (otherwise, it is unknown how to generate such a set in general).

In order to generate such a set, we will require a helper function to determine if a given integer is a prime power:
```py
import numpy as np
from toqito.matrices import gen_pauli
from sympy import isprime
```

```py
def is_prime_power(n: int) -> bool:
    """Determine if number is prime power."""
    p = 2
    primes = []
    for i in range(2, n + 1):
        if isprime(i):
            primes.append(i)
    for i in primes:
        if n % i == 0:
            c = 0
            while n % i == 0:
                n //= i
                c += 1
            if n == 1:
                return True
            return False
```

Putting the above together, we can write our function:

```py
def mubs(dim: int) -> list[np.ndarray]:
    """Function to generate list of MUBs for a given dimension.
    
    Note that this function only works if the dimension provided is prime 
    or a power of a prime. Otherwise, we don't know how to generate general MUBs.
    """
    
    # The first basis will always be the standard basis:
    mats = [np.eye(dim)]

    pauli_x = gen_pauli(1, 0, dim)
    if isprime(dim):
        pauli_z = gen_pauli(0, 1, dim)

        for j in range(dim, 0, -1):
            _, eigen_vec = np.linalg.eig(pauli_x @ pauli_z**(j))
            mats.append(eigen_vec)
    elif is_prime_power(dim) and not isprime(dims):
        raise(f"Dimension {dim} is a prime power but not prime (more complicated no support atm).")
    else:
        raise(f"No general construction of MUBs is known for dimension: {dim}.")

    mubs: list[np.ndarray] = []
    for i in range(len(mats)):
        nrows, ncols = mats[i].shape[0], mats[i].shape[1]
        for row in range(nrows):
            mub: list[np.ndarray] = []
            for col in range(ncols):
                mub.append(mats[i][col][row])
            mubs.append(np.array(mub))
    return mubs
```

While most of the above does the heavy lifting, we still require unit tests and proper documentation.

vprusso:
Closes #259 

- Feature to generate a MUB in prime dimensions
- Improving function to check if set of vectors is a MUB
- Tests for both MUB and check if MUB.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ coveralls = "^1.9"
 setuptools = ">65.5.1"
 sphinx = "6.2.1"
 sphinx_rtd_theme = "*"
-sympy = "^1.12"
+sympy = "==1.12"
 
 
 [build-system]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ coveralls = "^1.9"
 setuptools = ">65.5.1"
 sphinx = "6.2.1"
 sphinx_rtd_theme = "*"
+sympy = "^1.12"
 
 
 [build-system]
--- a/toqito/state_props/is_mutually_unbiased_basis.py
+++ b/toqito/state_props/is_mutually_unbiased_basis.py
@@ -6,7 +6,7 @@ from typing import Any
 import numpy as np
 
 
-def is_mutually_unbiased_basis(vec_list: list[np.ndarray | list[float | Any]]) -> bool:
+def is_mutually_unbiased_basis(vectors: list[np.ndarray | list[float | Any]]) -> bool:
     r"""
     Check if list of vectors constitute a mutually unbiased basis [WikMUB]_.
 
@@ -80,17 +80,25 @@ def is_mutually_unbiased_basis(vec_list: list[np.ndarray | list[float | Any]]) -
     :return: :code:`True` if :code:`vec_list` constitutes a mutually unbiased basis, and
              :code:`False` otherwise.
     """
-    if len(vec_list) <= 1:
+    if len(vectors) <= 1:
         raise ValueError("There must be at least two bases provided as input.")
 
-    dim = vec_list[0][0].shape[0]
-    for i, item in enumerate(vec_list):
-        for j, item2 in enumerate(vec_list):
+    num_vectors = len(vectors)
+    dim = vectors[0].shape[0]
+
+    # We expect the number of vectors to be a multiple of the dimension.
+    if num_vectors % dim != 0:
+        return False
+
+    num_bases = num_vectors // dim
+
+    # Check the inner product between vectors from different bases.
+    for i in range(num_bases):
+        for j in range(i + 1, num_bases):
             for k in range(dim):
-                if i != j:
-                    if not np.isclose(
-                        np.abs(np.inner(item[k].conj().T[0], item2[k].conj().T[0])) ** 2,
-                        1 / dim,
-                    ):
+                for l in range(dim):
+                    # Compute inner product between vectors from different bases.
+                    inner_product = np.abs(np.vdot(vectors[i * dim + k], vectors[j * dim + l])) ** 2
+                    if not np.isclose(inner_product, 1 / dim):
                         return False
     return True
--- a/toqito/state_props/is_mutually_unbiased_basis.py
+++ b/toqito/state_props/is_mutually_unbiased_basis.py
@@ -78,9 +78,6 @@ def is_mutually_unbiased_basis(vectors: list[np.ndarray | list[float | Any]]) ->
     :return: :code:`True` if :code:`vec_list` constitutes a mutually unbiased basis, and
              :code:`False` otherwise.
     """
-    if len(vectors) <= 1:
-        raise ValueError("There must be at least two bases provided as input.")
-
     num_vectors = len(vectors)
     dim = vectors[0].shape[0]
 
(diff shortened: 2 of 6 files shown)


toqito/state_props/is_mutually_unbiased_basis.py
```diff
@@ -80,17 +78,25 @@ def is_mutually_unbiased_basis(vec_list: list[np.ndarray | list[float | Any]]) -
     :return: :code:`True` if :code:`vec_list` constitutes a mutually unbiased basis, and
              :code:`False` otherwise.
     """
-    if len(vec_list) <= 1:
+    if len(vectors) <= 1:
         raise ValueError("There must be at least two bases provided as input.")
```
purva-thakre, line 82: Should we make an issue for checking what happens when `is_check_something` has been provided an empty array as an input? 

For the current function, I think the output will be `true` if 2 or more empty arrays are provided as input. In the tests, I do see 1 function checking for the `ValueError` on line 82.

toqito/states/mutually_unbiased_basis.py
```diff
@@ -0,0 +1,82 @@
...
+    .. math::
+        M_0 = \left\{|0\rangle, |1\rangle \right\}, \\
+        M_1 = \left\{\frac{|0\rangle + |1\rangle}{\sqrt{2}}, \frac{|0\rangle - |1\rangle}{\sqrt{2}}\right\}
+        M_2 = \left\{\frac{|0\rangle + i|1\rangle}{\sqrt{2}}, \frac{|0\rangle - i|1\rangle}{\sqrt{2}}\right\}
+
+    References
+    ==========
+    .. [WikMUB] Wikipedia: Mutually unbiased bases
+        https://en.wikipedia.org/wiki/Mutually_unbiased_bases
+
```
purva-thakre, line 28: Same as what I said in `state_props/is_mutually_unbiased_basis`. 

I might be wrong but I don't think this reference has been cited in this function's docstring. So, using `docname in docnames` won't automatically generate a reference.

toqito/states/mutually_unbiased_basis.py
```diff
@@ -0,0 +1,82 @@
...
+    pauli_x = gen_pauli(1, 0, dim)
+    if isprime(dim):
+        pauli_z = gen_pauli(0, 1, dim)
+
+        for j in range(dim, 0, -1):
+            _, eigen_vec = np.linalg.eig(pauli_x @ pauli_z**(j))
+            mats.append(eigen_vec)
+    elif _is_prime_power(dim) and not isprime(dim):
+        raise ValueError(f"Dimension {dim} is a prime power but not prime (more complicated no support atm).")
```
purva-thakre, line 44:
```suggestion
        raise ValueError(f"Dimension {dim} is a prime power but not prime (more complicated no support at the moment).")
```
Minor nit.

toqito/states/tests/test_mutually_unbiased_basis.py
```diff
@@ -0,0 +1,49 @@
...
+@pytest.mark.parametrize("dim", [2, 3, 5, 7])
+def test_mutually_unbiased_basis(dim):
+    np.testing.assert_equal(is_mutually_unbiased_basis(mutually_unbiased_basis(dim)), True)
+
+
+@pytest.mark.parametrize("dim", [4, 8, 9])
+def test_mutually_unbiased_basis_prime_power_not_prime(dim):
+    """Dimension is a prime power but not prime (this is not presently supported)."""
+    with pytest.raises(
+        ValueError, match=re.escape(f"Dimension {dim} is a prime power but not prime (more complicated no support atm).")
```
purva-thakre, line 20:
```suggestion
        ValueError, match=re.escape(f"Dimension {dim} is a prime power but not prime (more complicated no support at the moment).")
```

toqito/states/mutually_unbiased_basis.py
```diff
@@ -0,0 +1,82 @@
...
+    else:
+        raise ValueError(f"No general construction of MUBs is known for dimension: {dim}.")
```
purva-thakre, line 46:
```
        raise ValueError(f"No general construction of MUBs is known for dimension: {dim}.")
```
Might want to add a link to your repo in this function's docstring or in this `ValueError` message.  

https://github.com/vprusso/qubit-upb

purva-thakre (approved):
LGTM except for some minor changes.

pyproject.toml
```diff
@@ -63,6 +63,7 @@ coveralls = "^1.9"
 setuptools = ">65.5.1"
 sphinx = "6.2.1"
 sphinx_rtd_theme = "*"
+sympy = "^1.12"
```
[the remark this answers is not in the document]
vprusso: Absolutely, I'm 100% on board with that. The caret indeed I think could lead to some subtle issues depending on whether things change on the `sympy` side of things. Updated!

toqito/states/mutually_unbiased_basis.py
vprusso: Ah, these are similarly named but different mathematical objects; this function is for mutually unbiased bases while the qubit UPB repo is unextendible product bases.

toqito/states/mutually_unbiased_basis.py
vprusso: Ah, let me add that citation. Good call!

toqito/state_props/is_mutually_unbiased_basis.py
vprusso: Hmm, actually, I'm not sure this check is necessary. A vector of length 1 isn't necessarily an error but is just not a MUB. I've removed this check and adapted the tests accordingly. Thanks for catching that!

Issue #273, opened 2023-11-22
Feature: Check if matrix is "totally positive"


vprusso:
It would be nice to add a feature to check whether a given matrix is "[totally positive](https://en.wikipedia.org/wiki/Totally_positive_matrix)".

One initial attempt at checking this in Python is given by:
```py
import numpy as np
from itertools import combinations

def is_totally_positive(X, sub_sizes=None, tol=None):
    """
    Determines whether a matrix is totally positive.

    Args:
    X: A square matrix.
    sub_sizes: List of sizes of submatrices to consider. Default is all sizes up to min(X.shape).
    tol: Tolerance for considering values as positive. Default is max(X.shape) * eps.

    Returns:
    itp: 1 if the matrix is totally positive, 0 otherwise.
    wit: The submatrix that is not totally positive, if any.
    """
    sX = X.shape
    wit = 0

    if sub_sizes is None:
        sub_sizes = range(1, min(sX) + 1)
    if tol is None:
        tol = max(sX) * np.finfo(X.dtype).eps * np.linalg.norm(X, 'fro')

    for j in sub_sizes:
        # Handle 1x1 determinants separately
        if j == 1:
            r, c = np.where(np.minimum(np.real(X), -np.abs(np.imag(X))) < -tol)
            if r.size > 0:
                return 0, (r[0], c[0])

        # Handle larger determinants
        else:
            sub_ind_r = list(combinations(range(sX[0]), j))
            sub_ind_c = list(combinations(range(sX[1]), j)) if sX[0] != sX[1] else sub_ind_r

            for kr in sub_ind_r:
                for kc in sub_ind_c:
                    d = np.linalg.det(X[np.ix_(kr, kc)])
                    if d < -tol or abs(np.imag(d)) > tol:
                        return 0, (kr, kc)

    return 1, wit

# Example usage
X = np.array([[1, 2], [3, 4]])
itp, wit = is_totally_positive(X)
print("Is Totally Positive:", itp)
print("Witness:", wit)
```

Would need to add examples for the docs/references, and also tests to see that it properly flags both totally positive and non-totally positive matrices.

PR #274, opened 2023-11-22
Adding is totally positive feature.

vprusso:
Closes: #273 

- Adding feature to check if matrix is "totally positive"
- Tests for checking this functionality.

### DIFF
--- a/toqito/matrix_props/tests/test_is_totally_positive.py
+++ b/toqito/matrix_props/tests/test_is_totally_positive.py
@@ -1,0 +1,33 @@
+"""Test is_totally_positive."""
+import numpy as np
+import pytest
+
+from toqito.matrix_props import is_totally_positive
+
+
+@pytest.mark.parametrize("mat, tol, sub_sizes, expected_result", [
+    # 2x2 matrix that is known to be totally positive.
+    (np.array([[1, 2], [2, 5]]), 1e-6, None, True),
+    # 2x2 matrix that is not totally positive.
+    (np.array([[1, -2], [3, 4]]), 1e-6, None, False),
+    # 3x3 that is totally positive.
+    (np.array([[1, 2, 3], [2, 5, 8], [3, 8, 14]]), 1e-6, None, True),
+    # 3x3 that is not totally positive
+    (np.array([[1, 2, 3], [-1, -2, -3], [2, 5, 8]]), 1e-6, None, False),
+    # Matrix with complex entries (which should not be totally positive).
+    (np.array([[1 + 1j, 2], [3, 4]]), 1e-6, None, False),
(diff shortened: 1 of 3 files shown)


toqito/matrix_props/tests/test_is_totally_positive.py
```diff
@@ -0,0 +1,33 @@
...
+    # Test specifying custom sizes for submatrices.
+    (np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]), 1e-6, [1, 2], False),
+    # Test specifying a custom tolerance level.
+    (np.array([[1, 2], [3, 4 - 1e-10]]), 1e-9, None, False),
+    # Test a matrix that is just a single row or column.
+    (np.array([[1, 2, 3]]), 1e-6, None, True),
+    # Test the identity matrix, which should be totally positive.
+    (np.identity(3), 1e-6, None, True),
+    # Empty matrix is considered to be totally positive.
+    (np.array([]), 1e-6, None, True),
```
purva-thakre, line 30:
```
    (np.array([]), 1e-6, None, True),
```
Shouldn't this input raise an error instead of saying the input `is_totall_positive`?

vprusso: The concept of a "totally positive matrix" typically applies to matrices with real entries where all minors (determinants of all square submatrices) are positive. For an empty matrix, which has no entries and therefore no minors, the definition doesn't strictly apply.

However, in a mathematical context, one might argue that since there are no minors to violate the condition of being positive, the empty matrix could technically be considered totally positive by default. This interpretation aligns with certain conventions in mathematics where vacuous truths are considered true (i.e., a statement that is true because it doesn't contain any instances that make it false).

So, IOW, I think so long as we are clear that the empty matrix is considered totally positive, that should be okay. Thoughts?

purva-thakre: Sure, in a mathematical context, I do agree with what you are saying. 

As this is a defined function, suppose it was used in another function's definition which has a bug. It would not alert the user that their function is not working as intended because somewhere in the middle they started working with an empty array when they had originally intended to work with a non-empty array. 

I think we should not return either `True` or `False` but instead return a `ValueError`. Either way, I leave this up to whatever you think is appropriate.

purva-thakre (approved):

toqito/matrix_props/tests/test_is_totally_positive.py
vprusso: Hmm, yeah I agree with this now. I've gone ahead and updated the logic and tests such that an empty matrix is flagged and raises a `ValueError`. 

Thanks again for the feedback, @purva-thakre !

PR #278, opened 2023-11-24
Coverage for realignment.

### LINKED ISSUE #237: Add tests for toqito/channels/realignment.py: 68
Check error is raised in tests. It might also be better to expand on the `ValueError` message. 

https://github.com/vprusso/toqito/blob/f47dea627bbfcd4d2f232c88d83da1d4e3539b60/toqito/channels/realignment.py#L67-L68

vprusso:
Closes #237 

- Removed value error invalid dim check as I don't believe this case can be triggered.
- Slight cleanup to reformatting of test file.


purva-thakre (approved):
Merge when ready!

pylint has flagged a wanring related the changes here: 

```
************* Module channels.realignment
toqito/channels/realignment.py:57:4: W0612: Unused variable 'eps' (unused-variable)
```

toqito/channels/realignment.py
```diff
@@ -64,8 +64,6 @@ def realignment(input_mat: np.ndarray, dim: int | list[int] = None) -> np.ndarra
 
     if isinstance(dim, int):
         dim = np.array([int(dim), int(dim_mat[0] / dim)])
-        if np.abs(dim[1] - np.round(dim[1])) >= 2 * dim_mat[0] * eps:
```
purva-thakre, line 67: What was this line originally supposed to check for?

vprusso: This function was adapted from QETLAB, so it was put in from here:

```
https://github.com/nathanieljohnston/QETLAB/blob/4f50dbf6ccd4a71c99d4f1afdcd85e9bfda5e9ac/Realignment.m#L38
```

TBH though, I'm not sure of a way to be able to generate such an example that triggers.

PR #280, opened 2023-11-24
Remove `pydocstyle` + Add `ruff`

### LINKED ISSUE #277: Alternative to `pydocstyle`
It looks like `pydocstyle` [will not be maintained anymore.](https://github.com/PyCQA/pydocstyle) 

This issue is to remove `pydocstyle` from github workflows + replace it with something else. 

This might be a good alternative: https://github.com/PyCQA/flake8-docstrings

This issue is also related to some things discussed in #201

purva-thakre:
## Description
Fixes #277

All the available rules are available at this link: https://docs.astral.sh/ruff/rules/

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [ ] Ready to go


vprusso:
Good call on `ruff` by the way! I've seen that pop up a couple times elsewhere and I quite like that project. Nice, @purva-thakre !

.github/workflows/build-test-actions.yml
```diff
@@ -28,10 +28,8 @@ jobs:
       - name: Run pylint
         run: poetry run pylint --exit-zero toqito/
```
purva-thakre, line 29: Might be better to remove this `pylint` check as `ruff` lets us control what is checked based on the available rules. 

https://docs.astral.sh/ruff/rules/#pyflakes-f

The workflow finally flagged lines>120 as errors due to `ruff`.

vprusso: Yep, perfect, let's kill the `pylint` line.

And while that's probably going to lead to lots of flagged lines, it's great that it's catching! Think we can use `black` to auto-format the lines of offending files for length to automate that process.

purva-thakre: I think we should change our mind about removing `pylint`. :grimacing:

I just realized `ruff` does not implement all the `pylint` rules yet. https://github.com/astral-sh/ruff/issues/970

For example, the following earnings are raised by pylint but not by ruff. 

image

vprusso: Good call. All for that! 👍

toqito/matrix_ops/inner_product.py
```diff
@@ -36,7 +36,7 @@ def inner_product(v1: np.ndarray, v2: np.ndarray) -> float:
     :param v1: v1 and v2, both vectors of dimenstions :math:`(n,1)` where :math:`n>1`.
     :param v2: v1 and v2, both vectors of dimenstions :math:`(n,1)` where :math:`n>1`.
     :return: The computed inner product.
-    """
+    """  # noqa: E501
```
vprusso, line 39: Out of curiouisity I wonder why this is being flagged here. It's for a line that's too long, but there doesn't seem to be anything too long here?

toqito/matrix_props/is_totally_positive.py
```diff
@@ -19,7 +20,7 @@ def is_totally_positive(mat, tol=1e-6, sub_sizes=None):
         \begin{pmatrix}
             1 & 2 \\
             3 & 4
-        \end{pmatrix}
+        \\end{pmatrix}
```
vprusso, line 23: Hmm, is there are reason there are two `\\` here?

vprusso (commented):
Putting in a few comments. Not sure if this is ready for review of not.

toqito/matrix_ops/inner_product.py
purva-thakre: It's not flagged there. This is not ignored if it is added inside a docstring. For now, I asked `ruff` to ignore that docstring due to lines 12 and 20.

toqito/matrix_props/is_totally_positive.py
purva-thakre: For some reason, `isort` or `black` made this change. I'll check it out later.

purva-thakre:
Closing this branch for now due to too many local conflicts. Will create a new branch and PR.

PR #292, opened 2023-11-28
Fixing bug in state exclusion code.

vprusso:
Had a silly bug that I patched for `state_exclusion`.


vprusso:
Thanks for flagging those! Fixed and merging now. Thanks for the quick review, @purva-thakre !

PR #286, opened 2023-11-26
Attempt 2 for Ruff

purva-thakre:
## Description
Fixes #277 and what was started in #280

## Todos
Notable points that this PR has either accomplished or will accomplish.

Some things flagged by isort and pycodestyle have already been completed. Following are the things flagged by pydocstyle. 

  -  [x] toqito/matrix_ops/tests/test_inner_product.py:16:5: D103 Missing docstring in public function
  - [x] toqito/matrix_ops/tests/test_outer_product.py:14:5: D103 Missing docstring in public functio
  - [x] toqito/matrix_ops/tests/test_outer_product.py:24:5: D103 Missing docstring in public function
  - [x] toqito/matrix_ops/tests/test_unvec.py:14:5: D103 Missing docstring in public function
  - [x] toqito/matrix_ops/tests/test_vec.py:12:5: D103 Missing docstring in public function
  - [x] toqito/matrix_ops/tests/test_vectors_from_gram_matrix.py:43:5: D103 Missing docstring in public function
  - [x] toqito/matrix_ops/tests/test_vectors_to_gram_matrix.py:24:5: D103 Missing docstring in public function
  - [x] ~~toqito/matrix_props/tests/test_is_orthonormal.py:8:5: D404 First word of the docstring should not be "This"~~
  - [x] ~~toqito/matrix_props/tests/test_is_orthonormal.py:19:5: D404 First word of the docstring should not be "This"~~
  - [x] toqito/matrix_props/tests/test_is_totally_positive.py:30:5: D103 Missing docstring in public function
  - [x] toqito/matrix_props/tests/test_is_totally_positive.py:38:5: D103 Missing docstring in public function
  - [x] toqito/matrix_props/tests/test_kp_norm.py:24:5: D103 Missing docstring in public function
  - [x] toqito/perms/tests/test_antisymmetric_projection.py:29:5: D103 Missing docstring in public function
  - [x] toqito/rand/tests/test_random_density_matrix.py:19:5: D103 Missing docstring in public function
  - [x] toqito/rand/tests/test_random_density_matrix.py:33:5: D103 Missing docstring in public function
  - [x] toqito/rand/tests/test_random_state_vector.py:20:5: D103 Missing docstring in public function
  - [x] toqito/rand/tests/test_random_unitary.py:10:5: D103 Missing docstring in public function
  - [x] toqito/rand/tests/test_random_unitary.py:20:5: D103 Missing docstring in public function
[...]

Issue #293, opened 2023-11-28
Add tests for toqito/state_props/is_separable.py:  77->84, 114->128, 136-188


purva-thakre:
https://github.com/vprusso/toqito/blob/70ecd264508824d0991fb954d41020a7b9a915a2/toqito/state_props/is_separable.py#L77-L81

Above `ValueError` can be removed. 

https://github.com/vprusso/toqito/blob/70ecd264508824d0991fb954d41020a7b9a915a2/toqito/state_props/is_separable.py#L114-L125

https://github.com/vprusso/toqito/blob/70ecd264508824d0991fb954d41020a7b9a915a2/toqito/state_props/is_separable.py#L135-L188

Yash-10:
#612 added more conditions in `is_separable`, but the test coverage needs to be improved. In particular, these lines are currently untested:

```
93->100, 127->139, 148, 159-182, 189-253, 273->295, 291, 297-306, 310
```

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L93-L100

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L127-L139

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L148

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L159-L182

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L189-L253

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L273-L295

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L291

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L297-L306

https://github.com/vprusso/toqito/blob/3d20a16d506f63f934555efc5cdc6865ddd45ffe/toqito/state_props/is_separable.py#L310

vprusso:
[...]

Issue #240, opened 2023-11-14
Add tests for toqito/matrix_ops/tensor.py: 140->144


purva-thakre:
Last `if` is not covered by existing tests. 

https://github.com/vprusso/toqito/blob/f47dea627bbfcd4d2f232c88d83da1d4e3539b60/toqito/matrix_ops/tensor.py#L140-L144

Might also be better to use `pytest.parametrize` to condense some of the existing tests.

PR #298, opened 2023-11-28
Add `pydcostyle` + `pylint` in `ruff` rules

purva-thakre:
## Description
Fixes #286 

## Todos

Edit: Removed checklist. 

## Questions
-  [ ] Question1

## Status
-  [ ] Ready to go

### DIFF
--- a/docs/articles.bib
+++ b/docs/articles.bib
@@ -105,6 +105,14 @@
   url = {https://link.aps.org/doi/10.1103/Cabello_2002_NParticle}
 }
 
+@misc{Chen_2003_Matrix,
+      title={A matrix realignment method for recognizing entanglement}, 
+      author={Kai Chen and Ling-An Wu},
+      year={2003},
+      eprint={quant-ph/0205017},
+      archivePrefix={arXiv},
+      primaryClass={quant-ph}
+}
 
 @article{Chen_2014_Symmetric,
    title={Symmetric extension of two-qubit states},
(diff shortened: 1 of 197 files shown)


purva-thakre:
@vprusso I am choosing to ignore `PLR2004` because https://github.com/vprusso/toqito/pull/286/commits/2841708da37d52aa9589c186deaf4c48db271121 led to a drop in code coverage. 

I think it's Ok to ignore this because I personally can't understand the need to define `dim = 2` to replace ` if dim_something < 2 :` with `if dim_something < dim :`

If you would prefer follow this rule, I can create an issue to remove this rule from `ignore` after #293, #294 and #240 are completed.

vprusso:
> @vprusso I am choosing to ignore `PLR2004` because [2841708](https://github.com/vprusso/toqito/commit/2841708da37d52aa9589c186deaf4c48db271121) led to a drop in code coverage.
> 
> https://docs.astral.sh/ruff/rules/magic-value-comparison/
> 
> I think it's fine to ignore this because I personally can't understand the need to define `dim = 2` to replace ` if dim_something < 2 :` with `if dim_something < dim :`
> 
> If you would prefer follow this rule, I can create an issue to remove this rule from `ignore` after #293, #294 and #240 are completed.

Right! I was kind of thinking the same thing. I mean, I *guess* it often does make sense to not have "magic numbers" floating around. However, checking if something is equal to a given value like `if X < 0` is flagged seems a bit crazy!

Agreed, and thank you for making that change!

vprusso (approved):
Went through the changes and LGTM! MWR! 🚀

PR #319, opened 2023-12-01
Pytest: `toqito/matrix_ops/tensor.py`

### LINKED ISSUE #239: Add tests for toqito/helper/npa_hierarchy.py: 91->90
One of the hierarchy levels is not covered by tests. 

https://github.com/vprusso/toqito/blob/f47dea627bbfcd4d2f232c88d83da1d4e3539b60/toqito/helper/npa_hierarchy.py#L90-L93

### LINKED ISSUE #240: Add tests for toqito/matrix_ops/tensor.py: 140->144

purva-thakre:
## Description
Fixes #240  

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [ ] Ready to go

### DIFF
--- a/toqito/helper/tests/test_npa_hierarchy.py
+++ b/toqito/helper/tests/test_npa_hierarchy.py
@@ -60,7 +60,7 @@ def test_cglmp_inequality(k):
     objective = cvxpy.Maximize(i_b)
     problem = cvxpy.Problem(objective, npa)
     val = problem.solve()
-    np.testing.assert_equal(np.allclose(val, 2.914, atol=1e-3), True)
+    assert abs(val - 2.914) <= 1e-3
 
 
 @pytest.mark.parametrize("k, expected_size", [("1+a", 9), ("1+ab", 25)])
@@ -78,4 +78,4 @@ def test_cglmp_dimension(k, expected_size):
     for variable in problem.variables():
         if variable.name() == "R":
             r_size = variable.shape[0]
-    np.testing.assert_equal(np.allclose(r_size, expected_size), True)
+    assert r_size == expected_size
--- a/toqito/helper/tests/test_npa_hierarchy.py
+++ b/toqito/helper/tests/test_npa_hierarchy.py
@@ -60,7 +60,7 @@ def test_cglmp_inequality(k):
     objective = cvxpy.Maximize(i_b)
     problem = cvxpy.Problem(objective, npa)
     val = problem.solve()
-    assert abs(val - 2.914) <= 1e-3
+    assert pytest.approx(val, 1e-3) == 2.914
 
 
 @pytest.mark.parametrize("k, expected_size", [("1+a", 9), ("1+ab", 25)])
--- a/toqito/matrix_ops/tensor.py
+++ b/toqito/matrix_ops/tensor.py
@@ -141,7 +141,7 @@ def tensor(*args) -> np.ndarray:
             result = args[0][0]
             for i in range(1, len(args[0])):
                 result = np.kron(result, args[0][i])
-        return result
+            return result
 
     # Tensor product one matrix `n` times with itself.
     if len(args) == 2 and isinstance(args[1], int):
--- a/toqito/matrix_ops/tensor.py
+++ b/toqito/matrix_ops/tensor.py
@@ -119,7 +119,7 @@ def tensor(*args) -> np.ndarray:
     result = None
 
     # Input is provided as a list of numpy matrices.
-    if len(args) == 1 and isinstance(args[0], list):
+    if (len(args) == 1 and isinstance(args[0], list)) or (len(args) == 1 and isinstance(args[0], np.ndarray)):
         if len(args[0]) == 1:
             return args[0][0]
         if len(args[0]) == 2:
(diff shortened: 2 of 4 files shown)


purva-thakre:
@vprusso Just an FYI : `test coverage` is going to fail for a while because % will drop in value. 

`pytest` had been also checking `test_something` lines after you changed the directory structure. Before `coverage` was expected to ignore `tests/` directory.

toqito/matrix_ops/tensor.py
```diff
@@ -141,7 +141,7 @@ def tensor(*args) -> np.ndarray:
             result = args[0][0]
             for i in range(1, len(args[0])):
                 result = np.kron(result, args[0][i])
-        return result
+            return result
```
purva-thakre, line 145: @vprusso I think these 3 if conditions are not properly defined. The intended input in this case is supposed to be a numpy array inside a list. 

The if condition at L140 is not triggered because the first if at L136 also might be triggered by the same set of conditions. Should I go ahead and remove L140-144?

vprusso: I think so, yeah. If you do, do any of the test cases fail?

purva-thakre:
@vprusso This is ready for a review. 

I got rid of the `.shape` argument because I did not think that was really useful. 

I also got rid of a  bunch of `if` conditions. Let me know if I missed some use case that could have been useful here.

toqito/helper/tests/test_npa_hierarchy.py
```diff
@@ -60,7 +60,7 @@ def test_cglmp_inequality(k):
     objective = cvxpy.Maximize(i_b)
     problem = cvxpy.Problem(objective, npa)
     val = problem.solve()
-    np.testing.assert_equal(np.allclose(val, 2.914, atol=1e-3), True)
+    assert abs(val - 2.914) <= 1e-3
```
vprusso, line 63: Hmm, I am wondering why this change needed to be made. I was under the impression that the `atol` parameter would check what you're checking here, namely that the difference between `val` and 2.914 is within 3 decimal places.

toqito/matrix_ops/tests/test_tensor.py
```diff
@@ -5,174 +5,64 @@
 from toqito.matrix_ops import tensor
 from toqito.states import basis
 
+e_0, e_1 = basis(2, 0), basis(2, 1)
+matrix1 = np.array([[1, 2]])
```
vprusso, line 9: Since these appear to only be used in one place, I wonder if we just instead define these as inline values as done in the rest of the tests. So, for instance:

```py
    (np.array([matrix1, matrix2, matrix3, matrix4], dtype=object), 1, np.kron(
        np.kron(matrix1, np.kron(matrix2, matrix3)), matrix4)),
```

would become:

```py
    (np.array([np.array([[3], [4]]), np.array([[3], [4]]), np.array([[5, 6]]), np.array([[7, 8]])], dtype=object), 1, np.kron(
        np.kron(np.array([[3], [4]]), np.kron(np.array([[3], [4]]), np.array([[5, 6]]))),np.array([[7, 8]]))),
```

toqito/matrix_ops/tests/test_tensor.py
```diff
@@ -5,174 +5,64 @@
...
+    # tensor product of array of 4 arrays of identity matrices
+    (np.array([np.identity(2), np.identity(2), np.identity(2), np.identity(2)]), 1, np.identity(16)),
+    # tensor product with a numpy array containing three or more matrices
+    (np.array([matrix1, matrix2, matrix3, matrix4], dtype=object), 1, np.kron(
+        np.kron(matrix1, np.kron(matrix2, matrix3)), matrix4)),
+    # tensor product of 1 matrix inside a list
+    ([np.array([np.identity(4)])], 1, np.identity(4))])
+def test_tensor_multiple_input(test_input, len_input, expected):
+    """Test function works as expected."""
+    if len_input == 2:
```
vprusso, line 52: [nit]: Seems a bit odd that the case for `len_input == 2` precedes `len_input == 1`.

vprusso (approved):
Nice work! I think you've also made the code in the `tensor.py` file much less confusing.  Modulo the minor comments, MWR!

toqito/matrix_ops/tests/test_tensor.py
purva-thakre: Yeah, I thought about that but then your suggestion also makes that particular parameter unreadable. Where does the parentheses begin, stop etc.

toqito/helper/tests/test_npa_hierarchy.py
purva-thakre: I initially checked for both being equal to each other. Then, realized had to account for the tolerance value. So, used this.  

I am actually trying to undo the current style where you check for something then also make sure that something's check was true. The new changes to that line while using numpy's testing functionality would be:

```
np.testing.assert_equal(abs(val-2.914), 1e-3)
```

vprusso: I think you can also use `rtol` without the `abs()` function as this just checks for the relative tolerance. Or, the way you have it here is also fine!

PR #324, opened 2023-12-03
Adding check for linear independence.

### LINKED ISSUE #142: Feature: Check for linearly independent / dependent vectors
Given a collection of vectors, determine if they are linearly independent or not. Something like the following:

```python
import numpy as np
from toqito.matrix_ops import vec

def is_linearly_independent(vectors) -> bool:
    """Determine if all of the states in the ensemble are linearly independent."""
    vecs = tuple([vec(vector) for vector in vectors])
    mat = np.array(vecs).T

    return np.alltrue(np.linalg.matrix_rank(mat) == len(vecs))
```

Adding tests and documentation for the feature would also be required (as it is done for the other functions in `toqito` as well.

vprusso:
Closes #142

### DIFF
--- a/docs/_autosummary/toqito.matrix_props.is_linearly_independent.rst
+++ b/docs/_autosummary/toqito.matrix_props.is_linearly_independent.rst
@@ -1,0 +1,6 @@
+﻿toqito.matrix\_props.is\_linearly\_independent
+==============================================
+
+.. currentmodule:: toqito.matrix_props
+
+.. autofunction:: is_linearly_independent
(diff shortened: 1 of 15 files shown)


purva-thakre:
Something I recently caught wrt to the new additions, we do want to make sure we add them to the documentation. It's up to you if you want to use this PR for all of the previously merged PRs or create a separate PR. 

this PR, #327, #274, #272

purva-thakre (approved):

vprusso:
> Something I recently caught wrt to the new additions, we do want to make sure we add them to the documentation. It's up to you if you want to use this PR for all of the previously merged PRs or create a separate PR.
> 
> this PR, #327, #274, #272

Oof, very good call, thank you for flagging this! 

I also missed `trine` and `is_antidistinguishable` as well. Added those (as well as the ones for `is_linearly_independent`). 

Good call!

PR #335, opened 2023-12-06
Missing tests: `is_block_positive`

### LINKED ISSUE #241: Add tests for toqito/matrix_props/is_block_positive.py: 62, 66->70, 70->78, 73, 109

purva-thakre:
## Description
Related to some uncovered lines in #241

Last line is still not covered by this PR.  

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [x] Ready to go

### DIFF
--- a/toqito/matrix_props/is_block_positive.py
+++ b/toqito/matrix_props/is_block_positive.py
@@ -45,7 +45,6 @@ def is_block_positive(
         :filter: docname in docnames
 
 
-    :raises ValueError: If `dim` is a scalar, it must evenly divide the length of the matrix.
     :raises RuntimeError: Unable to determine k-block positivity. Please consider increasing the relative tolerance or
                             the effort level.
     :param mat: A bipartite Hermitian operator.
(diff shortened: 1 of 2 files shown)


purva-thakre:
To do: Search for open ended discussions in published refs for below. 

https://github.com/vprusso/toqito/blob/2226c46ddfb1d4e2c263ff7b02ebb9723717a718/toqito/matrix_props/is_block_positive.py#L109-L111

vprusso:
> @vprusso Do you mind explaining the use-case scenario for these lines? I am struggling with defining the input matrix and dims for this case.
> 
> https://github.com/vprusso/toqito/blob/2226c46ddfb1d4e2c263ff7b02ebb9723717a718/toqito/matrix_props/is_block_positive.py#L71-L75

All I can really say is that I believe this was an artifact of the way in which the same type of thing was checked in the QETLAB implementation. I'm not sure I have a more satisfying impression of this at the moment. There was a similar check in the `realignment.py` code as well, as I ended up removing it as I wasn't able to even construct a test case that would hit this block. 

Not sure if that's overly zealous or if perhaps there's a reason why this makes sense to do in MATLAB vs. Python. So we could either remove it, or, keep it for now and expect that it will remain uncovered.

vprusso (approved):
🚀

PR #372, opened 2023-12-14
Fix pytest bug

### LINKED ISSUE #367: Pytest coverage bug
`pytest` is supposed to skip all `tests` sub-folders when it calculates coverage %. 

But some of the tests do show up with `missing` tests. This issue is to make sure `omit` lines in `coveragerc` are entered correctly. 

image

https://github.com/vprusso/toqito/blob/0121c6926da6d00d14958853eff36607a40547a5/.coveragerc#L11-L14

purva-thakre:
## Description
Fixes #367

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [x] Ready to go

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -10,6 +10,8 @@ show_missing = True
 precision = 2
 omit =
         *__init__*
+        */tests/*
+        *tests/*
         */tests*
         *tests*
         virtualenvs/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -10,6 +10,7 @@ show_missing = True
 precision = 2
 omit =
         *__init__*
+        */tests*
         *tests*
         virtualenvs/*
 


.coveragerc
```diff
@@ -10,6 +10,9 @@ show_missing = True
 precision = 2
 omit =
         *__init__*
+        */tests/*
+        *tests/*
+        */tests*
         *tests*
```
purva-thakre, line 16: I don't know which of these lines worked to fix the bug. I propose we keep these.

vprusso:
Hmm. Any thoughts as to why the code coverage fell here?

purva-thakre:
> Hmm. Any thoughts as to why the code coverage fell here?

Some test files were being counted in the code coverage. Now, they are not.

vprusso:
> Some test files were being counted in the code coverage. Now, they are not.

Ah, of course. I guess it's odd to me that they were counted at all in the first place. I think this LGTM and can be pushed in (ignoring the codecov thing).

vprusso (approved):

Issue #362, opened 2023-12-12
Failure of docstring examples in `matrix_props`


purva-thakre:
To replicate the errors listed below, see the workflow in the description of #360 

- [ ] toqito/matrix_props/is_projection.py

```
026                 0 & 1 \\
027                 0 & 1
028             \end{pmatrix}
029 
030     our function indicates that this is indeed a projection matrix.
031 
032     >>> from toqito.matrix_props import is_projection
033     >>> import numpy as np
034     >>> A = np.array([[0, 1], [0, 1]])
035     >>> is_projection(A)
Expected:
    True
Got:
    False

```

- [ ] toqito/matrix_props/majorizes.py

```
032     :math:`\text{Tr}_{\mathcal{B}}(\rho)` majorizes
033     :math:`\text{Tr}_{\mathcal{A}}(\rho)`.
034 
035     >>> from toqito.matrix_props import majorizes
036     >>> from toqito.states import max_entangled
037     >>> from picos import partial_trace
038     >>>
039     >>> v_vec = max_entangled(3)
040     >>> rho = v_vec * v_vec.conj().T
041     >>> majorizes(partial_trace(rho, [1]), rho)

UNEXPECTED EXCEPTION: TypeError('The expression has shape 9×9 so it cannot be decomposed into subsystems of shape 2×2.')

```

PR #378, opened 2023-12-16
Fixing bug in projection matrix function.

vprusso:
Closes: #362 

- Actually, there was a bug in this function that was indirectly caught in the docstring, so that's good!


purva-thakre (approved):
Was the `is_positive_semidefinite` added due to a special use case scenario?

vprusso:
> Was the `is_positive_semidefinite` added due to a special use case scenario?

Yep, exactly, that was my fault!

Issue #399, opened 2023-12-29
Failure on macOS


purva-thakre:
As noted in #374, `toqito.perms` module might not work as intended on `macOS`.  The following test fails on `macOS` which is why it is skipped in a Github workflow for `macOS`. 

https://github.com/vprusso/toqito/blob/8606650b98608330c8b89414f7fb641992517ee4/toqito/perms/tests/test_antisymmetric_projection.py#L20-L23

To fix this issue: Check functions from `toqito.perms` in a `macOS` virtual machine or an actual `macOS` device.  

This could possibly be an issue in `scipy` and/or `numpy`. If this is the case, recreate the issue with a simple reproducible example without using functions in `toqito`. For more info, see https://github.com/numpy/numpy/issues/25395#issuecomment-1865066958)

## Failure Message
```
FAILED toqito/perms/tests/test_antisymmetric_projection.py::test_antisymmetric_projection[3-3-True-expected_result3] - IndexError: boolean index did not match indexed array along dimension 0; dimension is 1 but corresponding boolean dimension is 27
```

Bchass:
If no one has taken interest in this issue and is not part of a  project, I'd like to take this one on.

I use a Mac on a daily basis.

vprusso:
Another pair of eyes would be great if you're willing to spare them on this, @Bchass !

Thank you, as always, for your interest and support!

Issue #63, opened 2021-05-19
Enhancement: Prefer `np.testing.array_equal` over `assertEqual` in testing


vprusso:
As pointed out by @theRoughCode in https://github.com/vprusso/toqito/issues/44, the `np.testing.array_equal` pattern should be preferred over the `assertEqual` pattern in the unit testing code. The reason for this preference is that the former provides more helpful testing data from failed instances in comparison with the latter. This task will involve finding instances of `assertEqual` and altering those to `np.testing.array_equal` where appropriate.

theRoughCode:
Another thing to consider when fixing this is that [this line](https://github.com/vprusso/toqito/blob/834549e6b0d667777927ce50bb2bff3b28bae8d9/tests/test_nonlocal_games/test_nonlocal_game.py#L53) might not be testing what was intended:
```python
        # It may be possible for the lower bound to not be equal to the correct
        # quantum value (as it may get stuck in a local minimum), but it should
        # never be possible for the lower bound to attain a value higher than
        # the true quantum value.
        self.assertLessEqual(np.isclose(res, np.cos(np.pi / 8) ** 2, rtol=1e-02), True)
```

what would end up here is:
```python
 self.assertLessEqual(True, True)
```

Instead, what we want might be something like this:
```python
        expected = np.cos(np.pi / 8) ** 2
        self.assertLessEqual(res, expected)
        np.testing.assert_allclose(res, expected, rtol=1e-02)
```

However, I tried this but the "less than or equal to" check failed because `res` was slightly greater than `expected`:
```
E       AssertionError: 0.8535545972243008 not less than or equal to 0.8535533905932737
```

@vprusso Should we be truncating the values to a certain decimal place and then performing this check?

purva-thakre:
We have decided to switch to using `pytest` instead of `np.testing`. Closing this as #488 has more details on what to do.

PR #423, opened 2024-01-12
macos failure fix

Bchass:
## Description
Resolves: https://github.com/vprusso/toqito/issues/399

Same issue: https://github.com/scikit-learn/scikit-learn/issues/10562 tests were failing only for `macos` with `np.testing.assert_allclose`.

https://github.com/numpy/numpy/issues/7726 discussion on why numpy [informs](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) users that `allclose` is the equivalent to `np.testing.assert_allclose` when they provide different default values.

Another discussion: https://github.com/numpy/numpy/issues/3183#issuecomment-350892896

Based on both of these discussions, not much came out of them. It really comes down to maintainers to set the path forward for issues like this. It's hard to say what the core issue is exactly. ~~One can say that it's on `numpy/scikit` but on the other hand one can say that it's an issue with `macos`.~~

From what I've read so far surrounding this issue, it's an issue with numpy. Not so much a bug though, more of a decision the maintainers of numpy made a long time ago.

## Questions
-  [x] For future tests going forward, what should be recommended ? `assert np.allclose` or `np.testing.assert_allclose`? It's a bit unknown when it comes to `macos`.

    - `assert np.allclose` unfortunately does not give the same amount of detail as `np.testing.assert_allclose`.

## Status
-  [x] Ready to go

### DIFF
--- a/.github/workflows/build-test-actions.yml
+++ b/.github/workflows/build-test-actions.yml
@@ -33,10 +33,6 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           sudo apt-get install -y libblas-dev liblapack-dev
-      - name: Install blas, lapack for macOS
-        if: runner.os == 'macOS'
-        run: |
-          brew install openblas lapack
       - name: Install Toqito + dependencies
         run: |
           python -m pip install --upgrade pip
--- a/toqito/perms/tests/test_antisymmetric_projection.py
+++ b/toqito/perms/tests/test_antisymmetric_projection.py
@@ -29,4 +29,4 @@ anti_proj_3_3_partial[21] = 0.40824829
 def test_antisymmetric_projection(dim, p_param, partial, expected_result):
     """Test function works as expected for a valid input."""
     proj = antisymmetric_projection(dim=dim, p_param=p_param, partial=partial).todense()
-    assert np.allclose(proj, expected_result)
+    assert abs(proj - expected_result).all() <= 1E-3
--- a/toqito/perms/tests/test_antisymmetric_projection.py
+++ b/toqito/perms/tests/test_antisymmetric_projection.py
@@ -1,5 +1,4 @@
 """Test antisymmetric_projection."""
-import platform
 
 import numpy as np
 import pytest
@@ -17,9 +16,6 @@ anti_proj_3_3_partial[15] = -0.40824829
 anti_proj_3_3_partial[19] = -0.40824829
 anti_proj_3_3_partial[21] = 0.40824829
 
-# https://docs.python.org/3/library/platform.html
-# Darwin is the system name for macOS
-@pytest.mark.skipif(platform.system() == "Darwin", reason="3-3-True-expected_result3 fails for macOS")
 @pytest.mark.parametrize("dim, p_param, partial, expected_result", [
     # Dimension is 2 and p is equal to 1.
     (2, 1, False, np.array([[1, 0], [0, 1]])),
@@ -33,4 +29,4 @@ anti_proj_3_3_partial[21] = 0.40824829
 def test_antisymmetric_projection(dim, p_param, partial, expected_result):
     """Test function works as expected for a valid input."""
     proj = antisymmetric_projection(dim=dim, p_param=p_param, partial=partial).todense()
-    np.testing.assert_allclose(proj, expected_result)
+    assert np.allclose(proj, expected_result)
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -5,10 +5,6 @@ Getting started
 ===============
 
 .. warning::
-    We recommend :code:`toqito` be installed only on a Linux machine. There is a 
-    `known issue <https://github.com/vprusso/toqito/issues/399#issue-2060205963>`_ with a module in :code:`toqito` not
-    working as intended on macOS.
-
     Efficacy of :code:`toqito` has not been verified on Windows. 
 
 ----------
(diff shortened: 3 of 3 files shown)


purva-thakre:
Great find @Bchass ! I did not think it would be this quick.

To answer your question, another option is to use `assert abs(test_result - expected_result).all() <= 1E-3` or something like `assert (test_result = expected_result).all()`. Could you please check if either of the two cause any failures for macOS? 

I usually rely on using `.all()`.

I haven't had a chance to read the linked discussions in `numpy` and `scikit`. I'll add more related to this later. 

Note: The description of this PR is probably something we want to keep in mind for #63

Bchass:
Will do! I'll do some more testing with what you recommended later tonight or sometime during this long weekend.

Bchass:
https://github.com/numpy/numpy/issues/7726#issuecomment-225308392

More of the discussion around this topic that took place a long time ago.

vprusso:
Excellent sleuthing, @Bchass! Great find! I'm all good for this one to be merged. All good on your end too, @purva-thakre ?

purva-thakre:
>I'm all good for this one to be merged

@vprusso How about we wait a little bit to check failures on the different ways to check for numpy array equality? We need to standardize this across all tests. Right now, tests use different options depending on who added the test. It's possible there are other options for this that have not been used by us.

I want to search those options and have @Bchass test these out since they have access to a `macOS` device.

Bchass:
I agree with @purva-thakre to wait a bit. Want to break into some other tests. The interesting part is the test never failed locally. Only with GitHub Actions.

Bchass:
> To answer your question, another option is to use `assert abs(test_result - expected_result).all() <= 1E-3` or something like `assert (test_result = expected_result).all()`. Could you please check if either of the two cause any failures for macOS?

`assert abs(test_result - expected_result).all() <= 1E-3` passes locally and remotely. See: https://github.com/Bchass/toqito/pull/3

`assert (test_result = expected_result).all()` fails locally.

This is only for this one test case. I'll try some more test cases after.

purva-thakre:
>The interesting part is the test never failed locally on a mac. Only with GitHub Actions.

@Bchass There was a similar issue in MItiq due to some changes I introduced to the code to satisfy `mypy`. The test would fail on `macos`. https://github.com/unitaryfund/mitiq/pull/2127

@natestemen did a great job of looking for the source of the issue. His view as to why the tests passed locally on macos but not in Github actions was due to the differences in the resources available in both scenarios.

purva-thakre:
> `assert abs(test_result - expected_result).all() <= 1E-3` passes locally and remotely. See: [Bchass#3](https://github.com/Bchass/toqito/pull/3)
> 
> `assert (test_result = expected_result).all()` fails locally.

This is really weird though. I thought `.all()` would fail evenly for both scenarios. Maybe I should look more into the linked `numpy` discussions. Thanks again!

Bchass:
Thanks for the link to that issue. Maybe this is following in the same realm? I'm going to explore more and see if I can dig further into why `np.testing.assert_allclose` only fails remotely and maybe get it to pass finally.

purva-thakre:
@Bchass Could you please check if the following also pass or fail randomly?

- https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html
- https://numpy.org/doc/stable/reference/generated/numpy.array_equiv.html
- https://numpy.org/doc/stable/reference/generated/numpy.allclose.html
- https://numpy.org/doc/stable/reference/generated/numpy.equal.html
- https://docs.pytest.org/en/latest/reference/reference.html#pytest.approx

So far, I am leaning towards using `assert abs(test_result - expected_result).all() <= 1E-3` across all tests. But we'll wait for @vprusso's comment on this. Seems like comparing the difference of the elements of two arrays requires fewer resources than comparing the equality of the two.

Bchass:
> @Bchass Could you please check if the following also pass or fail randomly?
> 
> * https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html
> * https://numpy.org/doc/stable/reference/generated/numpy.array_equiv.html
> * https://numpy.org/doc/stable/reference/generated/numpy.allclose.html
> * https://numpy.org/doc/stable/reference/generated/numpy.equal.html
> * https://docs.pytest.org/en/latest/reference/reference.html#pytest.approx
> 
> So far, I am leaning towards using `assert abs(test_result - expected_result).all() <= 1E-3` across all tests. But we'll wait for @vprusso's comment on this. Seems like comparing the difference of the elements of two arrays requires fewer resources than comparing the equality of the two.
> 
> If we use `allclose`, then we will have to make sure there are no weird approximation issues.

All pass locally/remotely. `assert_allclose` fails due to specific indices given in the test case. A few days ago `assert_allclose` passed in my remote branch when supplying whole numbers and not using the specific indices.

purva-thakre:
Then, let's stick to using `assert abs(test_result - expected_result).all() <= 1E-3`. 

@vprusso Thought on this?

.github/workflows/build-test-actions.yml
```diff
@@ -33,10 +33,6 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           sudo apt-get install -y libblas-dev liblapack-dev
-      - name: Install blas, lapack for macOS
-        if: runner.os == 'macOS'
-        run: |
-          brew install openblas lapack
```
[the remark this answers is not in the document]
Bchass: >As a dependency for many of the solvers, you will need to ensure you have the BLAS and LAPACK mathematical libraries installed on your machine. If you have numpy working on your machine (installed as a toqito dependency), it is likely that you already have these libraries on your machine.

`macos` already ships with BLAS and LAPACK installed by default under the Accelerate framework: https://developer.apple.com/documentation/accelerate/blas/

Numpy will install them by default if your machine doesn't have them, specifically: https://www.openblas.net/

toqito/perms/tests/test_antisymmetric_projection.py
```diff
@@ -33,4 +29,4 @@
 def test_antisymmetric_projection(dim, p_param, partial, expected_result):
     """Test function works as expected for a valid input."""
     proj = antisymmetric_projection(dim=dim, p_param=p_param, partial=partial).todense()
-    np.testing.assert_allclose(proj, expected_result)
+    assert np.allclose(proj, expected_result)
```
purva-thakre, line 32:
```suggestion
    assert abs(proj - expected_result).all() <= 1E-3
```

Bchass: > So far, I am leaning towards using assert abs(test_result - expected_result).all() <= 1E-3 across all tests

Are you still leaning towards this direction? Or just for `test_antisymmetric_projection`?

purva-thakre: Let's make these changes on a case-by-case basis. It will be a very big PR if we do this for all-in-one PR. 

Plus, before we make these changes for all tests, we should discuss #63 with @vprusso.

vprusso: Yeah, I'd (selfishly) opt for a more piecemeal approach as I think my small brain can only handle MRs of a certain size :)

purva-thakre:
Going to go ahead and merge this PR. Thanks @Bchass !

PR #425, opened 2024-01-15
State distinguishability improvement

### LINKED ISSUE #420: Real-valued objective function for quantum state distinguishability
Quantum state distinguishability objective function must be real-valued. That is, in `state_opt/state_distinguishability.py` alter this line:

```py
    objective = cvxpy.Maximize(sum(obj_func))
```

to

```py
    objective = cvxpy.Maximize(cvxpy.real(sum(obj_func)))
```

vprusso:
Closes #420


toqito/state_opt/state_distinguishability.py
```diff
@@ -69,61 +72,83 @@ def state_distinguishability(
...
-
-    :param states: A list of states provided as either matrices or vectors.
-    :param probs: Respective list of probabilities each state is selected.
-    :param dist_method: Method of distinguishing to use.
-    :return: The optimal probability with which Bob can distinguish the state.
+    :param vectors: A list of states provided as vectors.
+    :param probs: Respective list of probabilities each state is selected. If no
+                  probabilities are provided, a uniform probability distribution is assumed.
+    :param solver: Optimization option for `picos` solver. Default option is `solver_option="cvxopt"`.
+    :param primal_dual: Option for the optimization problem.
```
purva-thakre, line 87:
```suggestion
    :param primal_dual: Option for the optimization problem. Default option is `"dual"`.
```

toqito/state_opt/state_distinguishability.py
```diff
@@ -69,61 +72,83 @@ def state_distinguishability(
     >>> e_11 = e_1 * e_1.conj().T
     >>> states = [e_00, e_11]
     >>> probs = [1 / 2, 1 / 2]
-    >>> res = state_distinguishability(states, probs)
+    >>> res, _ = state_distinguishability(states, probs)
```
purva-thakre, line 75: I think it might be better to add a few examples that show off the different solver options: `primal_dual`. A user has the choice of `dual` or `primal`. 

This might be a better option as the private functions  `_min_error_primal` and ` _min_error_dual` are not going to show up in the docs build.

Also, for the existing example, make it obvious somehow that we are using the default option of `dual`.

toqito/state_opt/state_distinguishability.py
```diff
@@ -1,12 +1,15 @@
...
-from .state_helper import __is_probs_valid, __is_states_valid
+from toqito.state_ops import pure_to_mixed
 
 
 def state_distinguishability(
-    states: list[np.ndarray], probs: list[float] = None, dist_method: str = "min-error"
+    vectors: list[np.ndarray],
+    probs: list[float] = None,
+    solver: str = "cvxopt",
+    primal_dual: str = "dual",
```
purva-thakre, line 12: Since you have removed `dist_method`, you should remove the options for it from the docstring. 

Stuff under `math::` will also change due to this.

toqito/state_opt/tests/test_state_distinguishability.py
```diff
@@ -1,92 +1,34 @@
...
+    assert abs(val - expected_result) <=1e-8
+
+
+@pytest.mark.parametrize("vectors, probs, solver, primal_dual", [
+    # Bell states (default uniform probs with dual).
+    ([bell(0), bell(1), bell(2), e_0], None, "cvxopt", "dual"),
+])
+def test_state_distinguishability_invalid_vectors(vectors, probs, solver, primal_dual):
+    """Test function works as expected for an invalid input."""
+    with pytest.raises(ValueError):
```
purva-thakre, line 33:
```suggestion
    with pytest.raises(ValueError, match = "Vectors for state distinguishability must all have the same dimension."):
```
It's better to use `match` because I have noticed instances of pytest thinking a line is uncovered because we did not make it check the error message.

purva-thakre (changes requested):

toqito/state_opt/state_distinguishability.py
vprusso: Yep, added some additional details in the docstring.

toqito/state_opt/state_distinguishability.py
vprusso: Yep, good point. Changed!

vprusso:
Thanks again for the feedback, @purva-thakre 

I believe I have addressed your comments. I have also added a function `is_distinguishable` which functions very similarly to `is_antidistinguishable` (but checks for distinguishability over antidistinguishability) 

This should also close: https://github.com/vprusso/toqito/issues/426

toqito/state_opt/state_distinguishability.py
```diff
@@ -23,17 +26,8 @@ def state_distinguishability(
     Alice chooses :math:`i` with probability :math:`p_i` and creates the state :math:`\rho_i` Bob
     wants to guess which state he was given from the collection of states.
```
purva-thakre, line 28:
```suggestion
    Alice chooses :math:`i` with probability :math:`p_i` and creates the state :math:`\rho_i`. Bob
    wants to guess which state he was given from the collection of states.

```

purva-thakre (approved):
LGTM!

PR #454, opened 2024-02-01
Python versioning

### LINKED ISSUE #444: pylint does not work for python >3.12
As noticed in a recent recent workflow run, one of the checks for code style does not work exactly as expected for python 3.12 because pylint is not compatible with that version.

The runner resorts to using a lower version for python.

image

We will keep this issue open until a new release with `pylint` becomes compatible with >3.12 or all `pylint` rules used by us have been incorporated into `ruff`. 

https://github.com/astral-sh/ruff/issues/970

Interestingly enough, `pylint` only specifies python version as `>=3.8`. So, the message we get could be due to another dependency?

https://github.com/pylint-dev/pylint/blob/c2294becb4adbaa3a5ba1cedd72b3c8995beb359/pyproject.toml#L35

purva-thakre:
## Description
Fixes #444 

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [ ] Ready to go

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ keywords = ["quantum information", "quantum computing", "nonlocal games"]
 
 
 [tool.poetry.dependencies]
-python = ">=3.10,<=3.12"
+python = ">=3.10,<=4"
 cvxpy = "1.4.2"
 cvxopt = "1.3.2"
 more-itertools = "10.2.0"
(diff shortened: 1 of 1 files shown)


purva-thakre:
Going to go ahead and merge this as the issue appears to be resolved. Don't think it needs an additional review.

PR #455, opened 2024-02-01
Linkcheck workflow

### LINKED ISSUE #398: Broken links in documentation
Correct broken links in docs.

If `make linkcheck` is run in `toqito/docs`, some of the links in docs and docstrings appear to be broken. 

- [x] Correct broken links : Fixed by https://github.com/vprusso/toqito/commit/8d0b7f6f98213a2c7703c33843843919f4102b40 in  #392 
- [x] Add a workflow to check this

purva-thakre:
## Description
Fixes #398 

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [ ] Ready to go


docs/conf.py
```diff
@@ -12,6 +12,8 @@
 #
 # import os
 # import sys
+# sys.path.insert(0, os.path.abspath('.'))
```
vprusso, line 15: Should we remove these lines?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ${{ matrix.os }}
+    env:
+      PYTHON: ${{ matrix.python-version }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Toqito + dependencies
```
vprusso, line 30: Nitpick `Toqito` -> `toqito`

.github/workflows/docs.yml
```diff
@@ -0,0 +1,41 @@
...
+    branches:
+    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-to-specific-branches-occurs
+        - master
+
+jobs:
+  broken-links:
+    strategy:
+      matrix:
+          # Add versions here to expand support matrix
+        python-version: ["3.12"]
```
vprusso, line 19: It's fine if we only need to do 3.12 here, but just checking to make sure this shouldn't be expanded to all versions of Python that we support.

vprusso (approved):
A few minor comments, but overall LGTM! 🚀

docs/conf.py
purva-thakre: I want to keep these. 

Apparently, sphinx can have some import issues and this is one of the possible solutions for it. 

I do not know why we did not get those issues. Just in case we end up getting it in the future, I left those in.

.github/workflows/docs.yml
purva-thakre: I don't think `make linkcheck` will work differently for different versions of python. Which is why I chose to have 1 job for this.

PR #459, opened 2024-02-02
Add a workflow to check for examples in docs

### LINKED ISSUE #289: Automate examples in Docs
Right now, all the examples in the documentation are hard-coded such that the output of a function is manually inserted. 

For example, the to-be calculated values of `channel_fidelity` are already in the docstring. 

https://github.com/vprusso/toqito/blob/63ba15d420c3d8f83a87d5de1edf508bb3457d8e/toqito/channel_metrics/channel_fidelity.py#L38-L48

We want to instead move toward calculating these whenever the documentation is built. Might need a new Github Actions workflow. 

https://stackoverflow.com/questions/52965999/best-method-for-limiting-float-precision-comparison-in-python-doctest
https://stackoverflow.com/questions/2428618/how-to-test-floats-results-with-doctest
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html
https://stackoverflow.com/questions/12391390/does-sphinx-run-my-code-on-executing-make-html
https://stackoverflow.com/questions/58189699/use-sphinx-doctest-with-example
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#showing-code-examples

- [ ] Make examples in docs executable: `toqito/channel_metrics`
- [ ] Make examples in docs executable: `toqito/channel_ops`
- [ ] Make examples in docs executable: `toqito/channel_props`
- [ ] Make examples in docs executable: `toqito/channels`
- [ ] Make examples in docs executable: `toqito/helper`
- [ ] Make examples in docs executable: `toqito/matrices`
- [ ] Make examples in docs executable: `toqito/matrix_ops`
- [ ] Make examples in docs executable: `toqito/matrix_props`
- [ ] Make examples in docs executable: `toqito/measurement_ops`
- [ ] Make examples in docs executable: `toqito/measurement_props`
- [ ] Make examples in docs executable: `toqito/nonlocal_games`
- [ ] Make examples in docs executable: `toqito/perms`
- [ ] Make examples in docs executable: `toqito/rand`
- [ ] Make examples in docs executable: `toqito/state_metrics`
- [ ] Make examples in docs executable: `toqito/state_ops`
- [ ] Make examples in docs executable: `toqito/state_opt`
- [ ] Make examples in docs executable: `toqito/state_props`
- [ ] Make examples in docs executable: `toqito/states`

purva-thakre:
## Description
Fixes #289 

## Todos
  -  [ ] TODO 1

## Questions
-  [ ] Question1

## Status
-  [ ] Ready to go


purva-thakre:
To do: If I locally run `make linkcheck` or `make doctest` will add files https://github.com/vprusso/toqito/pull/459/commits/c632cef56abf5a91bad34705945c384eab17986f

purva-thakre:
Doctest is not correctly identifying the examples in the docstrings.

No tests are run in the added job.

image

Locally, I can see multiple failures due to float comparisons. 

Step 1: We want this workflow to fail.

purva-thakre:
> Doctest is not correctly identifying the examples in the docstrings.

This should be fixed with https://github.com/vprusso/toqito/pull/459/commits/9813786b4173a016a7b4da0d3aa05c852a42b51a. Apparently, `make html` has to be used before `make doctest`. 

All the examples are correctly identified now. The issue that I am running into is that the Github Actions environment can't find any of the installed packages.

purva-thakre:
Fixed the import issue. Yay! If `poetry` is used to install a package, have to make sure `poetry run` is used before the command we want to run. This is because `poetry` installs things in a virtual environment.

image

purva-thakre:
@vprusso heads up! This PR is going to have a huge diff. I can't run `make doctest` for 1 folder. Here's a link to all that's failing: https://github.com/vprusso/toqito/actions/runs/7824351650/job/21346779560?pr=459#step:5:2040

Do you want me to ping you everytime I finish up with 1 module?

purva-thakre:
To do: figure out a way to declare a global approximation for all float outputs.

image

toqito/nonlocal_games/xor_game.py
```diff
@@ -70,8 +70,8 @@ class XORGame:
     >>> import numpy as np
     >>> from toqito.nonlocal_games.xor_game import XORGame
     >>> chsh = XORGame(prob_mat, pred_mat)
-    >>> chsh.quantum_value()
-    0.8535533885683664
+    >>> '%.2f' % chsh.quantum_value()
+    '0.85'
```
purva-thakre, line 74: How to clarify why we use `'%.2f' % ` in the output?

This is a workaround to doctest's float failures. With the new changes, docstring expects a string `'0.85'` instead of a float ` 0.8535533885683664`.

vprusso: If you're asking for clarification as to why we truncate, part of the reason for that is the accuracy of the solvers, as they are only able to sometimes get a few digits of accuracy before deviating. Does that answer the question?

purva-thakre: Kinda. I think we might have to have a one liner statement for whenever we use something like `'%.2f' %` in the doc examples. 

I am making `doctest` compare a string in the screenshots but a user does not have to do so. I don't know how to clarify this in every example.

toqito/channel_metrics/diamond_norm.py
```diff
@@ -38,8 +39,8 @@ def diamond_norm(choi_1: np.ndarray, choi_2: np.ndarray) -> float:
     >>> choi_dephasing = dephasing(dim=2)
     >>> choi_identity = np.identity(2**2)
     >>> dn = diamond_norm(choi_dephasing, choi_identity)
-    >>> print("Diamond norm between dephasing and identity channels: ", dn)
-    Diamond norm between depolarizing and identity channels:  -3.5989938811978797e-09
+    >>> print("Diamond norm between dephasing and identity channels: ", '%.2f' % dn)
+    Diamond norm between dephasing and identity channels:  -0.00
```
purva-thakre, line 43: Do we want to keep the precision on L42 and L30? Or the new changes are fine as a demo?

toqito/states/trine.py
```diff
@@ -21,11 +21,12 @@ def trine() -> list[np.ndarray]:
...
-    >>> trine()
+    >>> trine()  # doctest: +ELLIPSIS, +NORMALIZE_WHITESPACE
     [array([[1],
-       [0]]), array([[-0.5      ],
-       [-0.8660254]]), array([[-0.5      ],
-       [ 0.8660254]])]
+            [0]]), array([[-0.5      ],
+            [-0.8660254]]), array([[-0.5      ],
+            [ 0.8660254]])]
+
```
purva-thakre, line 29: I do not know how to pass this test without the doctest flags. Expected vs what doctest got are identical. 

image

toqito/rand/random_circulant_gram.py
```diff
@@ -24,7 +24,8 @@ def random_circulant_gram(dim: int) -> np.ndarray:
     =========
     Generate a random circulant Gram matrix of dimension 4.
 
-    >>> from this_module import random_circulant_gram
+    >>> import numpy as np
+    >>> from toqito.rand import random_circulant_gram
```
purva-thakre, line 28: For all output in the `rand` module, some of the examples output a matrix. 

Is it Ok to make `doctest` skip testing these matrices? https://stackoverflow.com/a/7622355/10241324

purva-thakre:
@vprusso This PR is ready for a review. Feel free to take your time. 

Except for some of the outstanding comments in this PR, there were some last-minute failures that I chose to skip for now. https://github.com/vprusso/toqito/pull/459/commits/6d8ddf15eee061b5f46b9669ef43bacf7dbe23f0 , https://github.com/vprusso/toqito/pull/459/commits/5add9a0f621f0b6c616d7a2d30c6cf18ae0f1ac3

toqito/nonlocal_games/nonlocal_game.py
```diff
@@ -308,8 +308,8 @@ def quantum_value_lower_bound(
         ...                    pred_mat[a_alice, b_bob, x_alice, y_bob] = 1
         >>>
         >>> chsh = NonlocalGame(prob_mat, pred_mat)
-        >>> chsh.quantum_value_lower_bound()
-        0.8535533840915605
+        >>> chsh.quantum_value_lower_bound()   # doctest: +SKIP
```
vprusso, line 311: Should we have the skip here or just truncate the value?

toqito/rand/random_circulant_gram.py
vprusso: Hmm, I wonder if we should make a separate issue to return an `array` from `numpy` instead of a `matrix`. These functions that return a `matrix` object I think are kind of an older artifact of how `numpy` used to do things.

toqito/rand/random_density_matrix.py
```diff
@@ -31,7 +31,7 @@ def random_density_matrix(
 
     >>> from toqito.rand import random_density_matrix
     >>> complex_dm = random_density_matrix(2)
-    >>> complex_dm
+    >>> complex_dm # doctest: +SKIP
```
vprusso, line 34: Does the need for this warning go away if we specify the type of the output?

toqito/rand/random_density_matrix.py
```diff
@@ -46,7 +46,7 @@ def random_density_matrix(
 
     >>> from toqito.rand import random_density_matrix
     >>> real_dm = random_density_matrix(2, is_real=True)
-    >>> real_dm
+    >>> real_dm # doctest: +SKIP
```
vprusso, line 49: Same question as before.

toqito/rand/random_density_matrix.py
```diff
@@ -61,7 +61,7 @@ def random_density_matrix(
 
     >>> from toqito.rand import random_density_matrix
     >>> bures_mat = random_density_matrix(2, distance_metric="bures")
-    >>> bures_mat
+    >>> bures_mat # doctest: +SKIP
```
vprusso, line 64: And here too.

toqito/state_props/is_unextendible_product_basis.py
```diff
@@ -39,7 +39,7 @@ def is_unextendible_product_basis(vecs: list[np.ndarray], dims: list[int]) -> tu
     >>>
     >>> non_upb_tiles = np.array([tile(i) for i in range(4)])
     >>> dims = np.array([3, 3])
-    >>> is_unextendible_product_basis(non_upb_tiles, dims)
+    >>> is_unextendible_product_basis(non_upb_tiles, dims)  # doctest: +ELLIPSIS, +NORMALIZE_WHITESPACE
```
vprusso, line 42: Hmm, are these comments needed here (i.e. the `doctest` stuff)

toqito/states/bb84.py
```diff
@@ -21,12 +21,12 @@ def bb84() -> np.ndarray:
 
     >>> from toqito.states import bb84
     >>> x = bb84()
-    >>> print(f"|0> = {x[0][0].T}, \n |1> = {x[0][1].T}")
-    >>> print(f"|+> = {x[1][0].T}, \n |-> = {x[1][1].T}")
+    >>> print(f"|0> = {x[0][0].T}, \n |1> = {x[0][1].T}")  # doctest: +ELLIPSIS, +NORMALIZE_WHITESPACE
```
vprusso, line 24: Hmm, same here. I'm wondering if we just need to change something to remove this ignore we're doing.

toqito/states/max_mixed.py
```diff
@@ -43,14 +43,15 @@ def max_mixed(dim: int, is_sparse: bool = False) -> [np.ndarray, sparse.dia_matr
...
+           [0. , 0.5]])
+
 
     One may also generate a maximally mixed state returned as a sparse matrix
 
     >>> from toqito.states import max_mixed
-    >>> max_mixed(2, is_sparse=True)
-        <2x2 sparse matrix of type '<class 'numpy.float64'>'
+    >>> max_mixed(2, is_sparse=True) # doctest: +ELLIPSIS, +NORMALIZE_WHITESPACE
+    <2x2 sparse matrix of type '<class 'numpy.float64'>'
```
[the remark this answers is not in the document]
vprusso: Hmm. I see. Wonder if this will not be an issue when/if we switch from `matrix` to `np.array` objects universally.

toqito/states/trine.py
vprusso: It's difficult to know from just the screenshot, but I actually think there is an additional space from the number to the end bracket (you can see how the line outputs are slightly off).

I think you if you copy/paste the "expected" output and put that in the comment, that should fix it (I think).

vprusso (changes requested):
Few minor comments, but overall, looks good!

toqito/rand/random_density_matrix.py
purva-thakre: Not really. 

The output is supposed to be randomly generated density matrix. The expected value can differ over each run of `make doctest`. So, an easier option was to skip this test.

toqito/rand/random_density_matrix.py
purva-thakre: Same as my other reply https://github.com/vprusso/toqito/pull/459#discussion_r1484922829

toqito/rand/random_density_matrix.py
purva-thakre: Same as my other reply https://github.com/vprusso/toqito/pull/459#discussion_r1484922829

toqito/state_props/is_unextendible_product_basis.py
purva-thakre: Yes, otherwise the test would fail even when I made sure I copied & pasted the expected output.

toqito/matrices/cyclic_permutation.py
```diff
@@ -13,26 +13,27 @@ def cyclic_permutation_matrix(n: int, k: int = 1) -> np.ndarray:
...
-    >>> from this_module import cyclic_permutation_matrix
+    >>> from toqito.matrices import cyclic_permutation_matrix
     >>> n = 4
     >>> cyclic_permutation_matrix = cyclic_permutation_matrix(n)
     >>> print(cyclic_permutation_matrix)
-    ([[0, 0, 0, 1],
-    ...[1, 0, 0, 0],
-    ...[0, 1, 0, 0],
-    ...[0, 0, 1, 0]])
+    [[0 0 0 1]
```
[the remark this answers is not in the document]
purva-thakre: Perhaps in a separate issue?

toqito/nonlocal_games/nonlocal_game.py
purva-thakre: The very first thing I tried was truncating the value. Unfortunately, this function was calculating two different values: `0.75 or 0.85` whenever `make doctest` was run. 

Easier option was to make doctest skip this.

toqito/states/bb84.py
purva-thakre: We are not ignoring this output. `ELLIPSIS` will make sure doctest is satisfied with a substring match and `NORMALIZE_WHITESPACE` will make sure doctest doesn't fail the test when the whitespace in the expected and calculated outputs is not identical. 

image

image

toqito/states/max_mixed.py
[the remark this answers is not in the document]
purva-thakre: > Wonder if this will not be an issue when/if we switch from matrix to np.array objects universally.

Don't know.

I thought doctest was being very particular about the whitespace. Maybe 1 is supposed to be a tab while the other is not. `ruff` did not like it when I used tab to create the whitespace.

purva-thakre:
Why are the following tests failing at random? Locally, if I see these 3 failures, I will run `make doctest` again and again until the workflow passes. 

image

image

image

toqito/state_opt/state_distinguishability.py
```diff
@@ -48,22 +48,32 @@ def state_distinguishability(
...
     >>> from toqito.state_opt import state_distinguishability
     >>> states = [bell(0), bell(1), bell(2), bell(3)]
     >>> probs = [1 / 4, 1 / 4, 1 / 4, 1 / 4]
     >>> res, measurements = state_distinguishability(vectors=states, probs=probs, primal_dual="primal")
-    >>> np.around(measurements[0], decimals=5)
-    [[ 0.5+0.j -0. +0.j  0. +0.j  0.5-0.j]
-     [-0. -0.j  0. +0.j -0. -0.j -0. -0.j]
-     [ 0. -0.j -0. +0.j  0. +0.j  0. -0.j]
-     [ 0.5+0.j -0. +0.j  0. +0.j  0.5+0.j]]
+    >>> np.around(measurements[0], decimals=5)  # doctest: +SKIP
```
purva-thakre, line 72: Had to add this `SKIP` here because the py312 job kept failing due to below:

image

vprusso (approved):

PR #472, opened 2024-02-14
Adding package lines for sdist.

### LINKED ISSUE #207: Installing Toqito
Based on the struggle in #204, need to figure out how an editable installation of `toqito` should be packaged. 

Using `pip install -e.` did not install all the required packages. As this is a Python package and pip is popular, it would be good to figure this out especially where is pip getting the info from.

One option would be to use `dev_requirements.txt` created from a working installation via poetry. Then use `pip install dev_reqs`.

Last comment in the above linked issue talks about using `poetry` as a better dependency resolver. But if changes are made to `pyproject.toml` before using `poetry lock (in toqito), poetry install`, the newer version of `pyproject.toml` will not be used. 

https://python-poetry.org/docs/faq/#why-is-the-dependency-resolution-process-slow
https://stackoverflow.com/a/73499097/10241324
https://stackoverflow.com/questions/58961497/how-to-update-poetrys-lock-file-without-upgrading-dependencies
https://stackoverflow.com/questions/58708936/pythonpoetry-installation-is-getting-stuck

```
Installing dependencies from lock file
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.

```

vprusso:
Closes #207 

- Added needed lines to include files for `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,8 @@ setuptools.setup(
     long_description=long_description,
     long_description_content_type="text/markdown",
     url="https://github.com/vprusso/toqito",
-    packages=setuptools.find_packages(),
+    packages=setuptools.find_packages(where="toqito"),
+    package_dir={"": "toqito"},
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Intended Audience :: Science/Research",


purva-thakre (approved):
LGTM! Do we want to do a patch release?

As for the failing doctest, I don't know why those 3 tests fail and pass at random. I should probably create an issue for it.

vprusso:
Yeah, we should probably patch a release in once this merges. Should be able to get to that tomorrow. Thanks for the help in narrowing this one down!

Issue #471, opened 2024-02-13
Output of `toqito/matrices/cyclic_permutation.py` is a 2D list


purva-thakre:

purva-thakre:
[Bchass](https://github.com/Bchass) I can't assign it to you because you don't show up in the suggestions, for some reason.

![image](https://github.com/vprusso/toqito/assets/66048318/cc2222cb-17ea-4068-9afb-470ac3844add)

Let's assume, this is assigned to you.

vprusso:
@purva-thakre yeah, this is a silly thing that Git does where unless the user has commented in the thread, they won't show up. @Bchass if you comment here, you will then show up, and we can subsequently assign the issue to you.

Bchass:
Oh GitHub

vprusso:
> Oh GitHub

Lol, thank you, @Bchass  :)

Bchass:
Do we just want to change on how to call the function? Or add `return np.array` in the code for readability? 

```
>>> from toqito.matrices import cyclic_permutation_matrix
>>> n = 4
>>> k = 3
>>> cyclic_permutation_matrix(n,k)
array([[0, 1, 0, 0],
       [0, 0, 1, 0],
       [0, 0, 0, 1],
       [1, 0, 0, 0]])


>>> cyclic_permutation_matrix = cyclic_permutation_matrix(n, k)
>>> print(cyclic_permutation_matrix)
[[0 1 0 0]
 [0 0 1 0]
 [0 0 0 1]
 [1 0 0 0]]
```

purva-thakre:
I haven't really dug into what the cause of this issue is. There is some inconsistency in the output expected by doctest vs what it got. 

If I were to use something similar to what's shown below for the expected output of `cyclic_permutation_matrix`, `make doctest` runs into a failure because it expects a 2D list not an array. 

https://github.com/vprusso/toqito/blob/4416bce906c307185320cd9fac367c046e28bfb1/toqito/matrices/clock.py#L41-L44

Bchass:
I'll see what else I can come up with

Bchass:
This is definitely not a straight forward one. Even when do I get the output to match what is expected, it still fails. Maybe whitespace? I'm still chugging away on this.

purva-thakre:
>Even when do I get the output to match what is expected, it still fails. Maybe whitespace?

If you create a draft PR, we could talk about this more.

vprusso:
[...]

PR #486, opened 2024-02-23
test expected output

Bchass:
## Description
Fixes https://github.com/vprusso/toqito/issues/471

## Changes

- update examples to try and line up with what doctest expects

## Checklist


purva-thakre:
You might want to rebase your PR. What you're working with is from an older commit.

Also, you might have to make changes to the code of the function to get to `array` output (probably)

image

purva-thakre:
@Bchass You could try to figure out where the function changes a numpy array to a 2D list at each step by making python print the variable type. 

https://github.com/vprusso/toqito/blob/eb5820721884dae5b9fdaad68913b2d96785167d/toqito/matrices/cyclic_permutation.py#L57-L62

For example, temporarily use `print(type(variable_name))` after each line in the function code. If you use the function, you should be able see where the function transitions from manipulating a `np.ndarray` type to a `2D list`. 

Hopefully, this makes sense? Again, I haven't had the chance to dig into the issue yet but this is what my first step would have been.

Bchass:
Thanks for the info! I'll look into it more at some point this weekend.

Bchass:
@purva-thakre Stepping through the function, it never converted to a 2D list. It boiled down to whitespace and how I was calling the function in the examples.

Came across: https://numpy.org/doc/stable/reference/generated/numpy.set_printoptions.html and was able to format the result how `doctest` wanted it.

purva-thakre:
>Stepping through the function, it never converted to a 2D list. It boiled down to whitespace and how I was calling the function in the examples.

That's interesting!

toqito/matrices/cyclic_permutation.py
```diff
@@ -59,4 +56,5 @@ def cyclic_permutation_matrix(n: int, k: int = 1) -> np.ndarray:
     p_mat[0, -1] = 1
 
     result_mat = np.linalg.matrix_power(p_mat, k)
+    np.set_printoptions(edgeitems=n*n, linewidth=40)
```
purva-thakre, line 59: Maybe you can get rid of this line. If the issue was related to using `print(cyclic_permutation_matrix(n, k))` vs `cyclic_permutation_matrix(n, k)`, then this addition is not required as you have corrected these lines in the example.

Bchass: I'm sure I'll be taking on some more!

Bchass:
More failures to investigate tomorrow

purva-thakre:
@Bchass You can ignore the drop in `pytest` coverage. The `linkcheck` workflow passes now, previous failure could have been related to the being down being down or something.

purva-thakre (approved):

Issue #497, opened 2024-03-01
Output of `toqito/matrices/gen_gell_mann.py` is of type `matrix`


purva-thakre:

Bchass:
🙋‍♂️

PR #521, opened 2024-03-28
gen_gell_mann refactor

Bchass:
## Description
Resolves: https://github.com/vprusso/toqito/issues/497

## Changes

  -  [x] Refactor to return array

## Checklist

### DIFF
--- a/toqito/matrices/gen_gell_mann.py
+++ b/toqito/matrices/gen_gell_mann.py
@@ -1,10 +1,13 @@
 """Generalized Gell-Mann matrices."""
 
+
 import numpy as np
 from scipy import sparse
 
 
-def gen_gell_mann(ind_1: int, ind_2: int, dim: int, is_sparse: bool = False) -> np.ndarray | sparse.lil_matrix:
+def gen_gell_mann(
+    ind_1: int, ind_2: int, dim: int, is_sparse: bool = False
+) -> np.ndarray:
     r"""Produce a generalized Gell-Mann operator :cite:`WikiGellMann`.
 
     Construct a :code:`dim`-by-:code:`dim` Hermitian operator. These matrices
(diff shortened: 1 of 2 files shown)


vprusso:
I think this looks good to me. The one failure is one that we already expect (confirming with @purva-thakre). Aside from that, if you think this is ready, @Bchass feel free to merge. Thank you again for your contribution!

vprusso:
(but maybe have @purva-thakre confirm prior to merging :) )

toqito/matrices/gen_gell_mann.py
```diff
@@ -1,10 +1,12 @@
 """Generalized Gell-Mann matrices."""
 
+
 import numpy as np
-from scipy import sparse
 
 
-def gen_gell_mann(ind_1: int, ind_2: int, dim: int, is_sparse: bool = False) -> np.ndarray | sparse.lil_matrix:
```
purva-thakre, line 7: @Bchass Thank you for this PR! Is it fine if we merge this PR next week?

I want to understand the difference between using `scipy.sparse` vs. `np.ndarray`. I am a little bit confused about how the return type was previously `matrix` when the `scipy.sparse` [documentation](https://docs.scipy.org/doc/scipy/reference/sparse.html) says they have moved to using arrays. 

Considering other functions in `toqito` also use `scipy.sparse`, I might have to create additional issues for those that require similar changes. For example, see below:

https://github.com/vprusso/toqito/blob/2714b92648a84b5b8b09bb35aab7f31da9fbc975/toqito/perms/symmetric_projection.py#L12

Bchass: Merging next week is totally fine. My initial thought after this refactor is what you already hit on - need to remove all sparse matrix classes instead of converting a matrix to an array. I think it would be a better approach to refactoring all to use an array from the start.

vprusso: Sounds like a plan to me, @Bchass . Do let us know if we can provide any additional input or help, and thank you again!

purva-thakre (approved):
LGTM! Feel free to merge!

PR #522, opened 2024-03-28
Pretty good pretty bad measurements

### LINKED ISSUE #147: Feature: Pretty Good Measurement
Given a collection of vectors or matrices representing quantum states, compute the corresponding "pretty good measurement". 

A formal definition of the "pretty good measurement" can be found in Section 3.1 (page 137; "The pretty good measurement") of John Watrous's [textbook](https://cs.uwaterloo.ca/~watrous/TQI/TQI.pdf).

In order to define the pretty good measurement, one must first construct the "average state" (simply sum up the density operators in your collection) as specified in Equation 3.49. Then, compute the pretty good measurement by way of either Equation 3.50 (if the average state is positive definite) or by way of Equation 3.52 (if the average state is not positive definite).

Note that computing negative fractional powers of matrices will benefit from the `scipy.linalg.fractional_matrix_power` function (documentation [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.fractional_matrix_power.html))

The functionality for this should be created in `measurements/pretty_good_measurement.py` with corresponding unit tests found in `tests/test_measurements/test_pretty_good_measurement.py`. Be sure to also update the docs in `/docs/measurements.rst` under "Measurements" with `toqito.measurements.pretty_good_measurement.py`.

vprusso:
Closes #147 

## Description
Adding the ability to generate a set of either "pretty good" or "pretty bad" measurement operators.

### DIFF
--- a/toqito/measurements/pretty_bad_measurement.py
+++ b/toqito/measurements/pretty_bad_measurement.py
@@ -7,7 +7,7 @@ from toqito.measurements import pretty_good_measurement
 def pretty_bad_measurement(states: list[np.ndarray], probs: list[float] | None = None) -> list[np.ndarray]:
     r"""Return the set of pretty bad measurements from a set of vectors and corresponding probabilities.
 
-    This computes the "pretty bad measurement" is defined in :cite:`Hughston_1993_Complete` and is an analogous idea to
+    This computes the "pretty bad measurement" as defined in :cite:`Hughston_1993_Complete` and is an analogous idea to
     the "pretty good measurement" from :cite:`McIrvin_2024_Pretty`. The "pretty bad measurement" is useful in the
     context of state exclusion where the pretty good measurement is often used for minimum-error quantum state
     discrimination.
--- a/toqito/measurements/pretty_bad_measurement.py
+++ b/toqito/measurements/pretty_bad_measurement.py
@@ -12,6 +12,18 @@ def pretty_bad_measurement(states: list[np.ndarray], probs: list[float] | None =
     context of state exclusion where the pretty good measurmennt is often used for minimum-error quantum state
     discrimination.
 
+    The pretty bad measurement (PBM) is defined in terms of the pretty good measurement (PGM). Recall that the PGM is
+    defined as a set of POVMs :math:`(G_1, \ldots, G_n)` such that
+
+    .. math::
+        G_i = P^{-1/2} \left(p_i \rho_i\right) P^{-1/2} \quad \text{where} \quad
+        P = \sum_{i=1}^n p_i \rho_i.
+
+    By proxy, the corresponding PBM is defined as a set of POVMs :math:`(B_1, \ldots, B_n)` where
+
+    .. math::
+        B_i = \frac{1}{n - 1} \left(\mathbb{I} - G_i\right).
+
     See Also
     ========
     pretty_good_measurement
--- a/toqito/measurements/pretty_bad_measurement.py
+++ b/toqito/measurements/pretty_bad_measurement.py
@@ -1,0 +1,43 @@
+"""Compute the set of pretty bad measurements from an ensemble."""
+import numpy as np
+
+
+def pretty_bad_measurement(vectors: list[np.ndarray], probs: list[float]) -> list[np.ndarray]:
+    r"""Return the set of pretty bad measurements from a set of vectors and corresponding probabilities.
+
+    This computes the "pretty bad measurement" is defined in :cite:`Hughston_1993_Complete,` and is an analogous idea to
+    the "pretty good measurement" from :cite:`McIrvin_2024_Pretty,`. The "pretty bad measurement" is useful in the
+    context of state exclusion where the pretty good measurmennt is often used for minimum-error quantum state
+    discrimination.
+
+    See Also
+    ========
+    pretty_good_measurement
+
+    Examples
+    ========
+    Consider the depolarizing and identity channels in a 2-dimensional space. The depolarizing channel parameter is
+    set to 0.2:
+
+    >>> import numpy as np
+    >>> from toqito.channels import depolarizing
+    >>> from toqito.channel_metrics import diamond_norm
+    >>> choi_depolarizing = depolarizing(dim=2, param_p=0.2)
+    >>> choi_identity = np.identity(2**2)
+    >>> dn = diamond_norm(choi_depolarizing, choi_identity)
+    >>> print("Diamond norm between depolarizing and identity channels: ", '%.2f' % dn)
+    Diamond norm between depolarizing and identity channels:  -0.00
+
+    References
+    ==========
+        .. bibliography::
+            :filter: docname in docnames
+
+
+    :raises ValueError: If matrices are not of equal dimension.
+    :raises ValueError: If matrices are not square.
+    :param choi_1: A 4**N by 4**N matrix (where N is the number of qubits).
+    :param choi_2: A 4**N by 4**N matrix (where N is the number of qubits).
+
+    """
+    pass
--- a/toqito/measurements/pretty_bad_measurement.py
+++ b/toqito/measurements/pretty_bad_measurement.py
@@ -34,9 +34,9 @@ def pretty_bad_measurement(states: list[np.ndarray], probs: Optional[list[float]
     >>> pgm = pretty_bad_measurement(states, probs)
     >>> pgm
     [array([[0.16666667, 0.        ],
-        [0.        , 0.5       ]]), array([[ 0.41666667, -0.14433757],
-        [-0.14433757,  0.25      ]]), array([[0.41666667, 0.14433757],
-        [0.14433757, 0.25      ]])]
+           [0.        , 0.5       ]]), array([[ 0.41666667, -0.14433757],
+           [-0.14433757,  0.25      ]]), array([[0.41666667, 0.14433757],
+           [0.14433757, 0.25      ]])]
 
     References
     ==========
--- a/toqito/measurements/pretty_bad_measurement.py
+++ b/toqito/measurements/pretty_bad_measurement.py
@@ -12,8 +12,8 @@ def pretty_bad_measurement(states: list[np.ndarray], probs: list[float] | None =
     context of state exclusion where the pretty good measurement is often used for minimum-error quantum state
     discrimination.
 
-    The pretty bad measurement (PBM) is defined in terms of an offset of the pretty good measurement (PGM). Recall that the PGM is
-    defined as a set of POVMs :math:`(G_1, \ldots, G_n)` such that
+    The pretty bad measurement (PBM) is defined in terms of an offset of the pretty good measurement (PGM). Recall that
+    the PGM is defined as a set of POVMs :math:`(G_1, \ldots, G_n)` such that
 
     .. math::
         G_i = P^{-1/2} \left(p_i \rho_i\right) P^{-1/2} \quad \text{where} \quad
--- a/toqito/measurements/pretty_bad_measurement.py
+++ b/toqito/measurements/pretty_bad_measurement.py
@@ -2,6 +2,7 @@
 from typing import Optional
 
 import numpy as np
+
 from toqito.measurements import pretty_good_measurement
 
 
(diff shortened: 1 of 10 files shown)


purva-thakre:
Is it Ok if I review this by next Thursday? I want to wrap up the WIP RFC we have for Mitiq.

vprusso:
> Is it Ok if I review this by next Thursday? I want to wrap up the WIP RFC we have for Mitiq over the next few days.

Of course, no rush, and thank you again!

toqito/measurements/pretty_good_measurement.py
```diff
@@ -0,0 +1,67 @@
...
+    References
+    ==========
+        .. bibliography::
+            :filter: docname in docnames
+
+
+    :raises ValueError: If number of vectors does not match number of probabilities.
+    :raises ValueError: If probabilities do not sum to 1.
+    :param states: A collection of either states provided as either vectors or density matrices.
+    :param probs: A set of probabilities.
```
purva-thakre, line 47:
```suggestion
    :param probs: A set of fixed probabilities for a given ensemble of quantum states. The function assumes a uniform probability distribution if the fixed probabilities for the input states are not provided.
```

toqito/measurements/pretty_good_measurement.py
```diff
@@ -0,0 +1,67 @@
...
+
+    """
+    n = len(states)
+
+    # If not probabilities are explicitly given, assume a uniform distribution.
+    if probs is None:
+        probs = n * [1 / n]
```
purva-thakre, line 54: Are you assuming the input ensemble is always normalized? If not, there could be an additional check before lines 53-54 because assuming a uniform distribution by default might not work in the non-normalized scenario when the input probs are not provided.

toqito/measurements/pretty_good_measurement.py
```diff
@@ -0,0 +1,67 @@
...
+
+    if len(states) != len(probs):
+        raise ValueError(f"Number of states {len(states)} must be equal to number of probabilities {len(probs)}")
+
+    if not np.isclose(sum(probs), 1):
+        raise ValueError("Probability vector should sum to 1.")
+
+    states = [vector_to_density_matrix(state) for state in states]
```
purva-thakre, line 62:
```suggestion
    states = [vector_to_density_matrix(state) for state in states]
    # note that `vector_to_density_matrix` will return a density matrix if the provided input is a `rho-ensemble`. 
    # the function will alter the input only when the input ensemble is a set of vectors. 
```
This line confused me a bit for the case when the ensemble is a bunch of density matrices. So, I added these comments. Feel free to reject the addition.

toqito/measurements/pretty_good_measurement.py
```diff
@@ -0,0 +1,67 @@
...
+import numpy as np
+import scipy
+
+from toqito.matrix_ops import vector_to_density_matrix
+
+
+def pretty_good_measurement(states: list[np.ndarray], probs: list[float] | None = None) -> list[np.ndarray]:
+    r"""Return the set of pretty good measurements from a set of vectors and corresponding probabilities.
+
+    This computes the "pretty good measurement" as initially defined in :cite:`Hughston_1993_Complete`.
```
purva-thakre, line 11: It might be better to define what "pretty good measurement" is in the docstring. Looking through the docs, we seem to consistently define the math behind the function. 

For example, https://toqito.readthedocs.io/en/latest/autoapi/channel_metrics/channel_fidelity/index.html#channel_metrics.channel_fidelity.channel_fidelity

purva-thakre (commented):
Added some quick review comments as I make my way through the 2 papers used for this PR. Will add more comments/suggestions later!

BTW great job with the turnaround for the `pretty_bad_measurement` function! I saw that it appeared on the arXiv just last week. :)

toqito/measurements/pretty_good_measurement.py
vprusso: Hmm. I'm assuming that the probability vector is normalized (if provided). I check to ensure that the vector is a valid probability vector (i.e., it sums to 1). If such a vector is not provided, I decide to use a uniform distribution as a default (which is the section that you've highlighted here). 

Maybe I'm missing the broader point here though of what you're trying to convey.

toqito/measurements/pretty_good_measurement.py
vprusso: I might leave this out only because the docs for the function describe how it operates, so I feel like duplicating the description of the application of the function may be redundant.

toqito/measurements/pretty_good_measurement.py
vprusso: Yes, absolutely. This has been updated (and has also been done for the PBM as well). Good call!

toqito/measurements/pretty_bad_measurement.py
```diff
@@ -0,0 +1,68 @@
...
+    :raises ValueError: If probabilities do not sum to 1.
+    :param states: A collection of either states provided as either vectors or density matrices.
+    :param probs: A set of probabilities.
+
+    """
+    n = len(states)
+
+    # If not probabilities are explicitly given, assume a uniform distribution.
+    if probs is None:
+        probs = n * [1 / n]
```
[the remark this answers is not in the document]
vprusso: My comment to that comment also applies here :)

vprusso:
As an added benefit, I think I nipped that pesky docs bug that was causing pseudo-failures in this diff as well just FYI! (@purva-thakre )

toqito/measurements/pretty_good_measurement.py
purva-thakre: I am talking about a case when the provided input vector ensemble is not normalized such that it might not agree with a uniform probability distribution. 

For example, the function is provided a non-normalized input ensemble of 4 vectors but we skip providing an input probability distribution that would also normalize the input vector ensemble: `1/3, 1/6, 1/6, 1/3` is different from `1/4, 1/4, 1,4, 1,4`. 

I think it might be better to use the square root of the inner product of each vector in the non-normalized ensemble with itself to assign a probability distribution after the highlighted lines. Hopefully, this makes sense?

purva-thakre: Perhaps, a better example is the trine state example from both docstrings. 

image
 What if the input is provided as `u_1, u_2 (without -1/2), u_3 (without -1/2)`. It won't make sense to use a uniform probability distribution `1/3, 1/3, 1/3` in this case.

toqito/measurements/pretty_bad_measurement.py
```diff
@@ -0,0 +1,80 @@
...
+    See Also
+    ========
+    pretty_good_measurement
+
+    Examples
+    ========
+    Consider the collection of trine states.
+
+    .. math::
+        u_1 = |0\rangle, \quad
```
purva-thakre, line 36:
```suggestion
        u_0 = |0\rangle, \quad
```

toqito/measurements/pretty_good_measurement.py
```diff
@@ -0,0 +1,76 @@
...
+    See Also
+    ========
+    pretty_bad_measurement
+
+    Examples
+    ========
+    Consider the collection of trine states.
+
+    .. math::
+        u_1 = |0\rangle, \quad
```
purva-thakre, line 29:
```suggestion
        u_0 = |0\rangle, \quad
```

toqito/measurements/pretty_good_measurement.py
```diff
@@ -0,0 +1,76 @@
...
+
+from toqito.matrix_ops import vector_to_density_matrix
+
+
+def pretty_good_measurement(states: list[np.ndarray], probs: list[float] | None = None) -> list[np.ndarray]:
+    r"""Return the set of pretty good measurements from a set of vectors and corresponding probabilities.
+
+    This computes the "pretty good measurement" as initially defined in :cite:`Hughston_1993_Complete`.
+
+    The pretty good measurement (PGM) (also known as the "square root measurement) is the set of POVMs :math:`(G_1,
```
purva-thakre, line 13:
```suggestion
    The pretty good measurement (PGM) (also known as the "square root measurement") is the set of POVMs :math:`(G_1,
```

toqito/measurements/tests/test_pretty_bad_measurement.py
```diff
@@ -0,0 +1,75 @@
+"""Test pretty_bad_measurement."""
+
+import numpy as np
+import pytest
+
+from toqito.measurements import pretty_bad_measurement
+from toqito.states import bell, trine
+
```
purva-thakre, line 8: I think you could add a test in the test files for both PGM and PBM to verify the output of the functions satisfies all POVM properties:

- all sum up to the identity or close to it (span the entire space)
- are orthogonal to each other, positive semi-definite

toqito/measurements/pretty_bad_measurement.py
```diff
@@ -0,0 +1,80 @@
+"""Compute the set of pretty bad measurements from an ensemble."""
+import numpy as np
+
+from toqito.measurements import pretty_good_measurement
+
+
+def pretty_bad_measurement(states: list[np.ndarray], probs: list[float] | None = None) -> list[np.ndarray]:
+    r"""Return the set of pretty bad measurements from a set of vectors and corresponding probabilities.
+
+    This computes the "pretty bad measurement" is defined in :cite:`Hughston_1993_Complete` and is an analogous idea to
```
purva-thakre, line 10:
```suggestion
    This computes the "pretty bad measurement" as defined in :cite:`Hughston_1993_Complete` and is an analogous idea to
```

toqito/measurements/pretty_bad_measurement.py
```diff
@@ -0,0 +1,80 @@
...
+    the "pretty good measurement" from :cite:`McIrvin_2024_Pretty`. The "pretty bad measurement" is useful in the
+    context of state exclusion where the pretty good measurmennt is often used for minimum-error quantum state
```
purva-thakre, line 12:
```suggestion
    context of state exclusion where the pretty good measurement is often used for minimum-error quantum state
```

toqito/measurements/pretty_bad_measurement.py
```diff
@@ -0,0 +1,80 @@
...
+    discrimination.
+
+    The pretty bad measurement (PBM) is defined in terms of the pretty good measurement (PGM). Recall that the PGM is
```
purva-thakre, line 15:
```suggestion
    The pretty bad measurement (PBM) is defined in terms of an offset of the pretty good measurement (PGM). Recall that the PGM is
```

toqito/measurements/pretty_good_measurement.py
vprusso: Hmm, but I think there might be a distinction here, right? The probabilities in this case are intended to be treated as "the probability with which the state is selected from the set of states" (so, the ensemble is a tuple--(p_i, rho_i) where the state is a density operator and the p_i is a probability that corresponds to the probability that rho_i is selected). So, IOW, the 1/2 and -1/2 here in this case aren't necessarily relevant to the probability vector object (they are only relevant insofar as they are used to define the actual state).

Does that make sense?

toqito/measurements/tests/test_pretty_bad_measurement.py
vprusso: Hmm, I'm not sure about adding these extra tests. In some way, this would be testing the functionality of the convex optimization solvers. Since the constraints themselves enforce this property, it may be redundant to include tests on top of that to check if the measurements (the variables that are returned from the SDP) are valid POVMs. But then again, I can see an argument on the other side as well as we check the optimal value. I'm probably lean towards not adding this, but let me know what you think.

vprusso:
Modulo those comment responses, I believe this is in a good state to go. Good call on adding the issue for docs--I think that's a neat idea! Let me know if you disagree, but if you do agree, feel free to let me know or merge away! 🚀

toqito/measurements/pretty_good_measurement.py
purva-thakre: Actually, no.

Issue #529, opened 2024-04-05
Remove `csr_matrix` in toqito/matrices/pauli.py


vprusso:

Bchass:
🙋‍♂️

PR #539, opened 2024-04-05
pauli refactor

Bchass:
## Description
Closes: https://github.com/vprusso/toqito/issues/529

I left `is_sparse` in for the moment even though it's not technically being used in `pauli.py` due to the test cases relying on True or False. Would it make sense to completely remove `is_sparse` and drop `True` from the test cases? @purva-thakre  I believe so, but drafting this for PR for now.

## Changes
  -  [x] Remove `csr_matrix`

## Checklist

### DIFF
--- a/toqito/matrices/pauli.py
+++ b/toqito/matrices/pauli.py
@@ -2,10 +2,13 @@
 
 import numpy as np
 
+#from scipy import sparse
+from scipy.sparse import csr_array
+
 from toqito.matrix_ops import tensor
 
 
-def pauli(ind: int | str | list[int] | list[str], is_sparse: bool = False) -> np.ndarray:
+def pauli(ind: int | str | list[int] | list[str], is_sparse: bool = False) -> csr_array:
     r"""Produce a Pauli operator :cite:`WikiPauli`.
 
     Provides the 2-by-2 Pauli matrix indicated by the value of :code:`ind`. The
@@ -77,8 +80,8 @@ def pauli(ind: int | str | list[int] | list[str], is_sparse: bool = False) -> np
         :filter: docname in docnames
 
     :param ind: The index to indicate which Pauli operator to generate.
-    :param is_sparse: Returns a sparse matrix if set to True and a non-sparse
-                      matrix if set to False.
+    :param is_sparse: Returns a compressed sparse row array if set to True 
+                      and a non compressed sparse row array if set to False.
 
     """
     if isinstance(ind, (int, str)):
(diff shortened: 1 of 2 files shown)


toqito/matrices/pauli.py
```diff
@@ -92,9 +91,6 @@ def pauli(ind: int | str | list[int] | list[str], is_sparse: bool = False) -> np
         else:
             pauli_mat = np.identity(2)
 
-        if is_sparse:
-            pauli_mat = sparse.csr_matrix(pauli_mat)  # pylint: disable=redefined-variable-type
```
purva-thakre, line 96: @Bchass I think instead of removing `is_sparse` option  completely due to `sparse.csr_matrix`, we could try to switch to using `sparse.csr_array` instead. 

https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_array.html#scipy.sparse.csr_array

I haven't tried this out yet, FYI! If you run into a lot of issues, then we could simply remove that option in case we can't find other options.

Bchass: Looked into `sparse.csr_array` and it works a lot better for this use case. No need to alter any test cases.

toqito/matrices/pauli.py
```diff
@@ -1,12 +1,12 @@
...
 
 import numpy as np
-from scipy import sparse
+from scipy.sparse import csr_array
 
 from toqito.matrix_ops import tensor
 
 
-def pauli(ind: int | str | list[int] | list[str], is_sparse: bool = False) -> np.ndarray | sparse.csr_matrix:
+def pauli(ind: int | str | list[int] | list[str], is_sparse: bool = False) -> csr_array:
```
purva-thakre, line 9: Why remove `np.ndarray`? You get a `csr_array` for the output only when we provide `is_sparse` is not `False`.

Bchass: I was thinking back to: https://github.com/vprusso/toqito/pull/521#discussion_r1543072900 and that's why initially I removed `is_sparse` and made some changes to the test cases. But I'll wait for a further response.

purva-thakre: Yes, but `issparse` is a method for both sparse array and sparse matrix types. So, the structure of the function shouldn't change that much. Before, the return type was an array or a sparse matrix. Now, it can be an `np.ndarray` for all options of allowed inputs. 

https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.issparse.html#scipy.sparse.issparse

purva-thakre (approved):
LGTM! I'll wait for @vprusso to Ok this before merging.

PR #545, opened 2024-04-09
matrices: simplify gen_pauli_x construction

atomgardner:
Eliminate the reassignments in the gen_pauli_x construction by rolling along an axis.

### DIFF
--- a/toqito/matrices/gen_pauli_x.py
+++ b/toqito/matrices/gen_pauli_x.py
@@ -51,9 +51,4 @@ def gen_pauli_x(dim: int) -> np.ndarray:
     :return: :code:`dim`-by-:code:`dim` gen_pauli_x matrix.
 
     """
-    gen_pauli_x_mat = np.identity(dim)
-    gen_pauli_x_mat = np.roll(gen_pauli_x_mat, -1)
-    gen_pauli_x_mat[:, -1] = np.array([0] * dim)
-    gen_pauli_x_mat[0, -1] = 1
-
-    return gen_pauli_x_mat
+    return np.roll(np.identity(dim), -1, axis=1)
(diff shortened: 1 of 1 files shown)


toqito/matrices/gen_pauli_x.py
```diff
@@ -51,9 +51,4 @@ def gen_pauli_x(dim: int) -> np.ndarray:
     :return: :code:`dim`-by-:code:`dim` gen_pauli_x matrix.
 
     """
-    gen_pauli_x_mat = np.identity(dim)
-    gen_pauli_x_mat = np.roll(gen_pauli_x_mat, -1)
-    gen_pauli_x_mat[:, -1] = np.array([0] * dim)
-    gen_pauli_x_mat[0, -1] = 1
-
-    return gen_pauli_x_mat
+    return np.roll(np.identity(dim), -1, axis=1)
```
purva-thakre, line 54:
```suggestion
    return np.roll(np.identity(dim), -1, axis=1)
    # first column of the identity matrix becomes the last column due to `shift = -1` and `axis=1`
    # https://numpy.org/doc/stable/reference/generated/numpy.roll.html
```

purva-thakre (approved):
LGTM ! Added some minor comments because the `-1, axis=1` confused be a little bit. 

@vprusso Feel free to reject the changes as once can also go to numpy's documentation to look through the examples.

toqito/matrices/gen_pauli_x.py
vprusso: I think I'd opt to not have comments here as the call to `roll` based on the existing `numpy` docs is clear enough (at least in my mind). Or, if we are to include a comment, I would just include 

```py
# First column of the identity matrix becomes the last column due to `shift = -1` and `axis=1`
```

but place it above the line

```py
    return np.roll(np.identity(dim), -1, axis=1)
```

Either moving the comment or not including it work for me!

Issue #244, opened 2023-11-15
Refactor: The channel_dim function should live in channel_props


vprusso:
Right now, the `channel_dim.py` file lives in `helper/` but should most likely live in `channel_props/`.

purva-thakre:
Attempt 3 to resolve this issue also resulted in similar test failures. 

Even after we changed the directory structure, moving the function to `channel_props` caused some import issues where we have to specify the full import path `from toqito.channel_props.channel_dim import channel_dim`.

purva-thakre:
Check all import paths:

https://stackoverflow.com/a/6466248
https://docs.python.org/3/reference/import.html

purva-thakre:
Replaced by #677

PR #575, opened 2024-04-30
Random doctest failures

### LINKED ISSUE #473: Failing examples in the doctest workflow
3 tests fail at random when `make doctest` is run in `toqito/docs` with `TypeError: 'module' object is not callable`

1. `state_props/has_symmetric_extension`
2. `state_props/is_distinguishable`
3. `state_props/is_antidistinguishable`

The failing tests pass after the same command is run again. See https://github.com/vprusso/toqito/pull/459#issuecomment-1937792083 for more info. 

Probably an issue with one of the imported functions where the solver is running into issues?

purva-thakre:
## Description
Fixes #473 

This is technically **not a fix** as we have to specify the full import path for a bunch of files only. If this is indeed the source of the issue where Python could not distinguish between the module and functions, we should rename all `.py` files to not have the same names as the functions in them. 

For example, rename `toqito/state_props/is_antidistinguishable.py` to `toqito/state_props/is_antidistinguishable_function.py`

## Changes
Notable changes that this PR has either accomplished or will accomplish. Feel free to add more lines to the itemized list
below.

  -  [x] Added full import path for the failing doctest examples

## Checklist

### DIFF
--- a/toqito/state_props/has_symmetric_extension.py
+++ b/toqito/state_props/has_symmetric_extension.py
@@ -4,7 +4,7 @@ import numpy as np
 from picos import partial_trace
 
 from toqito.matrix_props import is_positive_semidefinite
-from toqito.state_opt import symmetric_extension_hierarchy
+from toqito.state_opt.symmetric_extension_hierarchy import symmetric_extension_hierarchy
 from toqito.state_props import is_ppt
 
 
--- a/toqito/state_props/has_symmetric_extension.py
+++ b/toqito/state_props/has_symmetric_extension.py
@@ -5,6 +5,9 @@ from picos import partial_trace
 
 from toqito.matrix_props import is_positive_semidefinite
 from toqito.state_opt.symmetric_extension_hierarchy import symmetric_extension_hierarchy
+
+# The full import path was specified here because the doctest workflow was failing when the above function could not be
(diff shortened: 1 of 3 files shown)


vprusso:
Definitely a useful one to make! Thank you, @purva-thakre !

purva-thakre:
@vprusso Ha! No worries, the random failures were driving me crazy. BTW you commented before I got the chance to edit the PR description. I think the issue is due to Python getting confused about importing the module or the function. An easier solution would be to rename all the `.py` files.

vprusso:
Hmm, renaming the files seems dangerous and could break backwards compatibility. Not sure exactly what's going on with the failures, but I would consider something like that as a last ditch effort I'd say.

purva-thakre:
So, do you want me to create an issue so that we work on this sometime in the future? I did add a comment in case we accidentally remove these. 

I have a feeling this PR and #244 are closely related.

vprusso:
For now if this works, that's great. Still curious why this is failing though as it seems to abide by the same pattern as other functions abide by. Curious, but for now, LGTM

vprusso (approved):

purva-thakre:
>  Still curious why this is failing though as it seems to abide by the same pattern as other functions abide by. 

Same here!

PR #602, opened 2024-05-30
Upated Indexing from 1 -> 0

### LINKED ISSUE #168: Refactor: Permute systems index offset
In much the same spirit as https://github.com/vprusso/toqito/issues/145, it would make sense to have a consistent indexing pattern amongst the functions offered in `toqito`. We should also make this change in the `permute_systems.py` file as well. 

That is, the indexing offset for `permute_systems` starts at `1` but to be consistent with the rest of the functions (and indeed, to be more intuitive from the perspective of a Python programmer) it should assume an index starting at `0`.

anushkrishnav:
## Description
Fixes #168 
Moving Permute system index offset from 0->1

## Changes
Update the code 

the docs and tests would be updated once the solution is approved 

## Checklist


purva-thakre:
> the docs and tests would be updated once the solution is approved

Can you go ahead and make changes to the unit tests and doc examples? It's difficult to see how the new changes would work otherwise.

anushkrishnav:
Yes I am on it right now

purva-thakre:
> fixed it by setting PYTHONPATH=. pytest

This depends on how your local virtual environment is setup.

anushkrishnav:
I have updated the tests, there is one case that's failing which is the 16x16 check part from that all the other cases are working well not sure whats going wrong, is there a way to verify what the right result would be ?

purva-thakre:
> there is one case that's failing which is the 16x16 check part 

Which one are you talking about? Can you link it?

anushkrishnav:
> > there is one case that's failing which is the 16x16 check part
> 
> Which one are you talking about? Can you link it?

Solved the problem , all cases pass now, made a small mistake fixed it

toqito/states/brauer.py
```diff
@@ -69,6 +69,7 @@ def brauer(dim: int, p_val: int) -> np.ndarray:
     matchings = perfect_matchings(2 * p_val)
     num_matchings = matchings.shape[0]
     state = np.zeros((dim ** (2 * p_val), num_matchings))
+    matchings = matchings - 1
```
purva-thakre, line 72: Can you please add a comment here as to why you are using `matchings - 1`? 

It will be helpful if we have to make changes to this function in the future.

purva-thakre (changes requested):
Getting close to merging this! 

I suggested some minor changes and also requested a clarification on a change you made.

anushkrishnav:
Will work on the change requests shortly , I think the comment was removed by mistake @purva-thakre

toqito/perms/permutation_operator.py
```diff
@@ -54,10 +53,10 @@ def permutation_operator(
     """
     # Allow the user to enter a single number for `dim`.
     if isinstance(dim, int):
-        dim = dim * np.ones(max(perm))
+        dim = [dim] * len(perm)
     if isinstance(dim, list):
         dim = np.array(dim)
 
     mat = sp.sparse.identity(int(np.prod(dim))) if is_sparse else np.identity(int(np.prod(dim)))
-    # Swap the rows of the identity matrix appropriately.
```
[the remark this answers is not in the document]
anushkrishnav: removed by accident will place the comment shortly

toqito/perms/tests/test_permute_systems.py
```diff
@@ -10,19 +10,19 @@ def test_permute_systems_vec():
...
-    res = permute_systems(test_input_mat, [2, 1])
+    res = permute_systems(test_input_mat, [1, 0])
 
     bool_mat = np.isclose(res, expected_res)
     np.testing.assert_equal(np.all(bool_mat), True)
 
 
 def test_permute_systems_m2_m2():
-    """Permute system for perm = [2,1] and dim = [2, 2]."""
+    """Permute system for perm = [0,1] and matrix of dimension 2-by-2."""
```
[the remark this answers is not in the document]
anushkrishnav: Sounds good

toqito/perms/tests/test_permute_systems.py
```diff
@@ -152,7 +151,7 @@ def test_permute_systems_invalid_perm_vector():
     """Invalid input for permute systems."""
     with np.testing.assert_raises(ValueError):
         test_input_mat = np.array([[1, 2, 3, 4], [5, 6, 7, 8], [9, 10, 11, 12], [13, 14, 15, 16]])
-        permute_systems(test_input_mat, [0, 1])
+        permute_systems(test_input_mat, [1, 2])
```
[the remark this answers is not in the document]
anushkrishnav: Gotcha will update the case and see how it works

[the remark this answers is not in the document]
anushkrishnav: The test passes fine, pushing the changes shortly

toqito/states/brauer.py
anushkrishnav: The perfect matching system Returns  a list of matching which start at an index 1 hence I had to do the -1 but thinking about that, I think its best I fix this at source and make the perfect matching return an array that starts from 0 than one . Does it make sense? 
```
    This is an example of how to generate all perfect matchings of the numbers 1, 2, 3, 4.

    >>> from toqito.perms import perfect_matchings
    >>> perfect_matchings(4)
    array([[1, 2, 3, 4],
           [1, 3, 2, 4],
           [1, 4, 3, 2]])
     ```

File: toqito/perms/permutation_operator.py [hunk 5]
```diff
@@ -54,10 +53,10 @@ def permutation_operator(
     """
     # Allow the user to enter a single number for `dim`.
     if isinstance(dim, int):
-        dim = dim * np.ones(max(perm))
+        dim = [dim] * len(perm)
```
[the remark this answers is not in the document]
anushkrishnav: So the line generates a numpy array filled with the max value of perm 
in this example `res = permutation_operator(2, [1, 3, 2])`, max of perm gives us an dimension array of 3s,  which makes sense.  But now that we have moved to a 0 index system,  for `res = permutation_operator(2, [0, 2, 1])` we would get an array of 2s for dimension when we need an array of 3s.

The change from max to len was made for that reason.
Also, I didn't use np.ones(len(perm)) because there is the if case in the next line that checks if dim is a list and converts it into a numpy array, 
`dim = [dim] * len(perm)` creates a dimension list `[2]*3` which then laters gets updated into an array, I can convert the list to an array in the same line if you wish 
`dim = [dim] * np.ones(len(perm))`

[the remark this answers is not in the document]
purva-thakre: I don't think the logic of the code block is working as you expect it to .

I used print statements to see how the `dim` argument is being used here. With your changes, 1 `if` condition creates a list object and the other one created a numpy array. Plus, once you use the first `if`, it shouldn't go into the second `if` considering you have created the `dim` you require but that's what this code block is currently doing. 

```python
    if isinstance(dim, int):
        dim = [dim] * len(perm)
        print(type(dim))
        print("instance is int dim", dim, '\n')
    if isinstance(dim, list):
        dim = np.array(dim)
        print(type(dim))
        print("instance is list dim",dim, '\n')
```

For `permutation_operator(2, [2, 1])`, the output due to the changes is
```
<class 'list'>
instance is int dim [2, 2] 

<class 'numpy.ndarray'>
instance is list dim [2 2] 

[[1. 0. 0. 0.]
 [0. 1. 0. 0.]
 [0. 0. 1. 0.]
 [0. 0. 0. 1.]] 

array([[1., 0., 0., 0.],
       [0., 0., 1., 0.],
       [0., 1., 0., 0.],
       [0., 0., 0., 1.]])
```

whereas you expect it to be 

```
<class 'numpy.ndarray'>
instance is int dim [2. 2.] 

[[1. 0. 0. 0.]
 [0. 1. 0. 0.]
 [0. 0. 1. 0.]
 [0. 0. 0. 1.]] 

array([[1., 0., 0., 0.],
       [0., 0., 1., 0.],
       [0., 1., 0., 0.],
       [0., 0., 0., 1.]])
```

toqito/perms/tests/test_perfect_matchings.py
```diff
@@ -32,24 +33,25 @@ def test_perfect_matchings_odd():
 def test_perfect_matchings_num_6():
     """All perfect matchings of size 6."""
     res = perfect_matchings(6)
+    print(res)
```
purva-thakre, line 36: remove this line

toqito/perms/tests/test_perfect_matchings.py
```diff
@@ -8,15 +8,16 @@
 def test_perfect_matchings_num_4():
     """All perfect matchings of size 4."""
     res = perfect_matchings(4)
-    expected_res = np.array([[1, 2, 3, 4], [1, 3, 2, 4], [1, 4, 3, 2]])
+    print(res)
```
purva-thakre, line 11: remove this line

toqito/perms/permutation_operator.py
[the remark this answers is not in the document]
purva-thakre: I'll add a comment later with an example about `max(perm)` vs. `len(perm)`. 

From what I found, it still doesn't make sense to me. Need to work on simple demo example before resolving this thread.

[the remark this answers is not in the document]
anushkrishnav: > Plus, once you use the first `if`, it shouldn't go into the second `if` considering you have created the `dim` you require but that's what this code block is currently doing.
> 
> ```python
>     if isinstance(dim, int):
>         dim = [dim] * len(perm)
>         print(type(dim))
>         print("instance is int dim", dim, '\n')
>     if isinstance(dim, list):
>         dim = np.array(dim)
>         print(type(dim))
>         print("instance is list dim",dim, '\n')
> ```
> 
> For `permutation_operator(2, [2, 1])`, the output due to the changes is
> 
> ```
> <class 'list'>
> instance is int dim [2, 2] 
> 
> <class 'numpy.ndarray'>
> instance is list dim [2 2] 
> 
> [[1. 0. 0. 0.]
>  [0. 1. 0. 0.]
>  [0. 0. 1. 0.]
>  [0. 0. 0. 1.]] 
> 
> array([[1., 0., 0., 0.],
>        [0., 0., 1., 0.],
>        [0., 1., 0., 0.],
>        [0., 0., 0., 1.]])
> ```
> 
> whereas you expect it to be
> 
> ```
> <class 'numpy.ndarray'>
> instance is int dim [2. 2.] 
> 
> [[1. 0. 0. 0.]
>  [0. 1. 0. 0.]
>  [0. 0. 1. 0.]
>  [0. 0. 0. 1.]] 
> 
> array([[1., 0., 0., 0.],
>        [0., 0., 1., 0.],
>        [0., 1., 0., 0.],
>        [0., 0., 0., 1.]])
> ```

I do see the issue , yes you are right the first if triggers if dim is an integer and the second if given as a list, which makes sense, will correct it, sorry not sure how I missed this, but regarding the max and len, in 1 index list the max perm is essentially the size of the perm but in 0 index its max+1 rather I changed it to be len which is the ideal one, but I am open to learn why this is incorrect or wont work here

vprusso:
Great, thanks for continuing to work on those changes, and feel free to let us know if you have any additional questions, @anushkrishnav !

anushkrishnav:
> Great, thanks for continuing to work on those changes, and feel free to let us know if you have any additional questions, @anushkrishnav !

No problem, I am just waiting for the review, It was fun working !

toqito/perms/permutation_operator.py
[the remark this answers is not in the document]
purva-thakre: > regarding the max and len, in 1 index list the max perm is essentially the size of the perm but in 0 index its max+1 rather I changed it to be len which is the ideal one

I am having some trouble trying to understand what you are saying here. Could you provide a simple example?

`max` finds the maximum value in an array or along an axis whereas `len` gives you the length of something. Where does the issue of shifted indices affect the output below? 

```python
perm = [7, 8, 9, 10]
print(max(perm))
print(len(perm))
```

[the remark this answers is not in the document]
anushkrishnav: > > regarding the max and len, in 1 index list the max perm is essentially the size of the perm but in 0 index its max+1 rather I changed it to be len which is the ideal one
> 
> I am having some trouble trying to understand what you are saying here. Could you provide a simple example?
> 
> `max` finds the maximum value along an axis (if we choose to use it) whereas `len` gives you the length of something. Where does the issue of shifted indices affect the output below?
> 
> ```python
> perm = [7, 8, 9, 10]
> print(max(perm))
> print(len(perm))
> ```
> 
> Why is `len` the ideal option when you shift the index by -1 position?

If my understanding is right, The perm parameter must be a valid permutation of the indices of the subsystems. This means it should be a rearrangement of the numbers from 0 to 𝑛 -1, where n is the number of subsystems.

If you have a system with 4 subsystems, valid permutations of the indices [0, 1, 2, 3] could be:

```
[0, 1, 2, 3] (identity permutation, no change)
[3, 2, 1, 0] (reverse order)
[1, 0, 3, 2] (swap first two and last two)
[2, 3, 0, 1] (rotate left)

```
so [7,8,9,10] will not be a valid perm list.
Next, the dim parameter should be based on the number of subsystems, doing `max(perm)` would get us 3 while this is a system with 4 subsystems, len() will give us the correct count of the system, earlier when we start perm list with a start index of 1 doing max() got us the right count of subsystems `max([1,2,3,4] ) = 4` but now that the same starts with 0 instead of 1 ` max([0,1,2,3])=3 ` which would be incorrect, So instead of adding +1 after max(),  I used len ()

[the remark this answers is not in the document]
purva-thakre: Let's just leave it as `max+1` then. I have a use-case scenario where I think the len option will give the incorrect result. However, I don't think I will be able to test it out over the next few days.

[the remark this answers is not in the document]
anushkrishnav: Sounds good will update the code to max +1

toqito/channels/partial_trace.py
```diff
@@ -189,10 +189,10 @@ def partial_trace(
         sys = np.array(sys)
     if isinstance(sys, int):
         sys = np.array([sys])
-    set_diff = list(set(list(range(1, num_sys + 1))) - set(sys + 1))
 
+    set_diff = list(set(list(range(0, num_sys))) - set(sys))
```
purva-thakre, line 193:
```suggestion
    set_diff = list(set(list(range(num_sys))) - set(sys))
```
Considering `list(range(4))` is equivalent to `list(range(0, 4))`, you don't need to specify the starting index every time if you want to start from 0.

toqito/channels/partial_transpose.py
```diff
@@ -167,8 +167,8 @@ def partial_transpose(
     sub_sys_vec_r = prod_dim_r * np.ones(int(sub_prod_r)) / sub_prod_r
     sub_sys_vec_c = prod_dim_c * np.ones(int(sub_prod_c)) / sub_prod_c
 
-    set_diff = list(set(list(range(1, num_sys + 1))) - set(sys + 1))
-    perm = (sys + 1).tolist()[:]
+    set_diff = list(set(list(range(0, num_sys ))) - set(sys ))
```
purva-thakre, line 170:
```suggestion
    set_diff = list(set(list(range(num_sys ))) - set(sys ))
```

toqito/nonlocal_games/quantum_hedging.py
```diff
@@ -90,8 +90,8 @@ def __init__(self, q_a: np.ndarray, num_reps: int) -> None:
         # action:
         #   π(y1 ⊗ y2 ⊗ x1 ⊗ x2) = y1 ⊗ x1 ⊗ y2 ⊗ x2
         # for all y1 ∈ Y1, y2 ∈ Y2, x1 ∈ X1, x2 ∈ X2.).
-        l_1 = list(range(1, self._num_reps + 1))
-        l_2 = list(range(self._num_reps + 1, self._num_reps**2 + 1))
+        l_1 = list(range(0, self._num_reps ))
```
purva-thakre, line 93:
```suggestion
        l_1 = list(range(self._num_reps ))
```

toqito/perms/antisymmetric_projection.py
```diff
@@ -84,7 +84,7 @@ def antisymmetric_projection(dim: int, p_param: int = 2, partial: bool = False)
     if dim < p_param:
         return np.zeros((dimp, dimp * (1 - partial)))
 
-    p_list = np.array(list(permutations(np.arange(1, p_param + 1))))
+    p_list = np.array(list(permutations(np.arange(0, p_param ))))
```
purva-thakre, line 87:
```suggestion
    p_list = np.array(list(permutations(np.arange(p_param ))))
```

toqito/perms/permute_systems.py
```diff
@@ -176,7 +177,7 @@ def permute_systems(
     prod_dim_r = int(np.prod(dim[0, :]))
     prod_dim_c = int(np.prod(dim[1, :]))
 
-    if sorted(perm) != list(range(1, num_sys + 1)):
+    if sorted(perm) != list(range(0, num_sys )):
```
purva-thakre, line 180:
```suggestion
    if sorted(perm) != list(range(num_sys )):
```

toqito/perms/swap.py
```diff
@@ -145,7 +145,7 @@ def swap(
         raise ValueError("InvalidSys: `sys` must be a vector with exactly two elements.")
 
     # Swap the indicated subsystems.
-    perm = np.array(range(1, num_sys + 1))
+    perm = np.array(range(0, num_sys ))
```
purva-thakre, line 148:
```suggestion
    perm = np.array(range(num_sys ))
```

toqito/perms/symmetric_projection.py
```diff
@@ -81,7 +81,7 @@ def symmetric_projection(dim: int, p_val: int = 2, partial: bool = False) -> [np
     if p_val == 1:
         return np.eye(dim)
 
-    p_list = np.array(list(permutations(np.arange(1, p_val + 1))))
+    p_list = np.array(list(permutations(np.arange(0, p_val ))))
```
purva-thakre, line 84:
```suggestion
    p_list = np.array(list(permutations(np.arange(p_val ))))
```

toqito/state_opt/optimal_clone.py
```diff
@@ -130,7 +130,7 @@ def optimal_clone(
         # sequence from: https://oeis.org/A023123
         q_a = tensor(q_a, num_reps)
         perm = []
-        for i in range(1, num_spaces + 1):
+        for i in range(0, num_spaces ):
```
purva-thakre, line 133:
```suggestion
        for i in range(num_spaces ):
```

toqito/channels/partial_trace.py
anushkrishnav: Agreed

purva-thakre:
Congrats on the bounty @anushkrishnav !

anushkrishnav:
> Congrats on the bounty @anushkrishnav !

Thank you !!!

Issue #601, opened 2024-05-30
Contribution Guide missing


anushkrishnav:
## Describe the bug
There is no contribution guide on the repo

vprusso:
Hi @anushkrishnav 

There is a contribution guide (you can find the file for this [here](https://github.com/vprusso/toqito/blob/8e2d225898f55c3ffdb15aacd433d923e76ba2c6/docs/contributing.rst#L7) and the rendered docs page for the guide [here](https://toqito.readthedocs.io/en/latest/getting_started.html#contributing)).

I'm going to close out this issue, but feel free to open it back up if you feel the guide is insufficient, and don't hesitate to propose any additional improvements. I appreciate the spirit of opening issues to improve the project if you find any friction anywhere else!

anushkrishnav:
Hey, Sorry for opening the issue I will send in a pr that adds the [doc](https://github.com/vprusso/toqito/blob/8e2d225898f55c3ffdb15aacd433d923e76ba2c6/docs/contributing.rst#L7) to the readme , I later discovered the rendered docs , which is pretty helpful

vprusso:
But the contributing guide is linked in the README:
<img width="878" alt="Screenshot 2024-05-30 at 7 49 47 AM" src="https://github.com/vprusso/toqito/assets/1562214/f776c77f-926e-4ffa-9336-7c4c21b798a2">

vprusso:
Ah, but it's broken. Well, that's not very helpful! :)

anushkrishnav:
ahh yehh ;)

PR #605, opened 2024-05-30
Update README.md

anushkrishnav:
Fixed a broken link to the contributing guide
#601

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ All contributions, bug reports, bug fixes, documentation improvements,
 enhancements, and ideas are welcome.
 
 A detailed overview of how to contribute can be found in the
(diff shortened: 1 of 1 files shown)


vprusso:
Hi @anushkrishnav  

Thanks for the fix. So I think we should opt to link to the rendered contributing guide (over the one that is the file on GitHub). 

So, for instance: this

```
[contributing guide](https://toqito.readthedocs.io/en/latest/getting_started.html#contributing).
```

anushkrishnav:
Sounds good , The rendered version would be a better alternative though I wouldn't mind reading the docs on the repo, the rendered is better

vprusso:
LGTM!

vprusso (approved):

PR #610, opened 2024-05-31
Regenerated and updated codecov token.

vprusso:
Some users mentioned that regenerating the token and updating it in the `.yml` file fixed this issue for them. Not sure if this will work, but let's hope!


purva-thakre (approved):
Fingers crossed!

PR #637, opened 2024-06-24
Removed lil sparse matrix for GHZ state.

vprusso:
Closes: #536 

- Removed `sparse.lil_matrix` from `ghz.py`
- Updated tests and docs (no longer need to use `.toarray()`.


purva-thakre (approved):
LGTM!

PR #651, opened 2024-07-07
Removing redundant note for doctest.

### LINKED ISSUE #650: Remove note on float for documentation
The note 

```
You do not need to use ‘%.2f’ % when you use this function.

....
```

It appears all over the place, and I find this a bit distracting (especially when noted numerous times in the same tutorial, etc.).

I'd opt to remove this note across the board.

vprusso:
Closes #650


purva-thakre:
> It appears all over the place, and I find this a bit distracting (especially when noted numerous times in the same tutorial, etc.)

I am fine with removing this if it is repeated across one page. 

We should mention it once for the other pages. The things required by `DOCTEST` don't make sense for someone who is not familiar with why we are something in the print output. What if they copy-paste an example into a local jupyter notebook? I had added it for this reason.

vprusso:
> > It appears all over the place, and I find this a bit distracting (especially when noted numerous times in the same tutorial, etc.)
> 
> I am fine with removing this if it is repeated across one page.
> 
> We should mention it once for the other pages. The things required by `DOCTEST` don't make sense for someone who is not familiar with why we are using something in the print output. What if they copy-paste an example into a local jupyter notebook? I had added it for this reason.

That makes sense--I think having it at least in one place definitely makes sense. Maybe in the contribution guide?

purva-thakre:
> I think having it at least in one place definitely makes sense. Maybe in the contribution guide?

What if we create a section called `Examples` in the `Getting Started` section? We could introduce all the tutorials, point a user to the examples in the API as well as have a note for the things we use for DOCTEST. 

Adding it to the contribution guide wouldn't be helpful for someone who is not interested in contributing to the project. Thoughts?

toqito/state_props/entanglement_of_formation.py
```diff
@@ -40,16 +40,8 @@ def entanglement_of_formation(rho: np.ndarray, dim: list[int] | int = None) -> f
...
-
-    .. note::
-        You do not need to use `'%.2f' %` when you use this function.
-
-        We use this to format our output such that `doctest` compares the calculated output to the
-        expected output upto two decimal points only. The accuracy of the solvers can calculate the
-        `float` output to a certain amount of precision such that the value deviates after a few digits
-        of accuracy.
+    >>> np.around(entanglement_of_formation(rho), decimals=2)
+    1.00
```
purva-thakre, line 44:
```suggestion
    >>> import numpy as np
    >>> np.around(entanglement_of_formation(rho), decimals=2)
    1.00
```
Sometimes `np.around` will round up to a higher number when we don't want it to. We don't see that here but when a function relies on optimization, `np.around(0.8)` gets rounded up to `1.0`. This is why I did not use it.

vprusso: Yes, but this would only occur if you don't specify the number of decimals. 

So, for instance:

```py
np.around(0.66)
1.0
```

However

```py
np.around(0.66, decimals=2)
0.66
```

Issue #620, opened 2024-06-05
Feature: Compute the spark of a matrix


vprusso:
Define a function and associated unit tests to compute the *[spark](https://en.wikipedia.org/wiki/Spark_(mathematics))* of a matrix.

Note that the spark is NP-hard to compute in general. Here is a candidate function for computing the spark of a matrix:

```py
def spark(matrix: np.ndarray) -> int:
    """Compute the spark of the given matrix."""

    # The spark of a matrix is equal to 1 if and only if the matrix has a zero column.
    if np.any(np.sum(matrix, axis=0) == 0):
        return 1

    # Otherwise, compute the NP-hard spark computation.
    _, n = matrix.shape
    for k in range(1, n + 1):
        for cols in combinations(range(n), k):
            submatrix = matrix[:, cols]
            if is_linearly_dependent(submatrix):
                return k
    # If no dependent set is found, return n + 1
    return n + 1
```

We would also need unit tests as well. Something like the following, for example:

```py
@pytest.mark.parametrize("matrix, expected_result", [
    # Spark is 3 since (no set of 1 or 2 columns that are linearly dependnet)
    # But there is a set of 3 columns that are linearly dependnet.
    (
        np.array([
            [1, 2, 0, 1],
            [1, 2, 0, 2],
            [1, 2, 0, 3],
            [1, 0, -3, 4],
        ]),
        3
    ),
    # Spark of matrix is 1 if there is a column of all zeros.
    (
        np.array([
            [1, 2, 0, 0],
            [1, 2, 0, 0],
            [1, 2, 0, 0],
            [1, 0, -3, 0],
        ]),
        1
    )
])
def test_spark(matrix, expected_result):
    np.testing.assert_allclose(spark(matrix), expected_result)
```

Presumably, this function would belong in `matrix_props/` and would require a more appropriate and thoroughly written docstring.

purva-thakre:
I did not know spark of a matrix existed!

The first page on Google search for this quantity also has a link to a character in the matrix movie. lol. 

![image](https://github.com/vprusso/toqito/assets/66048318/ae2ac209-78c5-4150-a2b7-8c821f7a97dd)

PR #780, opened 2024-09-13
Adding spark function and tests to matrix props.

vprusso:
Closes: #620 

- Adding `spark` function to compute the "[spark](https://en.wikipedia.org/wiki/Spark_(mathematics))" of a matrix.
- Adding reference and doc-string containing examples.
- Adding tests for spark function.


toqito/matrix_props/spark.py
```diff
@@ -0,0 +1,47 @@
...
+    ...               [0, 1, 1, 3],
+    ...               [1, 1, 2, 5]])
+    >>> spark(A)
+    3
+
+    Notes
+    =====
+    - If all columns are linearly independent, the function returns n_cols + 1.
+    - The time complexity of this implementation is O(2^n) in the worst case,
+      where n is the number of columns.
```
purva-thakre, line 27:
```suggestion
    - If all columns are linearly independent, the function returns $n_{\text{cols}} + 1$.
    - The time complexity of this implementation is $O(2^n)$ in the worst case,
      where $n$ is the number of columns.
```

purva-thakre: I think you should add a linearly independent example in addition to adding this note.

toqito/matrix_props/spark.py
```diff
@@ -0,0 +1,47 @@
...
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
+    :param mat: The input matrix.
+    :return: The spark of the input matrix :code:`mat`.
+
+    """
+    n_cols = mat.shape[1]
```
purva-thakre, line 38: minor nit: You should make it clear somewhere in the docstring that this method only works for a 2D numpy array.

toqito/matrix_props/spark.py
```diff
@@ -0,0 +1,47 @@
...
+
+    for k in range(1, n_cols + 1):
+        for cols in combinations(range(n_cols), k):
+            submatrix = mat[:, cols]
+            if np.linalg.matrix_rank(submatrix) < k:
+                return k
+
```
purva-thakre, line 45: Do you think we should add additional checks or unit tests for the case when the input matrix is not a square matrix? 

image

Screenshot taken from the linked Wikipedia page.

purva-thakre (changes requested):

toqito/matrix_props/spark.py
vprusso: Fixed!

toqito/matrix_props/spark.py
vprusso: Good point. Added!

toqito/matrix_props/spark.py
vprusso: Done!

vprusso:
@purva-thakre thanks for the comments! I believe I've addressed those, but let me know prior to merge!

purva-thakre (approved):

PR #795, opened 2024-09-20
Removing cvx_kron unused function.

vprusso:
- As `cvx_kron` is no longer used, am removing.


.gitignore
```diff
@@ -17,3 +17,4 @@ build/
 coverage.xml
 jupyter_execute/
 venv/*
+.aider*
```
purva-thakre, line 20: What is this new addition related to?

purva-thakre (approved):
LGTM! Thanks for noticing that we are not using this function anymore.

.gitignore
vprusso: Ah, that was some VSCode extension I was experimenting with. I removed that from the `.gitignore` file.

PR #797, opened 2024-09-20
Replacing '*' with '@' where appropriate.

vprusso:
- Replacing `*` operator with `@` operator for matrix or vector multiplications.


purva-thakre (approved):
LGTM! I think you missed 1 `*`. 

https://github.com/vprusso/toqito/blob/5dc4c71296731230816d29c0ef5a8a7fa6b1f6cd/toqito/perms/perfect_matchings.py#L67

vprusso:
> LGTM! I think you missed 1 `*`.
> 
> https://github.com/vprusso/toqito/blob/5dc4c71296731230816d29c0ef5a8a7fa6b1f6cd/toqito/perms/perfect_matchings.py#L67

As it happens, this needs to be `*` over `@`, but good catch!

vprusso:
> I don't think this fixed all the warnings. Just noticed a couple more:
> 
> https://github.com/vprusso/toqito/actions/runs/11151893975/job/30996320621?pr=821#step:5:3446
> 
> Should we create a new issue for these warnings?

Not sure if it's just me, but that link seems to freeze and not render any of the warnings that you might be seeing. If you see them though, definitely copy/paste them and create a separate issue as they should definitely be addressed. Thanks, @purva-thakre

purva-thakre:
You have to reload the page or scroll down to line 3446.

The output is big which is why I think ut's not rendering anything for you.

Issue #299, opened 2023-11-29
Add `mypy` to workflows


purva-thakre:
Add a check for mypy in https://github.com/vprusso/toqito/blob/200f3dc7a2178174d4ad6846f84686bc89c4c0ce/.github/workflows/build-test-actions.yml#L26

Right now, `~/toqito$ mypy .` fails due to following:

```
build/lib/tests/test_matrices/test_gell_mann.py:4: error: Skipping analyzing "scipy.sparse": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/tests/test_matrices/test_gell_mann.py:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
build/lib/tests/test_matrices/test_pauli.py:4: error: Skipping analyzing "scipy.sparse": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/channels/reduction.py:4: error: Skipping analyzing "scipy.sparse": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/matrices/gell_mann.py:5: error: Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/matrices/gen_gell_mann.py:6: error: Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/matrices/iden.py:4: error: Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/matrices/pauli.py:6: error: Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/matrix_props/sk_norm.py:8: error: Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/perms/antisymmetric_projection.py:6: error: Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/perms/perm_sign.py:6: error: Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
build/lib/toqito/perms/permute_systems.py:9: error: Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
[...]

PR #809, opened 2024-09-30
Fast exponentiation for tensor product

### LINKED ISSUE #803: Implementing tensor products of an object with itself using fast exponentiation
Currently, [the code for tensoring a matrix](https://github.com/vprusso/toqito/blob/dd08a53e76d600532fb7e8ec0be494ad8a8b05ef/toqito/matrix_ops/tensor.py#L135C1-L146C22) with itself is the following:
```python
    # Tensor product one matrix `n` times with itself.
    if len(args) == 2 and isinstance(args[1], int):
        num_tensor = args[1]
        if num_tensor == 1:
            return args[0]
        if num_tensor == 2:
            return np.kron(args[0], args[0])
        if num_tensor >= 3:
            result = np.kron(args[0], args[0])
            for _ in range(2, num_tensor):
                result = np.kron(result, args[0])
        return result
```
In particular, this is implemented in linear time. 

**Describe the solution you'd like**
Using fast exponentiation would reduce the number of operations performed, especially when using large number of tensors. Something along the lines of:

```python
    # Tensor product one matrix `n` times with itself.
    if len(args) == 2 and isinstance(args[1], int):
        num_tensor = args[1]
        if num_tensor == 1:
            return args[0]
        tensored_twice = np.kron(args[0], args[0])
        def fast_exponentiation(mat, t):
            if t == 1:
                return mat
            if t & 1:
                return np.kron(t, fast_exponentiation(tensored_twice, t >> 1))
            return fast_exponentation(tensored_twice, t >> 1)
        return fast_exponentiation(args[0], num_tensor)
```

Though maybe a bit overkill for the number of tensors we generally use, this can still save a few calls to `np.kron` for reasonable numbers of tensors.

tnemoz:
## Description
Implements the fact exponentiation algorithm to compute the tensor powers of a matrix. Fixes #803.

## Changes

  -  [x] Implements the fact exponentiation algorithm to compute the tensor powers of a matrix.

## Checklist


toqito/matrix_ops/tensor.py
```diff
@@ -135,15 +135,22 @@ def tensor(*args) -> np.ndarray:
...
+            if q == 1:
+                return matrix
+            tmp = fast_exp(tensored_twice, q >> 1)
+            if q & 1:
+                return np.kron(mat, tmp)
+            return tmp
+
+        return fast_exp(mat, num_tensor)
 
     # Tensor product between two or more matrices.
```
vprusso, line 155: Nice PR, @tnemoz ! 

How about considering the following layout:

```py
import numpy as np

def tensor(*args) -> np.ndarray:
    r"""Compute the Kronecker tensor product."""

    def fast_exp(matrix, q):
        """Efficient exponentiation by squaring."""
        if q == 1:
            return matrix
        tmp = fast_exp(matrix, q >> 1)
        tmp = np.kron(tmp, tmp)
        if q & 1:  # If q is odd
            tmp = np.kron(matrix, tmp)
        return tmp

    result = None

    # Input is provided as a list of numpy matrices.
    if (len(args) == 1 and isinstance(args[0], list)) or (len(args) == 1 and isinstance(args[0], np.ndarray)):
        if len(args[0]) == 1:
            return args[0][0]
        if len(args[0]) == 2:
            return np.kron(args[0][0], args[0][1])
        if len(args[0]) >= 3:
            result = args[0][0]
            for i in range(1, len(args[0])):
                result = np.kron(result, args[0][i])
        return result

    # Tensor product one matrix `n` times with itself.
    if len(args) == 2 and isinstance(args[1], int):
        num_tensor = args[1]
        if num_tensor == 0:
            return np.eye(args[0].shape[0])
        if num_tensor == 1:
            return args[0]
        return fast_exp(args[0], num_tensor)

    # Tensor product between two or more matrices.
    if len(args) == 2:
        return np.kron(args[0], args[1])
    if len(args) >= 3:
        result = args[0]
        for i in range(1, len(args)):
            result = np.kron(result, args[i])
        return result

    raise ValueError("The `tensor` function must take either a matrix or vector.")
```

Basically the main change(s) here are:

1. Moving the `fast_exp` function to the top of the `tensor` function to make it a bit more readable and less nested.
2. The updated fast_exp avoids unnecessary recomputation of the tensor product in the recursive step.
3. For n = 0, we now return the identity matrix, which aligns with the standard interpretation of a tensor product raised to the power of 0.

Any thoughts on this change?

tnemoz: > Nice PR

It isn't, I'm horrified to look back at the code I wrote :') I'm even a bit surprised it passes the test, I think I can add one that won't pass, there's a mistake in my code that isn't present in yours. Sorry for that! I'll modify ASAP.

Concerning the n=0 case, I'm all for returning something more canonic than None. However, I would've argued that the interpretation would have been to return `[[1]]`, since this is the neutral element for tensor products. In particular, by returning this, we keep the additivity property in the powers: $a^{\otimes n}\otimes a^{\otimes0}=a^{\otimes (n+0)}$, which wouldn't work with the identity  Or is there another interpretation that I don't know of?

vprusso: > It isn't, I'm horrified to look back at the code I wrote :') 

Completely disagree. In fact, I sincerely appreciate all of your efforts to contribute to |toqito>. On that note, thank you again for taking the time to add your code, and please be kinder to yourself!

> Concerning the n=0 case...

Yes, definitely. I think returning `None` is actually a bug and that `[[1]]` makes more sense. Making the change to reflect that would make sense if you are able to do so!

Thank you again for your efforts!

vprusso:
Nice. Thanks for those changes, @tnemoz ! 

That looks good to me, but I'll also defer to @purva-thakre for any additional comments that she may have. Thanks for your quick turn around, @tnemoz --great work as always!

purva-thakre (approved):
LGTM too! 

I just have one suggested change. Would it be possible to change `*args` in the user input field? In the docstring, we state that we allow 3 different types of inputs. Mentioning those types using `|` would make it easier to read the API. 

You will also have to make changes to  `:param args:`.

tnemoz:
> Would it be possible to change *args in the user input field? In the docstring, we state that we allow 3 different types of inputs. Mentioning those types using | would make it easier to read the API.

I'm unsure about how that would go to be fair. From what I now, if `*args` is type hinted, then it means that all possible additional arguments must be of that type. According to this rule, I don't see how it would be possible to assign a type to `args` here, since it could either by a collection of `np.ndarray` or a `np.ndarray` accompanied by an `int`. So this implementation:
```python
def tensor(*args: np.ndarray) -> np.ndarray:
```
doesn't work since the possibility of passing an `int` isn't represented. This one fixes the problem:
```python
def tensor(*args: np.ndarray | int | list[np.ndarray]) -> np.ndarray:
```
but it somehows also indicates that something like `tensor(2, 2)` is valid, while it doesn't really make sense. Also, it indicates that something like `tensor([e0, e1], [e0, e1])` should be accepted, while that isn't the case.

All in all, I'm not sure adding type hinting here would be beneficial, as it could indicate that several cases which should be valid, actually aren't.

That being said, I'm no expert in the topic. If you want to add the type hinting here, what is the approach that I should consider here?

purva-thakre:
> If you want to add the type hinting here, what is the approach that I should consider here?

So, this is a personal preference where I would rather user `named_param` rather than `*args` as the former provides easily readable details. 

Reading the docstring entry for `*args`, I would rename it to `input_mat`:

```py
def tensor(input_mat: list[np.ndarray]| np.ndarray, int| np.ndarray, ... , np.ndarray) -> np.ndarray:
```
Maybe you have to sandwich the last two allowed inputs in `()` or maybe not. 

We have a couple of functions like this and of course, I can't find them right now to link them here. lol.

purva-thakre:
Hi @tnemoz, for future scenarios, please make sure you create a branch from the master of your fork before you start working on a PR. This makes it easier for us to make changes to a PR branch, if needed.

I tried to check out this branch locally but I don't have permission to push changes into this.

purva-thakre:
Going to go ahead and merge this. The remaining discussion could be resolved as part of #299

tnemoz:
Hey @purva-thakre! Sorry for that :(
I'm not sure I understand the problem though: is it that I merged my master branch instead of a branch that I would have created from it (similar to my other current PR for instance)?

Something along the lines of "since my master branch is protected, you can't push on it"? Or not at all? Just to be sure I don't repeat the same mistake!

PR #813, opened 2024-10-01
Adding Dicke state function and tests.

### LINKED ISSUE #434: Feature: Dicke state
Would be nice to add `states/dicke_state.py`. 

For reference, QETLAB has a function [here](https://qetlab.com/DickeState) that implements this state. Also helpful, in Python, [this](https://github.com/Sunny-Zhu-613/pureb-public/blob/main/python/pyqet/dicke.py#L6) implementation might be able to be cleaned up and incorporated into `toqito` (with providing proper credit to both sources in either case). 

Creating the function in `states/dicke_state.py` would be part of the task, and the remaining part would be to add proper unit tests as well as a link in the `states.rst` file to the function.

vprusso:
Closes #434


toqito/states/tests/test_dicke.py
```diff
@@ -0,0 +1,50 @@
...
+        # Dicke state for 3 qubits and 1 excitation
+        (3, 1, np.array([0, 1/np.sqrt(3), 1/np.sqrt(3), 0, 1/np.sqrt(3), 0, 0, 0])),
+        # Dicke state for 4 qubits and 2 excitations (corrected expected state)
+        (4, 2, np.array([0, 0, 0, 1/np.sqrt(6), 0, 1/np.sqrt(6), 1/np.sqrt(6), 0,
+                         0, 1/np.sqrt(6), 1/np.sqrt(6), 0, 1/np.sqrt(6), 0, 0, 0])),
+    ],
+)
+def test_dicke_state(num_qubit, num_excited, expected_state):
+    """Test that dicke_state produces the correct vector state."""
+    np.testing.assert_array_almost_equal(dicke(num_qubit, num_excited), expected_state, decimal=6)
```
purva-thakre, line 21: we should not use `np.testing` as there's already an issue to move away from it.

vprusso: Good point, fixed in 0921559d.

PR #824, opened 2024-10-03
Fixing multiplication operator warning for extended nonlocal games.

vprusso:
Fixing the multiple warnings as shown in:
https://github.com/vprusso/toqito/actions/runs/11151893975/job/30996320621?pr=821#step:5:3446


purva-thakre:
The warning is still showing up.

vprusso:
> The warning is still showing up.

Yes, but this time at least it's only applicable to NLG (as opposed to ENLG) so I'll make the same switch for that!

purva-thakre (approved):

PR #826, opened 2024-10-04
Remove duplicate line rand orthonormal basis

vprusso:
Minor fixes to recent `random_orthonormal_basis` function


purva-thakre (approved):
Ahhh. Thanks for catching this!

PR #836, opened 2024-10-06
Changing from wagtail to rtd theme.

### LINKED ISSUE #649: Search inside documentation is broken
If I search for something in the search box of the docs, some suggestions show up. The link to these suggestions is [broken](https://toqito.readthedocs.io/en/latest/autoapi/matrices/gen_pauli_x/undefinedautoapi/matrices/gen_pauli_x/index.html). 

image

https://toqito.readthedocs.io/en/latest/autoapi/matrices/gen_pauli_x/index.html

vprusso:
Closes #649


purva-thakre:
I think we had decided to switch to a different theme because rtd does not allow a user to switch between light and dark modes. 

Have you tried using `sphinx <7.2` as suggested by the linked wagtail issue?

Let me try to search around for which other themes are available.

vprusso:
Personal preference, but I think RTD looks the best out of any of the themes that render: 
https://sphinx-themes.org/

On top of that, I noticed this theme also renders much better in my browser. Wagtail forced me to scroll down for certain API function pages. I think I'm happy to discard the light/dark option in favor of better page rendering and functional search. Thoughts?

purva-thakre:
Ok between the tradeoff of search documentation and light/dark mode. 

The thing that irks me about this theme is the blank white space on the right. Looking through the official documentation, it looks like that's how it is supposed to be.

vprusso:
> Ok between the tradeoff of search documentation and light/dark mode.
> 
> The thing that irks me about this theme is the blank white space on the right. Looking through the official documentation, it looks like that's how it is supposed to be.

Yeah, I think that's just how it is. Personally, I find this theme to be better than wagtail, but I guess that's personal preference. Let me know if you think I should go ahead with the update here (unless you're still trying out other themes)

purva-thakre:
> unless you're still trying out other themes

I made a note to do this before our weekly meeting.

purva-thakre:
What if we use `furo`? 

It looks similar to RTD without the blank space on the right. It also has the dark mode option that I tend to prefer. Search inside documentation also works similarly to RTD.

vprusso:
`furo` is pretty slick. Yeah, let's use that if we can. If you have the `furo` theme working on your end, feel free to hijack the PR. Otherwise, I can take it and swap the `rtd` theme out.

purva-thakre:
I will hijack the PR later today. I have it working on my end. Need to make a couple tweaks!

vprusso:
> I will hijack the PR later today. I have it working on my end. Need to make a couple tweaks!

Sounds good. Good find on the theme--I like it!

purva-thakre:
I don't know why the RTD build fails. Will try to figure out why we get the following:

```
Theme error:
no theme named 'furo' found (missing theme.toml?)
```

purva-thakre:
@vprusso this is ready

vprusso:
Nice find for the `req/docs.txt`. LGTM and nice theme choice! Great work, @purva-thakre !

PR #876, opened 2024-10-27
Parallelize classical value in nonlocal games

### LINKED ISSUE #12: Enhancement: Parallelize classical value of nonlocal game
The following loop inside of the `nonlocal_game.py` file under the `classical_value` function should be parallelized for large nonlocal games. 

```python
        for i in range(num_alice_outputs ** num_bob_inputs):
           ...
           tgval = np.sum(np.amax(pred_alice, axis=0))
            p_win = max(p_win, tgval)
        return p_win
```

This [parallelization is also performed in QETLAB](https://github.com/nathanieljohnston/QETLAB/blob/691036434e88a0bc26d6253d769d29a455dcb921/NonlocalGameValue.m#L128) where indeed the approach to calculating the classical value of a nonlocal game has been inspired from.

Shivansh20128:
## Description

Adding parallelism using `multiprocessing` python library. Added a new function `process_iteration` that provides reusability of code. the parallelization is done when `num_alice_outputs**num_bob_inputs` is more than 1000.
#Resolves #12

## Changes

  -  [x] Added a helper method to invoke multiprocessing when needed.
## Checklist

### DIFF
--- a/toqito/nonlocal_games/nonlocal_game.py
+++ b/toqito/nonlocal_games/nonlocal_game.py
@@ -179,14 +179,16 @@ class NonlocalGame:
 
         num_iterations = num_alice_outputs**num_bob_inputs
 
+        # we parallelize for large problems only
         if num_iterations > 1000:
-            with multiprocessing.Pool() as pool:
+            with multiprocessing.Pool() as pool:  # Creating a pool of processes for parallelization
                 tgvals = pool.starmap(
                     NonlocalGame.process_iteration,
                     [(i, num_bob_outputs, num_bob_inputs, pred_mat_copy, num_alice_outputs, num_alice_inputs)
                     for i in range(num_iterations)]
                 )
                 p_win = max(tgvals)
+        # Using single core implementation for small problems
         else:
             for i in range(num_iterations):
                 tgval = NonlocalGame.process_iteration(i, num_bob_outputs, num_bob_inputs, pred_mat_copy,
--- a/toqito/nonlocal_games/nonlocal_game.py
+++ b/toqito/nonlocal_games/nonlocal_game.py
@@ -1,5 +1,6 @@
 """Two-player nonlocal game."""
 
+import multiprocessing
 from collections import defaultdict
 
 import cvxpy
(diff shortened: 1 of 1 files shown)


toqito/nonlocal_games/nonlocal_game.py
```diff
@@ -120,6 +121,28 @@ def from_bcs_game(cls, constraints: list[np.ndarray], reps: int = 1) -> "Nonloca
 
         return cls(prob_mat, pred_mat, reps)
 
+    def process_iteration(i, num_bob_outputs, num_bob_inputs, pred_mat_copy, num_alice_outputs, num_alice_inputs):
```
vprusso, line 124: Could we include type-hints for the function arguments and return type?

toqito/nonlocal_games/nonlocal_game.py
```diff
@@ -153,20 +176,22 @@ def classical_value(self) -> float:
...
-            base = num_bob_outputs
-            digits = num_bob_inputs
-            b_ind = np.zeros(digits)
-            for j in range(digits):
-                b_ind[digits - j - 1] = np.mod(number, base)
-                number = np.floor(number / base)
-            pred_alice = np.zeros((num_alice_outputs, num_alice_inputs))
+        num_iterations = num_alice_outputs**num_bob_inputs
+
+        if num_iterations > 1000:
```
vprusso, line 181: It may be helpful to have a comment above the `if` and `else` statements stating when/why each of these gets invoked.

toqito/nonlocal_games/nonlocal_game.py
```diff
@@ -120,6 +121,28 @@ def from_bcs_game(cls, constraints: list[np.ndarray], reps: int = 1) -> "Nonloca
 
         return cls(prob_mat, pred_mat, reps)
 
+    def process_iteration(i, num_bob_outputs, num_bob_inputs, pred_mat_copy, num_alice_outputs, num_alice_inputs):
+        """Help the classical_value function as a helper method.
+
+        :return: A value between [0, 1] representing the tgval.
+        """
+        number = i
```
vprusso, line 129: I haven't tested this directly, but I think something like this might also work. Not sure if it's any better however:

```py
# Given parameters
number = i  # some integer value
base = num_bob_outputs
digits = num_bob_inputs

# Preallocate array for digit results
b_ind = np.zeros(digits)

# Compute each digit
for j in range(digits - 1, -1, -1):
    number, remainder = divmod(number, base)
    b_ind[j] = remainder
```

toqito/nonlocal_games/nonlocal_game.py
```diff
@@ -153,20 +176,22 @@ def classical_value(self) -> float:
...
-            digits = num_bob_inputs
-            b_ind = np.zeros(digits)
-            for j in range(digits):
-                b_ind[digits - j - 1] = np.mod(number, base)
-                number = np.floor(number / base)
-            pred_alice = np.zeros((num_alice_outputs, num_alice_inputs))
+        num_iterations = num_alice_outputs**num_bob_inputs
+
+        if num_iterations > 1000:
+            with multiprocessing.Pool() as pool:
```
vprusso, line 182: Do we want to maybe consider something like this? 

```py
# Determine tgvals based on iteration count
if num_iterations > 1000:
    with multiprocessing.Pool() as pool:
        tgvals = pool.starmap(
            NonlocalGame.process_iteration,
            [(i, num_bob_outputs, num_bob_inputs, pred_mat_copy, num_alice_outputs, num_alice_inputs)
             for i in range(num_iterations)]
        )
else:
    tgvals = [
        NonlocalGame.process_iteration(i, num_bob_outputs, num_bob_inputs, pred_mat_copy, num_alice_outputs, num_alice_inputs)
        for i in range(num_iterations)
    ]

# Compute maximum win probability
p_win = max(tgvals)
```

Shivansh20128:
> Out of curiosity, is there any indication as to how much faster the parallel version is compared to the single-core version?

For this function, my answer would be no. But I used `multiprocessing` library to run [this](https://github.com/Shivansh20128/Qiskit_OpenSource/blob/main/Issues/Parallel_processes.ipynb) Python notebook on my system. 
For the single core version, the time taken was about 25 seconds, while the parallel version took about 21 seconds.

vprusso:
> > Out of curiosity, is there any indication as to how much faster the parallel version is compared to the single-core version?
> 
> 
> 
> For this function, my answer would be no. But I used `multiprocessing` library to run [this](https://github.com/Shivansh20128/Qiskit_OpenSource/blob/main/Issues/Parallel_processes.ipynb) Python notebook on my system. 
> 
> For the single core version, the time taken was about 25 seconds, while the parallel version took about 21 seconds.
> 
> 

Interesting. I would have expected that the total time would have been subdivided by the number of cores on the machine, but perhaps 80 percent or so of the time is spent on the non parallelized parts of the process. 

Still, since it's faster and not much boilerplate for that boost, I'll still consider that a win!

toqito/nonlocal_games/nonlocal_game.py
Shivansh20128: added. done!

toqito/nonlocal_games/nonlocal_game.py
Shivansh20128: added. Done!

toqito/nonlocal_games/nonlocal_game.py
Shivansh20128: Yes, we can do this, but I think it doesn't affect the complexity. But if you want I can modify it to this.

toqito/nonlocal_games/nonlocal_game.py
Shivansh20128: It looks correct to me, but again, the time complexity won't be affected if we use this, as both the loops use the same number of iterations and the steps are the same (modulo operation has constant complexity). I think the only time we will save is not calculating the index `digits - j - 1`, but its insignificant.

```
number = i
        base = num_bob_outputs
        digits = num_bob_inputs
        b_ind = np.zeros(digits)

        for j in range(digits):
            b_ind[digits - j - 1] = np.mod(number, base)
            number = np.floor(number / base)

```

But again, I can modify the code if you want.
Thank you

vprusso: Right, `np.divmod` was used to get the quotient and remainder in a single step, effectively reducing the need for a for loop to compute each digit in the target base. Definitely not incredibly more efficient, so either way is fine by me!

toqito/nonlocal_games/nonlocal_game.py
vprusso: Yep. Since the tgval calculation is essentially the same in parallel and serial cases, you could refactor it into a single calculation to reduce redundancy. For instance, store the results in a list, then use max(tgvals) afterward. The way you do it is fine as well, so I'm okay with either approach!

toqito/nonlocal_games/nonlocal_game.py
Shivansh20128: I have made the suggested changes.

vprusso:
Thanks for your recent adaptations to the code, @Shivansh20128 ! 

This LGTM, but I'll defer to @purva-thakre for the final word. Thank you again for your contributions!

vprusso:
Thanks again, @Shivansh20128 !

Issue #767, opened 2024-08-30
Add checks for pseudo-unitary, pseudo-hermitian and pseudo-orthogonal in `matrix_props`


purva-thakre:
![image](https://github.com/user-attachments/assets/959aa5fb-0ce1-453a-9600-6a7fe0b9d392)

https://academic.oup.com/plms/article-abstract/s2-50/1/230/1543988

![image](https://github.com/user-attachments/assets/830ce739-e094-4ac4-b78d-5434119e40c8)

https://arxiv.org/pdf/math-ph/0302050

![image](https://github.com/user-attachments/assets/bae103dc-d167-4240-9539-515ec5665582)

https://nhigham.com/2021/04/28/what-is-a-pseudo-orthogonal-matrix/

harshvardhan-pandey:
@purva-thakre could you assign me the issue if no one is working on it?

Qubit1718:
Hi @purva-thakre , 

Created PR that solves this issue. Can you please take a look :)

purva-thakre:
Hi @Qubit1718 , I'd suggest you work with @harshvardhan-pandey as they asked the issue be assigned to them. Otherwise, I'd suggest you work on something else. 

I took a quick look. Your PR needs more work - examples in the docstrings, unit tests, proper way to add errors in Python (we do not use `return print("some error message.")`), etc.

Qubit1718:
Hi @purva-thakre , 

Thank you for your comments. I will work with @harshvardhan-pandey to make the extra changes.

harshvardhan-pandey:
Hi @Qubit1718
The task is fairly straightforward, and working together much is not possible. I'll try to make an appropriate PR ASAP. I'll let you know in case something comes up though :)

harshvardhan-pandey:
Had a small question regarding the definitions. The snippet with the definition of pseudo-unitary says when A is real and it follows the given condition, it is pseudo orthogonal. However, the definition of pseudo orthogonal doesn't constrain the signature matrix to have +1 and -1 entries separated. @purva-thakre could you confirm if the signature can be any diagonal matrix with diagonal entries being +1 or -1 or if it needs to have the structure highlighted in the first definition?

harshvardhan-pandey:
Also, for pseudo hermitian, do we need to check if the input is pseudo hermitian with respect to a given linear invertible operator or check if there exists such an operator? Because the latter seems like a particularly difficult problem.

vprusso:
[...]

PR #991, opened 2025-03-06
Checks for pseudo-unitary, pseudo-hermitian and pseudo-orthogonal in matrix_props 

Qubit1718:
This PR address the issue #767


toqito/matrix_props/is_pseudohermitian.py
```diff
@@ -0,0 +1,34 @@
+def is_pseudohermitian(mat: np.ndarray, signature_tup: tuple):
+    """
+    Checks if a matrix is pseudo-hermitian with respect to a given signature.
+
+    Args:
+        mat (np.ndarray): The matrix to check.
```
vprusso, line 6: If you type-hint in the function argument listing, this information in the doc-string is redundant and can be removed.

toqito/matrix_props/is_pseudohermitian.py
```diff
@@ -0,0 +1,34 @@
...
+        signature_tup (tuple): A tuple (p, q) where p is the number of +1's and q is the number of -1's in the metric.
+
+    Returns:
+        bool: True if the matrix is pseudo-hermitian, False otherwise.
+    """
+    # Check if the matrix is square
+    if mat.shape[0] != mat.shape[1]:
```
vprusso, line 13: We have a function for this that you can import from `toqito` (`from toqito.matrix_props import is_square`)

toqito/matrix_props/is_pseudohermitian.py
```diff
@@ -0,0 +1,34 @@
...
+        return print("Error: The signature does not match the matrix dimension.")
+        
+
+    # Construct the metric matrix
+    metric = np.diag([1] * p + [-1] * q)
+    inverse_metric = np.linalg.inv(metric)
+
+
+    # Use np.allclose for approximate equality
+    is_pseudo = np.allclose(metric @ mat @ inverse_metric, mat.conj().T)
```
vprusso, line 33: You can just return `np.allclose(metric @ mat @ inverse_metric, mat.conj().T)` for conciseness.

Also, we likely want to supply another argument to this function that dictates the precision. This should also then be passed to the `np.allclose` call as `atol` and/or `rtol` to define the absolute and relative tolerance. We can default these values to something reasonable. Refer to functions like `is_positive_semidefinite` in `toqito` for reference.

Qubit1718:
Hi @vprusso , 

I think both of us are working on the same feature and it seems like the other contributor's PR looks good and a better fit. I will try to work on the other issues. 



> I will also refer you to the MR here of another contributor for reference. It's likely that you'll want to ensure you're both not working on the same feature: #993

PR #993, opened 2025-03-06
Add is_pseudo_unitary

harshvardhan-pandey:
## Description

Added feature to check if a matrix is pseudo unitary (in regards to [Issue 767](https://github.com/vprusso/toqito/issues/767)).

## Changes

  -  Added is_pseudo_unitary
  -  Added test_is_pseudo_unitary
  -  Updated matrix_props/__init_.py

## Checklist

  -  [X] Use `ruff` for errors related to code style and formatting.
  -  [X] Verify all previous and newly added unit tests pass in `pytest`.
  -  [X] Check the documentation build does not lead to any failures. `Sphinx` build can be checked locally for any failures related to your PR
  -  [X] Use `linkcheck` to check for broken links in the documentation
  -  [X] Use `doctest` to verify the examples in the function docstrings work as expected.

### DIFF
--- a/toqito/matrix_props/is_pseudo_unitary.py
+++ b/toqito/matrix_props/is_pseudo_unitary.py
@@ -42,7 +42,7 @@ def is_pseudo_unitary(mat: np.ndarray, p: int, q: int, rtol: float = 1e-05, atol
     >>> from toqito.matrix_props import is_pseudo_unitary
     >>> A = np.array([[np.cosh(1), np.sinh(1)], [np.sinh(1), np.cosh(1)]])
     >>> is_pseudo_unitary(A, p=1, q=1)
-    True
+    np.True_
 
     However, the following matrix :math:B
 
@@ -56,7 +56,7 @@ def is_pseudo_unitary(mat: np.ndarray, p: int, q: int, rtol: float = 1e-05, atol
 
     >>> B = np.array([[1, 0], [1, 1]])
     >>> is_pseudo_unitary(B, p=1, q=1)
-    False
+    np.False_
 
     References
     ==========
--- a/toqito/matrix_props/is_pseudo_unitary.py
+++ b/toqito/matrix_props/is_pseudo_unitary.py
@@ -1,0 +1,88 @@
+"""Checks if matrix is pseudo unitary."""
+
+import numpy as np
+
+from toqito.matrix_props import is_square
+
+
+def is_pseudo_unitary(mat: np.ndarray, p: int, q: int, rtol: float = 1e-05, atol: float = 1e-08) -> bool:
+    r"""Check if a matrix is pseudo-unitary.
+
+    A matrix A of size (p+q)x(p+q) is pseudo-unitary with respect to a given signature matrix J if it satisfies
+
+    .. math::
+        A^* J A = J,
+
+    where:
+        - :math:A^* is the conjugate transpose (Hermitian transpose) of :math:A,
+        - :math:J is a diagonal matrix with first p diagonal matrix equal to 1 and next q diagonal entries equal to -1
+
+    Examples
+    ==========
+
+    Consider the following matrix:
+
+    .. math::
+        A = \frac{1}{\sqrt{2}} \begin{pmatrix}
+            cosh(1) & sinh(1) \\
+            sinh(1) & cosh(1)
+        \end{pmatrix}
+
+    with the signature matrix:
+
+    .. math::
+        J = \begin{pmatrix}
+            1 & 0 \\
+            0 & -1
+        \end{pmatrix}
+
+    Our function confirms that :math:A is pseudo-unitary:
+
+    >>> import numpy as np
+    >>> from toqito.matrix_props import is_pseudo_unitary
+    >>> A = np.array([[np.cosh(1), np.sinh(1)], [np.sinh(1), np.cosh(1)]])
+    >>> is_pseudo_unitary(A, p=1, q=1)
+    True
+
+    However, the following matrix :math:B
+
+    .. math::
+        B = \begin{pmatrix}
+            1 & 0 \\
+            1 & 1
+        \end{pmatrix}
+
+    is not pseudo-unitary with respect to the same signature matrix:
+
+    >>> B = np.array([[1, 0], [1, 1]])
+    >>> is_pseudo_unitary(B, p=1, q=1)
+    False
+
+    References
+    ==========
+
+    .. bibliography::
+        :filter: docname in docnames
+
+    :param mat: The matrix to check.
+    :param p: Number of positive entries in the signature matrix.
+    :param q: Number of negative entries in the signature matrix.
+    :param rtol: The relative tolerance parameter (default 1e-05).
+    :param atol: The absolute tolerance parameter (default 1e-08).
+    :raises ValueError: When p < 0 or q < 0.
+    :return: Return :code:True if the matrix is pseudo-unitary, and :code:False otherwise.
+
+    """
+    if p < 0 or q < 0:
+        raise ValueError("p and q must be non-negative")
+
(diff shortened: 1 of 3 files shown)


vprusso (approved):
I'll approve this, but I will away @purva-thakre 's review as well which may likely be next week sometime. Thanks for your contribution, @harshvardhan-pandey , and looking forward to seeing more of your work!

vprusso:
Ah, one thing I noticed is to request that you please add a test to cover all of the branches in the function. You can refer to the [code-cov report](https://app.codecov.io/gh/vprusso/toqito/pull/993?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vincent+Russo#66bb48c4a766fa26d7a723795d003ad0-R76) for where the missing coverage is:

vprusso:
> @vprusso I did update the issue of code coverage in a later commit. Though I don't know why the coverage percentage is still the same. Could you just confirm if there is still an issue on my end or if it takes some time to update?

Hmm, that's interesting, yes, I do see your tests for the lines in question. I believe you're correct and this is likely okay, but I'll wait for @purva-thakre to confirm when she has time. Thanks for pointing that out!

purva-thakre:
> though I don't know why the coverage percentage is still the same

this gets updated when all checks pass successfully. Until then, the report shows the older coverage %.

Since we had to approve and run the workflow for your PR, there was a bit of a lag between Vincent's feedback and the checks running successfully.

toqito/matrix_props/is_pseudo_unitary.py
```diff
@@ -0,0 +1,88 @@
...
+
+    """
+    if p < 0 or q < 0:
+        raise ValueError("p and q must be non-negative")
+
+    if not is_square(mat):
+        return False
+
+    if p + q != mat.shape[0]:
+        return False
```
purva-thakre, line 83: This should be a `ValueError` too.

toqito/matrix_props/tests/test_is_pseudo_unitary.py
```diff
@@ -0,0 +1,44 @@
...
+    """Test that random unitary matrix with signature diag([1, 1]) return True."""
+    mat = random_unitary(2)
+    np.testing.assert_equal(is_pseudo_unitary(mat, p=2, q=0), True)
+
+
+def test_is_pseudo_unitary_lorentz_boost():
+    """Test that Lorentz boost matrix with signature diag([1, -1]) returns True."""
+    theta = np.random.rand()
+    mat = np.array([[np.cosh(theta), np.sinh(theta)], [np.sinh(theta), np.cosh(theta)]])
+    np.testing.assert_equal(is_pseudo_unitary(mat, p=1, q=1), True)
```
purva-thakre, line 19: This could be combined into 1 test using https://docs.pytest.org/en/stable/how-to/parametrize.html. 

The only thing that's different is the input matrix and values of p & q. But both check if the provided matrix is unitary or pseudo-unitary.

toqito/matrix_props/tests/test_is_pseudo_unitary.py
```diff
@@ -0,0 +1,44 @@
...
+def test_is_not_pseudo_unitary_incorrect_signature_dimensions():
+    """Test that non-unitary matrix returns False."""
+    mat = np.array([[1, 0], [1, 1]])
+    np.testing.assert_equal(is_pseudo_unitary(mat, p=4, q=5), False)
+
+
+def test_is_pseudo_unitary_not_square():
+    """Input must be a square matrix."""
+    mat = np.array([[-1, 1, 1], [1, 2, 3]])
+    np.testing.assert_equal(is_pseudo_unitary(mat, p=1, q=1), False)
```
purva-thakre, line 37: same as my other comment about https://docs.pytest.org/en/stable/how-to/parametrize.html. 

The basic formula for 3 of these is the same.

toqito/matrix_props/is_pseudo_unitary.py
```diff
@@ -0,0 +1,88 @@
...
+    :param atol: The absolute tolerance parameter (default 1e-08).
+    :raises ValueError: When p < 0 or q < 0.
+    :return: Return :code:True if the matrix is pseudo-unitary, and :code:False otherwise.
+
+    """
+    if p < 0 or q < 0:
+        raise ValueError("p and q must be non-negative")
+
+    if not is_square(mat):
+        return False
```
purva-thakre, line 80: I wonder which option is better here:

- returning `False` for a check that is expected to be unsuccessful
- or raising a `ValueError`?

harshvardhan-pandey: `ValueError` seems a better fit. But then should we also change the `is_unitary` function for consistency?

purva-thakre: > ValueError seems a better fit

That's what I thought too. If the matrix is not square, then we already know it is not pseudounitary. 

> then should we also change the is_unitary function for consistency?

I think so. Pinging @vprusso for his thoughts.

vprusso: But we already have this check for the `is_unitary` function:
https://github.com/vprusso/toqito/blob/master/toqito/matrix_props/is_unitary.py#L79

vprusso: I'm not sure if `ValueError` is better. A non-square matrix shouldn't return a value error, as the value input is still a valid type (it's a numpy array). The matrix itself is just not unitary and should likely return `False` (instead of returning an error).

purva-thakre: > But we already have this check for the is_unitary function

https://github.com/vprusso/toqito/blob/38ff1f52d2696e0f1580973f0ce337218824c984/toqito/matrix_props/is_unitary.py#L79-L80

Yes, I think @harshvardhan-pandey is referring to how the function returns false when the matrix is not square. That's the consistency being referred to. If we decide to a raise an error for this function then we should do the same for is_unitary.

vprusso: > If we decide to a raise an error for this function then we should do the same for is_unitary.

Yes, but I don't think we should raise an error (if you refer to my follow-up comment above).

purva-thakre: Fine by me then. I just think there's no need to use this function if we already know the matrix is not a square matrix.

vprusso: Maybe I'm not following the conversation thread fully. 

>  I just think there's no need to use this function if we already know the matrix is not a square matrix.

Is "this function" `is_unitary`? Would we not want to check within that function whether it is square or not?

purva-thakre: > Is "this function" is_unitary?

yes. 

> Would we not want to check within that function whether it is square or not?

Definitely want to keep this. 

All I am saying is it seems kinda obvious if a non-square matrix is not unitary i.e. we don't really need a function to check this. Your previous comment is fine by me though. Ok with returning `False` for both functions in this scenario.

harshvardhan-pandey: @purva-thakre Then should the case where the matrix is square but it's not (p+q)x(p+q) also be flagged as False or should it be ValueError? Because even in then it is a valid input, it's just trivially non pseudo unitary.

purva-thakre: Flag it as False to be consistent with Vincent's comment.

harshvardhan-pandey:
I think everything works as it should. @purva-thakre could you confirm that I haven't missed anything.

purva-thakre (approved):

PR #1013, opened 2025-03-09
Added a function to generate random positive semidefinite operator.

### LINKED ISSUE #652: Generate random positive semidefinite operator
Allow the user to generate a random positive semidefinite (PSD) operator.

ShanmukhaKumar13:
## Description
Closes #652 

## Changes

  -  Added the function random_psd_operator that generates random positive semi definite operator.
  -  Added test_random_psd_operator


ShanmukhaKumar13:
@vprusso This PR implements the function to generate a random positive semidefinite operator and is ready for a review. Additionally, I have an issue with the documentation. In the example section the matrix shape is deformed. Please assist me in rectifying it.

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,59 @@
...
+    density matrix and then creates a symmetry matrix from it. Then the eigen values are made to their absolute values.
+    Then the positive semi definite operator is constructed by taking the dot product of the absolute eigen value matrix,
+    eigen vectors and eigen value transpose. Finally it is normalized to make its trace as 1. The resultant of the function will be a numpy ndarray 
+    which is positive semi-definite and unitary.
+
+    Example
+    ===========================
+    Generate a positive semi-definite operator of dimension 3.
+
+    >>> dim = 3
```
vprusso, line 21: The example should showcase the use of the function. Not a redefinition of the whole thing.

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,59 @@
...
+    >>> random_mat = np.random.rand(dim, dim)
+    >>> random_mat = (random_mat.T + random_mat) / 2
+    >>> eigenvalues, eigenvectors = np.linalg.eig(random_mat)
+    >>> eigenvalues = np.abs(eigenvalues)
+    >>> random_mat = np.dot(eigenvectors, np.dot(np.diag(eigenvalues), np.conj(eigenvectors).T))
+    >>> random_mat = random_mat / np.trace(random_mat)
+    >>> random_mat
+
+        [[0.35636099 0.17920518 0.14620245]
```
vprusso, line 30: If the matrices are not formatting well for you, refer to any other function that uses matrices in toqito for reference.

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,59 @@
...
+         [0.17920518 0.35165863 0.14467558]
+         [0.14620245 0.14467558 0.29198038]]
+
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
+    :param dim (int): The dimension of the operator.
```
vprusso, line 39: The types provided in the type-hinting args, so don't need to state these here.

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,59 @@
...
+
+    :return (np.ndarray): A random positive semi-definite operator with dimensions dim x dim.
+
+    :raises ValueError: If dim is not a positive integer.
```
vprusso, line 43: When does a value get raised? This comment is incorrect.

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,59 @@
+"Generates a random positive semi-definite operator."
+
+import numpy as np
+
+def random_psd_operator(dim: int) -> np.ndarray:
+    r''' Generate a random positive semi definite operator.
+
+    A positive semi-definite operator is a Hermitian operator that has only real and non-negative eigenvalues. 
+
+    This function generates a random positive semi-definite operator by constructing a symmetric matrix,
```
vprusso, line 10: This docstring needs to be written more clearly. For instance "eigenvalues" (not "eigen values"), statements like "The resultant of the function ..." could be worded more clearly, "creates a symmetry matrix" (it's not a "symmetry matrix" it's a "symmetric matrix", etc.

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,59 @@
+"Generates a random positive semi-definite operator."
+
+import numpy as np
+
+def random_psd_operator(dim: int) -> np.ndarray:
+    r''' Generate a random positive semi definite operator.
```
vprusso, line 6: Use `"""` not `'''`

vprusso, line 6: "semidefinite" not "semi definite"

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,59 @@
...
+
+    :raises ValueError: If dim is not a positive integer.
+
+    Notes:
+        The generated operator is symmetric and has non-negative eigenvalues.
+        The operator is normalized to have a trace of 1.
+
+    '''
+
+    random_mat = np.random.rand(dim, dim)
```
vprusso, line 51: The way this function is written is difficult to follow. Something more concise could be:

```py
    mat = (np.random.rand(dim, dim) + np.random.rand(dim, dim).T) / 2
    vals, vecs = np.linalg.eigh(mat)
    mat = vecs @ np.diag(np.abs(vals)) @ vecs.T
    return mat / np.trace(mat)
```

toqito/rand/tests/test_random_psd_operator.py
```diff
@@ -0,0 +1,37 @@
...
+def test_random_psd_operator(dim):
+    """Test for random_psd_operator function."""
+    # Generate a random positive semi-definite operator.
+    rand_psd_operator = random_psd_operator(dim)
+
+    # Ensure the matrix has the correct shape.
+    assert_equal(rand_psd_operator.shape, (dim, dim))
+
+    # Check if the matrix is a valid density matrix
+    assert is_density(rand_psd_operator), "Matrix should be a valid density matrix"
```
vprusso, line 31: Why does a random PSD matrix need to have trace 1? This would only be true if we were generating a random density matrix.

ShanmukhaKumar13:
Thank you for your feedback. I will review and address the suggested changes

toqito/rand/random_psd_operator.py
ShanmukhaKumar13: Sir, actually this line:

`mat = (np.random.rand(dim, dim) + np.random.rand(dim, dim).T) / 2`

creates two random independent matrices, which cannot guarantee a symmetric matrix.

To clarify:

- np.random.rand(dim, dim) generates a random matrix.
- np.random.rand(dim, dim).T generates another random matrix and takes its transpose.
- Adding these two matrices and scaling by 1/2 does not ensure symmetry.

A better approach to generate a symmetric matrix would be:

`mat = np.random.rand(dim, dim)`
`mat = (mat + mat.T) / 2`

This ensures that the resulting matrix is symmetric.

vprusso (commented):
Some minor comments, but it's coming along well, great work so far, @ShanmukhaKumar13 !

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,76 @@
...
+
+    >>> from toqito.matrix_props import is_positive_semidefinite
+    >>> is_positive_semidefinite(complex_psd_mat)
+    True
+
+    We can also generate random positive semidefinite matrices that are real-valued as follows.
+
+    >>> from toqito.rand import random_density_matrix
+    >>> real_psd_mat = random_density_matrix(2, is_real=True)
+    >>> real_psd_mat # doctest: +SKIP
```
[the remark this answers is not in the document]
purva-thakre: @vprusso If seed is not used, the generated random density matrix will be different for each run. Which is why these checks were skipped initially.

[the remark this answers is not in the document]
vprusso: Gotcha. We should likely fix the seed in the input parameters then and check against that matrix then, right?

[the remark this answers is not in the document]
purva-thakre: Yes, `seed` should be an optional param for the function.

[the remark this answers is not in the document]
purva-thakre: @ShanmukhaKumar13 In case it isn't clear, `seed` is used to control the output of a random function. We want you add this as an optional param to your function. 

If you are not using seed in your docstring examples, then you will need to use `SKIP` to make sure doctest skips the test. But if you use seed, then you don't need to have doctest skip the example. 

PTAL at `toqito/rand/random_unitary.py` for an example.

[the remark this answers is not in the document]
ShanmukhaKumar13: Sure, madam. I'm working on it.

toqito/rand/random_psd_operator.py
```diff
@@ -0,0 +1,84 @@
...
+
+    A positive semidefinite operator is a Hermitian operator that has only real and non-negative eigenvalues.
+    This function generates a random positive semidefinite operator by constructing a Hermitian matrix,
+    based on the fact that a Hermitian matrix can have real eigenvalues.
+
+    Examples
+    ========
+
+    Using :code:`toqito`, we may generate a random positive semidefinite matrix.
+    For :math:`dim=2`, this can be accomplished as follows.
```
purva-thakre, line 21:
```suggestion
    For :math:`\text{dim}=2`, this can be accomplished as follows.
```

vprusso (approved):
This LGTM! I'll also await the input from @purva-thakre as well. Nice work, @ShanmukhaKumar13 !

purva-thakre (approved):

PR #1026, opened 2025-03-13
Adding a function that finds the upper bound for a given bipartite Bell inequality

### LINKED ISSUE #563: Feature: Bell inequality upper bound for qubits
I have an implementation that is close to being complete for upper bounding the value of a qubit Bell inequality. The logic is heavily derived from the nice implementation [here](https://github.com/nathanieljohnston/QETLAB/blob/master/BellInequalityMaxQubits.m) in QETLAB.

```py
import cvxpy as cp
from scipy.sparse import eye
from itertools import combinations
from toqito.perms import permutation_operator, swap
from toqito.channels import partial_transpose

def MN_matrix(m, a, x):
    """
    Computes the matrices M_a^x and N_b^y.
    
    Args:
        m (int): The number of measurement settings for Alice and Bob.
        a (int): The specific measurement setting for Alice.
        x (int): The specific measurement setting for Bob.
    
    Returns:
        MN (sparse matrix): The computed matrix.
    """
    # Create a permutation list as done in MATLAB
    perm = list(range(1, m+2))  # MATLAB indices start from 1
    perm[0] = x + 1
    perm[x] = 1
    
    # Generate the identity matrix and the permutation matrix
    I = a * eye(2**(m+1), format='csc')  # Using compressed sparse column format
    P = permutation_operator([2]*(m+1), perm, 0, 1)
    
    # Calculate the matrix as in the MATLAB code
    MN = I + ((-1)**a) * P
    
    return MN


def bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val):
    ma, mb = joint_coe.shape
    oa = len(a_val)
    ob = len(b_val)

    # Ensure the input vectors are column vectors
    a_val = a_val.reshape(-1, 1)
    b_val = b_val.reshape(-1, 1)
    a_coe = a_coe.reshape(-1, 1)
    b_coe = b_coe.reshape(-1, 1)

    # Check if vectors a_val and b_val have only two elements
    if oa != 2 or ob != 2:
        raise ValueError("This script is only capable of handling Bell inequalities with two outcomes.")

    m = ma
    tot_dim = 2 ** (2 * m + 2)
    obj_mat = np.zeros((tot_dim, tot_dim), dtype=float)


    # Nested loops to compute the objective matrix
    for a in range(2):  # a = 0 to 1
        for b in range(2):  # b = 0 to 1
            for x in range(1, m+1):  # x = 1 to m (1-indexed in MATLAB, hence the range adjustment)
                for y in range(1, m+1):  # y = 1 to m
                    b_coeff = joint_coe[x-1, y-1] * a_val[a, 0] * b_val[b, 0]  # Adjust index for 0-based Python indexing
                    if y == 1:
                        b_coeff += a_coe[x-1, 0] * a_val[a, 0]  # Adjust for 0-based indexing
                    if x == 1:
                        b_coeff += b_coe[y-1, 0] * b_val[b, 0]  # Adjust for 0-based indexing
                    
                    # Adding the result of the tensor product to the objective matrix
                    MN_a_x = MN_matrix(m, a, x)  # Assuming MN_matrix function is already defined
                    MN_b_y = MN_matrix(m, b, y)
                    obj_mat += b_coeff * np.kron(MN_a_x, MN_b_y)  # Ensure it's in the correct format

    # Symmetrize the matrix to avoid numerical issues
    obj_mat = (obj_mat + obj_mat.T) / 2

    # Define auxiliary matrix (not used in the computation here)
    aux_mat = np.array([[1, 0, 0, 0],
                        [0, 0, 0, 0],
                        [0, 0, 0, 0],
                        [0, 0, 0, 0]])

    W = cp.Variable((2**(2*m), 2**(2*m)), symmetric=True)

    M = swap(W, [2, m+1], [2] * (2*m))
    X = swap(cp.kron(M, aux_mat), [m+1, 2*m+1], [2] * (2*m+2))
    Z = swap(X, [m+2, 2*m+1], [2] * (2*m+2))

    objective = cp.Maximize(cp.trace(Z @ obj_mat))

    # Define the constraints
    constraints = [cp.trace(W) == 1, W >> 0]

    # Adding PPT constraints
    for sz in range(1, m):
        # Generate all combinations of indices from 1 to 2*m-1 of size sz
        for ppt_partition in combinations(range(1, 2 * m-1), sz):
            # Convert to 0-based indexing for Python
            ppt_partition = [x - 1 for x in ppt_partition]
            # Partial transpose on the partition, ensuring it's positive semidefinite
            pt_matrix = partial_transpose(W, ppt_partition, [4] + [2] * (2 * (m - 1)))
            constraints.append(pt_matrix >> 0)

    # Solve the problem
    prob = cp.Problem(objective, constraints)
    prob.solve(solver="MOSEK", verbose=True) 

    # Return the results
    rho = W.value
    bmax = prob.value

    return bmax
```

Indeed, following the example from [this page](https://qetlab.com/BellInequalityMaxQubits), it is possible to replicate the answer for the I3322 Bell inequality:

```py
# Inputs:
joint_coe = np.array([
    [1, 1, -1], 
    [1, 1, 1],
    [-1, 1, 0]
])
a_coe = np.array([0, -1, 0])
b_coe = np.array([-1, -2, 0])
a_val = np.array([0, 1])
b_val = np.array([0, 1])

print(bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val))
>>> 0.25000
```

Great. However, it seems like this fails for another inequality (like something simple, i.e. the CHSH inequality):

```py
# Inputs for CHSH
joint_coe = np.array([
    [1, 1],
    [1, -1]
])
a_coe = np.array([0, 0])  # No single-party terms for CHSH
b_coe = np.array([0, 0])

a_val = np.array([1, -1])
b_val = np.array([1, -1])

print(bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val))
>>> 3.089617744060435
```

Which is definitely wrong (it should be $2 \sqrt{2} \approx 2.84$.

I think this is close, but it's hard to diagnose what the issue is without a MATLAB license.

EthanFeld:
## Description

Fixes #563 

Added the implementation for finding the upper bound for the maximum value of a Bell inequality(Issue 563)

## Changes

Added bell_inequality_max
Added test_bell_inequality_max
Updated matrix_props/_init.py
Updated refs.bib

## Checklist

### DIFF
--- a/toqito/state_opt/bell_inequality_max.py
+++ b/toqito/state_opt/bell_inequality_max.py
@@ -39,7 +39,7 @@ def bell_inequality_max(
 
 
     Consider the I3322 Bell inequality
-    
+
     .. math::
 
         \begin{aligned}
@@ -49,7 +49,7 @@ def bell_inequality_max(
         \end{aligned}
 
     The individual and joint coefficents and measurement values are encoded as matrices.
-    The upper bound can then be found in :code:'toqito' as follows.
+    The upper bound can then be found in :code:`toqito` as follows.
 
     >>> from toqito.state_opt import bell_inequality_max
     >>> import numpy as np
@@ -71,12 +71,12 @@ def bell_inequality_max(
         :filter: docname in docnames
 
     :raises ValueError: If a_val or b_val are not length 2.
-    :param joint_coe: The coefficents for terms containing both A and B
-    :param a_coe: The coefficent for terms only containing A
-    :param b_coe: The coefficent for terms only containing B
-    :param a_val: The value of each measurement outcome for A
-    :param b_val: The value of each measurement outcome for B
-    :return: The upper bound for the maximum violation of the Bell inequality
+    :param joint_coe: The coefficents for terms containing both A and B.
+    :param a_coe: The coefficent for terms only containing A.
+    :param b_coe: The coefficent for terms only containing B.
+    :param a_val: The value of each measurement outcome for A.
+    :param b_val: The value of each measurement outcome for B.
+    :return: The upper bound for the maximum violation of the Bell inequality.
 
     """
     m, _ = joint_coe.shape
--- a/toqito/state_opt/bell_inequality_max.py
+++ b/toqito/state_opt/bell_inequality_max.py
@@ -1,45 +1,27 @@
-"""Computes the upper bound for a given Bipartite bell inequality."""
+"""Computes the upper bound for a given bipartite Bell inequality."""
 
 from itertools import combinations
 
 import cvxpy as cp
-import mosek
 import numpy as np
-from scipy.sparse import eye
 
 from toqito.channels import partial_transpose
 from toqito.perms import permutation_operator, swap
 
 
-def MN_matrix(m: int, a: int, x: int) -> np.ndarray:
-    r"""Compute the matrices M_a^x and N_b^y.
-
-    Args:
-        m: The number of measurement settings for Alice and Bob.
-        a: The specific measurement setting for Alice.
-        x: The specific measurement setting for Bob.
-
-    Returns:
-        The computed matrix.
-
-    """
-    # Create a permutation list as done in MATLAB
-    perm = list(range(m + 1))  # MATLAB indices start from 1
-    perm[0] = x
-    perm[x] = 0
-
-    # Calculate the matrix as in the MATLAB code
-    MN = a * np.eye(2 ** (m + 1)) + ((-1) ** a) * permutation_operator([2] * (m + 1), perm, 0, 1)
-
-    return MN
-
-
-def bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val):
+def bell_inequality_max(
+    joint_coe: np.ndarray,
+    a_coe: np.ndarray,
+    b_coe: np.ndarray,
+    a_val: np.ndarray,
+    b_val: np.ndarray,
+    solver_name: str = "SCS",
+) -> float:
     r"""Return the upper bound for the maximum violation(Tsirelson Bound) for a given bipartite Bell Inequality.
 
-    This computes the upper bound for the maximum value of a given bipartite Bell Inequality using SDP.
+    This computes the upper bound for the maximum value of a given bipartite Bell Inequality using an SDP.
     The method is from :cite:`Navascues_2014_Characterization` and the implementation is based on :cite:`QETLAB_link`.
-    This is useful for various tasks in device independent quantum information processing
+    This is useful for various tasks in device independent quantum information processing.
 
     The function formulates the problem as a SDP problem in the following format
 
@@ -60,7 +42,7 @@ def bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val):
 
 
     Consider the I3322 Bell inequality
-    . . math::
+    .. math::
     I_{3322} = P(A_1 = B_1) + P(B_1 = A_2) + P(A_2 = B_2) + P(B_2 = A_3)
            - P(A_1 = B_2) - P(A_2 = B_3) - P(A_3 = B_1) - P(A_3 = B_3)
            \leq 2
@@ -77,8 +59,8 @@ def bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val):
     ... ])
     >>> a_coe = np.array([0, -1, 0])
     >>> b_coe = np.array([-1, -2, 0])
-    >>> a_val = np.array([0,1])
-    >>> b_val = np.array([0,1])
+    >>> a_val = np.array([0, 1])
+    >>> b_val = np.array([0, 1])
     >>> '%.3f' % bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val)
     '0.250'
 
@@ -95,74 +77,75 @@ def bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val):
     :param b_val: The value of each measurement outcome for B
     :return: The upper bound for the maximum violation of the Bell inequality
 
-
-
     """
     m, _ = joint_coe.shape
-    oa = len(a_val)
-    ob = len(b_val)
 
-    # Ensure the input vectors are column vectors
+    # Ensure the input vectors are column vectors.
     a_val = a_val.reshape(-1, 1)
     b_val = b_val.reshape(-1, 1)
     a_coe = a_coe.reshape(-1, 1)
     b_coe = b_coe.reshape(-1, 1)
 
-    # Check if vectors a_val and b_val have only two elements
-    if oa != 2 or ob != 2:
+    # Check if vectors a_val and b_val have only two elements.
+    if len(a_val) != 2 or len(b_val) != 2:
         raise ValueError("This script is only capable of handling Bell inequalities with two outcomes.")
 
     tot_dim = 2 ** (2 * m + 2)
     obj_mat = np.zeros((tot_dim, tot_dim), dtype=float)
 
-    # Nested loops to compute the objective matrix
+    # Nested loops to compute the objective matrix.
     for a in range(2):  # a = 0 to 1
         for b in range(2):  # b = 0 to 1
+            # Indicies below are adjusted to account for Python-MATLAB difference.
             for x in range(1, m + 1):  # x = 1 to m (1-indexed in MATLAB, hence the range adjustment)
                 for y in range(1, m + 1):  # y = 1 to m
-                    b_coeff = (
-                        joint_coe[x - 1, y - 1] * a_val[a, 0] * b_val[b, 0]
-                    )  # Adjust index for 0-based Python indexing
+                    b_coeff = joint_coe[x - 1, y - 1] * a_val[a, 0] * b_val[b, 0]
                     if y == 1:
-                        b_coeff += a_coe[x - 1, 0] * a_val[a, 0]  # Adjust for 0-based indexing
+                        b_coeff += a_coe[x - 1, 0] * a_val[a, 0]
                     if x == 1:
-                        b_coeff += b_coe[y - 1, 0] * b_val[b, 0]  # Adjust for 0-based indexing
+                        b_coeff += b_coe[y - 1, 0] * b_val[b, 0]
 
-                    # Adding the result of the tensor product to the objective matrix
-                    obj_mat += b_coeff * np.kron(MN_matrix(m, a, x), MN_matrix(m, b, y))
+                    # Construct Alice and Bob's extended measurement operators.
+                    perm_x = [x if i == 0 else (0 if i == x else i) for i in range(m + 1)]
+                    perm_y = [y if i == 0 else (0 if i == y else i) for i in range(m + 1)]
+                    M = a * np.eye(2 ** (m + 1)) + ((-1) ** a) * permutation_operator([2] * (m + 1), perm_x, 0, 1)
+                    N = b * np.eye(2 ** (m + 1)) + ((-1) ** b) * permutation_operator([2] * (m + 1), perm_y, 0, 1)
 
-    # Symmetrize the matrix to avoid numerical issues
+                    # Adding the result of the tensor product to the objective matrix.
+                    obj_mat += b_coeff * np.kron(M, N)
+
+    # Symmetrize the matrix to avoid numerical issues.
     obj_mat = (obj_mat + obj_mat.T) / 2
     aux_mat = np.array([[1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]])
 
-    # Construct the SDP problem
+    # Construct the SDP problem.
     W = cp.Variable((2 ** (2 * m), 2 ** (2 * m)), symmetric=True)
 
+    # Dimension boost W to the same dimension as obj_mat.
     M = swap(W, [2, m + 1], [2] * (2 * m))
     X = swap(cp.kron(M, aux_mat), [m + 1, 2 * m + 1], [2] * (2 * m + 2))
     Z = swap(X, [m + 2, 2 * m + 1], [2] * (2 * m + 2))
 
     objective = cp.Maximize(cp.trace(Z @ obj_mat))
 
-    # Define the constraints
+    # Define the constraints.
     constraints = [cp.trace(W) == 1, W >> 0]
 
-    # Adding PPT constraints
+    # Adding PPT constraints.
     for sz in range(1, m + 1):
-        # Generate all combinations of indices from 1 to 2*m-1 of size sz
+        # Generate all combinations of indices from 1 to 2*m-1 of size sz.
         for ppt_partition in combinations(range(1, 2 * m - 1), sz):
-            # Convert to 0-based indexing for Python
+            # Convert to 0-based indexing for Python.
             ppt_partition_updated = [x - 1 for x in ppt_partition]
-            # Partial transpose on the partition, ensuring it's positive semidefinite
+            # Partial transpose on the partition, ensuring it's positive semidefinite.
             pt_matrix = partial_transpose(W, ppt_partition_updated, [4] + [2] * (2 * (m - 1)))
             constraints.append(pt_matrix >> 0)
 
-    # Solve the problem
+    # Solve the problem.
     prob = cp.Problem(objective, constraints)
-    prob.solve(solver="MOSEK", verbose=False)
+    prob.solve(solver=solver_name, verbose=False)
 
-    # Return the results
-    rho = W.value
+    # Return the results.
     bmax = prob.value
 
     return bmax
--- a/toqito/state_opt/bell_inequality_max.py
+++ b/toqito/state_opt/bell_inequality_max.py
@@ -26,17 +26,13 @@ def bell_inequality_max(
     The function formulates the problem as a SDP problem in the following format
 
     .. math::
-    
-        \[
-        \max \operatorname{tr} \left( W \cdot \sum_{a,b,x,y} B^{xy}_{ab} M^x_a \otimes N^y_b \right),
-        \]
-        \[
-        \text{s.t.} \quad \operatorname{tr}(W) = 1, \quad W \geq 0,
-        \]
-        \[
-        W^{T_P} \geq 0, \quad \text{for all bipartitions } P.
-        \]
-        \]
+
+        \begin{multline}
+        \max \operatorname{tr}\!\Bigl( W \cdot \sum_{a,b,x,y} B^{xy}_{ab}\, M^x_a \otimes N^y_b \Bigr),\\[1ex]
+        \text{s.t.} \quad \operatorname{tr}(W) = 1,\quad W \ge 0,\\[1ex]
+        W^{T_P} \ge 0,\quad \text{for all bipartitions } P.
+        \end{multline}
+
 
     Examples
     =======
@@ -44,10 +40,12 @@ def bell_inequality_max(
 
     Consider the I3322 Bell inequality
     .. math::
-    
-    I_{3322} = P(A_1 = B_1) + P(B_1 = A_2) + P(A_2 = B_2) + P(B_2 = A_3)
-           - P(A_1 = B_2) - P(A_2 = B_3) - P(A_3 = B_1) - P(A_3 = B_3)
-           \leq 2
+
+        \begin{aligned}
+        I_{3322} &= P(A_1 = B_1) + P(B_1 = A_2) + P(A_2 = B_2) + P(B_2 = A_3) \\
+                 &\quad - P(A_1 = B_2) - P(A_2 = B_3) - P(A_3 = B_1) - P(A_3 = B_3) \\
+                 &\le 2
+        \end{aligned}
 
     The individual and joint coefficents and measurement values are encoded as matrices.
     The upper bound can then be found in :code:'toqito' as follows.
--- a/toqito/state_opt/bell_inequality_max.py
+++ b/toqito/state_opt/bell_inequality_max.py
@@ -76,6 +76,7 @@ def bell_inequality_max(
     :param b_coe: The coefficent for terms only containing B.
     :param a_val: The value of each measurement outcome for A.
     :param b_val: The value of each measurement outcome for B.
+    :param solver_name: The solver used.
     :return: The upper bound for the maximum violation of the Bell inequality.
 
     """
--- a/toqito/state_opt/bell_inequality_max.py
+++ b/toqito/state_opt/bell_inequality_max.py
@@ -26,6 +26,7 @@ def bell_inequality_max(
     The function formulates the problem as a SDP problem in the following format
 
     .. math::
+    
         \[
         \max \operatorname{tr} \left( W \cdot \sum_{a,b,x,y} B^{xy}_{ab} M^x_a \otimes N^y_b \right),
         \]
@@ -43,6 +44,7 @@ def bell_inequality_max(
 
     Consider the I3322 Bell inequality
     .. math::
+    
     I_{3322} = P(A_1 = B_1) + P(B_1 = A_2) + P(A_2 = B_2) + P(B_2 = A_3)
            - P(A_1 = B_2) - P(A_2 = B_3) - P(A_3 = B_1) - P(A_3 = B_3)
            \leq 2
--- a/toqito/state_opt/bell_inequality_max.py
+++ b/toqito/state_opt/bell_inequality_max.py
@@ -1,0 +1,168 @@
+"""Computes the upper bound for a given Bipartite bell inequality."""
+
+from itertools import combinations
+
+import cvxpy as cp
+import mosek
+import numpy as np
+from scipy.sparse import eye
+
+from toqito.channels import partial_transpose
+from toqito.perms import permutation_operator, swap
+
+
+def MN_matrix(m: int, a: int, x: int) -> np.ndarray:
+    r"""Compute the matrices M_a^x and N_b^y.
+
+    Args:
+        m: The number of measurement settings for Alice and Bob.
+        a: The specific measurement setting for Alice.
+        x: The specific measurement setting for Bob.
+
+    Returns:
+        The computed matrix.
+
+    """
+    # Create a permutation list as done in MATLAB
+    perm = list(range(m + 1))  # MATLAB indices start from 1
+    perm[0] = x
+    perm[x] = 0
+
+    # Calculate the matrix as in the MATLAB code
+    MN = a * np.eye(2 ** (m + 1)) + ((-1) ** a) * permutation_operator([2] * (m + 1), perm, 0, 1)
+
+    return MN
+
+
+def bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val):
+    r"""Return the upper bound for the maximum violation(Tsirelson Bound) for a given bipartite Bell Inequality.
+
+    This computes the upper bound for the maximum value of a given bipartite Bell Inequality using SDP.
+    The method is from :cite:`Navascues_2014_Characterization` and the implementation is based on :cite:`QETLAB_link`.
+    This is useful for various tasks in device independent quantum information processing
+
+    The function formulates the problem as a SDP problem in the following format
+
+    .. math::
+        \[
+        \max \operatorname{tr} \left( W \cdot \sum_{a,b,x,y} B^{xy}_{ab} M^x_a \otimes N^y_b \right),
+        \]
+        \[
+        \text{s.t.} \quad \operatorname{tr}(W) = 1, \quad W \geq 0,
+        \]
+        \[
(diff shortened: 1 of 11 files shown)


toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+from scipy.sparse import eye
+
+from toqito.channels import partial_transpose
+from toqito.perms import permutation_operator, swap
+
+
+def MN_matrix(m: int, a: int, x: int) -> np.ndarray:
+    r"""Compute the matrices M_a^x and N_b^y.
+
+    Args:
```
purva-thakre, line 17: Hi! Please take a look at the 'contributing' section of the documentation. We have a specific style for docstrings params etc. 

https://toqito.readthedocs.io/en/latest/contributing.html#references-in-docstrings

purva-thakre: Also, you need to add unit tests too to check the function you added works as expected. 

Are you a beginner to open-source? I can provide some links on how to write unit tests.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
+"""Computes the upper bound for a given Bipartite bell inequality."""
```
vprusso, line 1: "bipartite" should be lowercase and "Bell" should be uppercase.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
+"""Computes the upper bound for a given Bipartite bell inequality."""
+
+from itertools import combinations
+
+import cvxpy as cp
+import mosek
```
vprusso, line 6: This import is unused and not necessary.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
+"""Computes the upper bound for a given Bipartite bell inequality."""
+
+from itertools import combinations
+
+import cvxpy as cp
+import mosek
+import numpy as np
+from scipy.sparse import eye
```
vprusso, line 8: This is not used and can be removed.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+            # Partial transpose on the partition, ensuring it's positive semidefinite
+            pt_matrix = partial_transpose(W, ppt_partition_updated, [4] + [2] * (2 * (m - 1)))
+            constraints.append(pt_matrix >> 0)
+
+    # Solve the problem
+    prob = cp.Problem(objective, constraints)
+    prob.solve(solver="MOSEK", verbose=False)
+
+    # Return the results
+    rho = W.value
```
vprusso, line 165: `rho` does not appear to be used.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+        :filter: docname in docnames
+
+    :raises ValueError: If a_val or b_val are not length 2.
+    :param joint_coe: The coefficents for terms containing both A and B
+    :param a_coe: The coefficent for terms only containing A
+    :param b_coe: The coefficent for terms only containing B
+    :param a_val: The value of each measurement outcome for A
+    :param b_val: The value of each measurement outcome for B
+    :return: The upper bound for the maximum violation of the Bell inequality
+
```
vprusso, line 97: There is lots of extra spacing here that can be removed.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+    >>> from toqito.state_opt import bell_inequality_max
+    >>> import numpy as np
+    >>> joint_coe = np.array([
+    ... [1, 1, -1],
+    ... [1, 1, 1],
+    ... [-1, 1, 0]
+    ... ])
+    >>> a_coe = np.array([0, -1, 0])
+    >>> b_coe = np.array([-1, -2, 0])
+    >>> a_val = np.array([0,1])
```
vprusso, line 80: [nit]: Can you add a space between the comma? So `np.array([0, 1])`, etc.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+        W^{T_P} \geq 0, \quad \text{for all bipartitions } P.
+        \]
+        \]
+
+    Examples
+    =======
+
+
+    Consider the I3322 Bell inequality
+    . . math::
```
vprusso, line 63: It should be `.. math::` not `. . math::`.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+        for ppt_partition in combinations(range(1, 2 * m - 1), sz):
+            # Convert to 0-based indexing for Python
+            ppt_partition_updated = [x - 1 for x in ppt_partition]
+            # Partial transpose on the partition, ensuring it's positive semidefinite
+            pt_matrix = partial_transpose(W, ppt_partition_updated, [4] + [2] * (2 * (m - 1)))
+            constraints.append(pt_matrix >> 0)
+
+    # Solve the problem
+    prob = cp.Problem(objective, constraints)
+    prob.solve(solver="MOSEK", verbose=False)
```
vprusso, line 162: We shouldn't be using Mosek by default as this is a proprietary solver that users may not have access to. 

Instead, we can make the solver an optional argument parameter to the `bell_inequality_max` function. Something like `solver: str = "cvxopt"` should be included in the argument parameter listing.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+    perm[0] = x
+    perm[x] = 0
+
+    # Calculate the matrix as in the MATLAB code
+    MN = a * np.eye(2 ** (m + 1)) + ((-1) ** a) * permutation_operator([2] * (m + 1), perm, 0, 1)
+
+    return MN
+
+
+def bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val):
```
vprusso, line 37: We should include type-hints for both the arguments here and for the return of the function as well.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+    r"""Return the upper bound for the maximum violation(Tsirelson Bound) for a given bipartite Bell Inequality.
+
+    This computes the upper bound for the maximum value of a given bipartite Bell Inequality using SDP.
+    The method is from :cite:`Navascues_2014_Characterization` and the implementation is based on :cite:`QETLAB_link`.
+    This is useful for various tasks in device independent quantum information processing
```
vprusso, line 42: [nit]: Period at end of sentence.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+    # Calculate the matrix as in the MATLAB code
+    MN = a * np.eye(2 ** (m + 1)) + ((-1) ** a) * permutation_operator([2] * (m + 1), perm, 0, 1)
+
+    return MN
+
+
+def bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val):
+    r"""Return the upper bound for the maximum violation(Tsirelson Bound) for a given bipartite Bell Inequality.
+
+    This computes the upper bound for the maximum value of a given bipartite Bell Inequality using SDP.
```
vprusso, line 40: [nit]: using "an" SDP

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+    :param b_coe: The coefficent for terms only containing B
+    :param a_val: The value of each measurement outcome for A
+    :param b_val: The value of each measurement outcome for B
+    :return: The upper bound for the maximum violation of the Bell inequality
+
+
+
+    """
+    m, _ = joint_coe.shape
+    oa = len(a_val)
```
vprusso, line 102: There's not really a reason to define these as separate variables as we can just check the length where necessary (see the comment below).

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+    ob = len(b_val)
+
+    # Ensure the input vectors are column vectors
+    a_val = a_val.reshape(-1, 1)
+    b_val = b_val.reshape(-1, 1)
+    a_coe = a_coe.reshape(-1, 1)
+    b_coe = b_coe.reshape(-1, 1)
+
+    # Check if vectors a_val and b_val have only two elements
+    if oa != 2 or ob != 2:
```
vprusso, line 112:
```suggestion
    if len(a_val) != 2 or len(b_val) != 2:

```

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+
+    tot_dim = 2 ** (2 * m + 2)
+    obj_mat = np.zeros((tot_dim, tot_dim), dtype=float)
+
+    # Nested loops to compute the objective matrix
+    for a in range(2):  # a = 0 to 1
+        for b in range(2):  # b = 0 to 1
+            for x in range(1, m + 1):  # x = 1 to m (1-indexed in MATLAB, hence the range adjustment)
+                for y in range(1, m + 1):  # y = 1 to m
+                    b_coeff = (
```
vprusso, line 123: Is there a reason this is wrapped in parens and broken apart on multiple lines?

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+        The computed matrix.
+
+    """
+    # Create a permutation list as done in MATLAB
+    perm = list(range(m + 1))  # MATLAB indices start from 1
+    perm[0] = x
+    perm[x] = 0
+
+    # Calculate the matrix as in the MATLAB code
+    MN = a * np.eye(2 ** (m + 1)) + ((-1) ** a) * permutation_operator([2] * (m + 1), perm, 0, 1)
```
vprusso, line 32: Can just return this (instead of defining `MN` and returning that).

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+def MN_matrix(m: int, a: int, x: int) -> np.ndarray:
+    r"""Compute the matrices M_a^x and N_b^y.
+
+    Args:
+        m: The number of measurement settings for Alice and Bob.
+        a: The specific measurement setting for Alice.
+        x: The specific measurement setting for Bob.
+
+    Returns:
+        The computed matrix.
```
vprusso, line 23: We may wish to be a bit more specific about what the computed matrix here means.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+    oa = len(a_val)
+    ob = len(b_val)
+
+    # Ensure the input vectors are column vectors
+    a_val = a_val.reshape(-1, 1)
+    b_val = b_val.reshape(-1, 1)
+    a_coe = a_coe.reshape(-1, 1)
+    b_coe = b_coe.reshape(-1, 1)
+
+    # Check if vectors a_val and b_val have only two elements
```
vprusso, line 111: [nit]: Period at end of comment sentences

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,168 @@
...
+                    obj_mat += b_coeff * np.kron(MN_matrix(m, a, x), MN_matrix(m, b, y))
+
+    # Symmetrize the matrix to avoid numerical issues
+    obj_mat = (obj_mat + obj_mat.T) / 2
+    aux_mat = np.array([[1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]])
+
+    # Construct the SDP problem
+    W = cp.Variable((2 ** (2 * m), 2 ** (2 * m)), symmetric=True)
+
+    M = swap(W, [2, m + 1], [2] * (2 * m))
```
vprusso, line 141: Is there a way we can consolidate these lines instead of writing them as three separate statements? Maybe it's clearer as it is, but you can use your best judgment.

toqito/state_opt/bell_inequality_max.py
vprusso: You can likely look at any of the existing `test/` directories in the `toqito` project to see how unit tests are written as well (as this might be more instructive for writing tests in this case).

vprusso (changes requested):
Thanks for your contribution, @EthanFeld ! 

I added some comments and agree with the comment from @purva-thakre as well that we would want to add some unit tests to ensure that the proper upper bounds for known Bell inequalities (like the CHSH, I3322, etc.) are computed correctly. As well as tests to catch any erroneous input. 

Feel free to reach out if you have any questions, and thank you for your work thus far!

toqito/state_opt/bell_inequality_max.py
purva-thakre: @EthanFeld See if this helps you: https://third-bit.com/py-rse/testing.html

>  I wasn't sure if I should add it because its just a helper function

If you want this helper function to be available in the API then you should make sure everything about this is properly documented. Otherwise, use an `_` in front of the function name so that it will be a private function. 

https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

toqito/state_opt/bell_inequality_max.py
EthanFeld: I don't think there is. The only other way I've found to do it is a transpose and that would take more lines for reshaping.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,151 @@
...
+) -> float:
+    r"""Return the upper bound for the maximum violation(Tsirelson Bound) for a given bipartite Bell Inequality.
+
+    This computes the upper bound for the maximum value of a given bipartite Bell Inequality using an SDP.
+    The method is from :cite:`Navascues_2014_Characterization` and the implementation is based on :cite:`QETLAB_link`.
+    This is useful for various tasks in device independent quantum information processing.
+
+    The function formulates the problem as a SDP problem in the following format
+
+    .. math::
```
purva-thakre, line 28: add a blankline after you call mathmode. Right now, your equations are not rendered correctly because of this. 

https://toqito--1026.org.readthedocs.build/en/1026/autoapi/state_opt/bell_inequality_max/index.html

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,151 @@
...
+        W^{T_P} \geq 0, \quad \text{for all bipartitions } P.
+        \]
+        \]
+
+    Examples
+    =======
+
+
+    Consider the I3322 Bell inequality
+    .. math::
```
purva-thakre, line 45: same as my other comment about math mode

toqito/state_opt/bell_inequality_max.py
purva-thakre: that's weird. ruff shouldn't be doing this. let me double check.

toqito/state_opt/bell_inequality_max.py
vprusso: Gotcha, the way it is then is fine by me!

toqito/state_opt/bell_inequality_max.py
purva-thakre: I wonder if the issue is related to math mode having trouble with a multiline equation? And not the fact that there isn't a blankline between where you call mathmode and where the equation starts?

purva-thakre: Another multiline example that works:

https://github.com/vprusso/toqito/blob/5fff14a785fbe4a38f3e396d1898115e3b2a442a/toqito/state_metrics/fidelity_of_separability.py#L45-L61

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,147 @@
...
+        W^{T_P} \ge 0,\quad \text{for all bipartitions } P.
+        \end{multline}
+
+
+    Examples
+    =======
+
+
+    Consider the I3322 Bell inequality
+    .. math::
```
purva-thakre, line 42:
```suggestion
    Consider the I3322 Bell inequality
    
    .. math::
```

EthanFeld:
I fixed the issue with the math blocks.

vprusso:
> I fixed the issue with the math blocks.

Great. Is this ready for a re-review? If so, please do re-request myself and @purva-thakre using the "recycle symbols" next to our name so we are notified

Screenshot 2025-03-19 at 1 45 47 PM

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,108 @@
...
+    assert result >= 0
+
+
+# CHSH inequality test
+chsh_joint_coe = np.array([[1, 1], [1, -1]])
+chsh_a_coe = np.array([0, 0])
+chsh_b_coe = np.array([0, 0])
+chsh_a_val = np.array([1, -1])
+chsh_b_val = np.array([1, -1])
+chsh_expected = 2.828  # Tsirelson's bound
```
vprusso, line 95:
```suggestion
chsh_expected = np.sqrt(2) * 2  # Tsirelson's bound
```

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Since the exact optimal value is non-trivial to calculate by hand,
+    this test ensures the function returns a finite, non-negative float.
+    """
+    result = bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val)
+    assert isinstance(result, float)
+    assert result >= 0
+
+
+# CHSH inequality test
```
vprusso, line 89:
```suggestion
# CHSH inequality test.
```

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    with pytest.raises(ValueError):
+        bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val)
+
+
+# Degenerate case: all coefficients are zero (using minimal m = 1)
+degenerate_joint_coe = np.zeros((1, 1))
+degenerate_a_coe = np.zeros(1)
+degenerate_b_coe = np.zeros(1)
+
```
vprusso, line 53: You can likely just remove the variables here and define these in-line in the parameterization where needed.  
```suggestion
```

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,108 @@
...
+
+@pytest.mark.parametrize(
+    "joint_coe, a_coe, b_coe, a_val, b_val, expected",
+    [(degenerate_joint_coe, degenerate_a_coe, degenerate_b_coe, np.array([0, 1]), np.array([0, 1]), 0.0)],
```
vprusso, line 57:
```suggestion
    [(np.zeros((1,1)), np.zeros(1), np.zeros(1), np.array([0, 1]), np.array([0, 1]), 0.0)],
```

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,108 @@
...
+a_coe = np.array([0, -1, 0])
+b_coe = np.array([-1, -2, 0])
+a_val_valid = np.array([0, 1])
+b_val_valid = np.array([0, 1])
+expected_value = 0.250
+
+
+@pytest.mark.parametrize(
+    "joint_coe, a_coe, b_coe, a_val, b_val, expected",
+    [(joint_coe, a_coe, b_coe, a_val_valid, b_val_valid, expected_value)],
```
vprusso, line 20: I suggest using this test over known Bell inequalities. I did this for I3322, but you could also add the CHSH inequality, etc.

```suggestion
    # Bell I3322 inequality.
    [
        np.array([[1, 1, -1], [1, 1, 1], [-1, 1, 0]]),
        np.array([0, -1, 0]),
        np.array([-1, -2, 0]),
        np.array([0, 1]),
        np.array([0, 1]),
        0.250,
    ],
    # CHSH inequality
```

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,108 @@
...
+
+from toqito.state_opt import bell_inequality_max
+
+# Example inputs from the I3322 Bell inequality
+joint_coe = np.array([[1, 1, -1], [1, 1, 1], [-1, 1, 0]])
+a_coe = np.array([0, -1, 0])
+b_coe = np.array([-1, -2, 0])
+a_val_valid = np.array([0, 1])
+b_val_valid = np.array([0, 1])
+expected_value = 0.250
```
vprusso, line 15: (Assuming you adapt the suggestion below).
```suggestion
```

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,108 @@
+"""Tests for bell_inequality_max."""
+
+import numpy as np
+import pytest
+import scs
```
vprusso, line 5: There is no need for this import.

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,108 @@
...
+
+@pytest.mark.parametrize(
+    "joint_coe, a_coe, b_coe, a_val, b_val, expected",
+    [(joint_coe, a_coe, b_coe, a_val_valid, b_val_valid, expected_value)],
+)
+def test_bell_inequality_max_valid(joint_coe, a_coe, b_coe, a_val, b_val, expected):
+    """Test bell_inequality_max returns the expected value using valid input.
+
+    This test uses the I3322 Bell inequality.
+    """
```
vprusso, line 26:
```suggestion
    """Test bell_inequality_max returns the expected value using valid input."""
```

toqito/state_opt/tests/test_bell_inequality_max.py
vprusso, line 95: Suggestion is to remove this and put them in the parameterization above as suggested. 
```suggestion
```

purva-thakre:
@EthanFeld Looks like there is a fix spotted by ruff (quickly fixable) and a pytest failure.

https://github.com/vprusso/toqito/actions/runs/13970102762/job/39229900013?pr=1026#step:3:41 
https://github.com/vprusso/toqito/actions/runs/13970102762/job/39229899993?pr=1026#step:7:198

EthanFeld:
@purva-thakre Sorry about that, I was trying to commit the changes while I was away from my computer. I will fix it when I get back tomorrow.

EthanFeld:
@purva-thakre Thanks so much for your help and patience — I’ve just uploaded the fix!

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,113 @@
...
+        (
+            np.array([[1, 1, -1], [1, 1, 1], [-1, 1, 0]]),
+            np.array([0, -1, 0]),
+            np.array([-1, -2, 0]),
+            np.array([0, 1]),
+            np.array([0, 1]),
+            0.250,
+        )
+    ],
+    # CHSH inequality
```
vprusso, line 22: This comment here seems like it should include the CHSH test case, right?

EthanFeld: Sorry about that, not sure how that got in there.

toqito/state_opt/tests/test_bell_inequality_max.py
```diff
@@ -0,0 +1,112 @@
...
+    this test ensures the function returns a finite, non-negative float.
+    """
+    result = bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val)
+    assert isinstance(result, float)
+    assert result >= 0
+
+
+# CHSH inequality test.
+
+joint_coe = np.array([[1, 1], [1, -1]])
```
vprusso, line 86: I think you can likely combine the test in `test_bell_inequality_max_valid` with this one, right? You would just parameterize of the inequalities (i.e. I3322, CHSH, etc.). I'm not sure I see why these are two separate tests (e.g. they should just be one test with different parameterizations).

I think I may have made this comment before, but maybe it wasn't explained properly?

EthanFeld: Sorry about that, I just fixed it.

vprusso (approved):
LGTM! I'll wait for @purva-thakre to give this a look as well before merging.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,149 @@
...
+    >>> b_val = np.array([0, 1])
+    >>> '%.3f' % bell_inequality_max(joint_coe, a_coe, b_coe, a_val, b_val)
+    '0.250'
+
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
+    :raises ValueError: If a_val or b_val are not length 2.
```
purva-thakre, line 73:
```suggestion
    :raises ValueError: If `a_val` or `b_val` are not length 2.
```

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,149 @@
...
+        \text{s.t.} \quad \operatorname{tr}(W) = 1,\quad W \ge 0,\\[1ex]
+        W^{T_P} \ge 0,\quad \text{for all bipartitions } P.
+        \end{multline}
+
+
+    Examples
+    =======
+
+
+    Consider the I3322 Bell inequality
```
purva-thakre, line 41: Add a reference for this inequality. I don't think the one I have provided below is the original paper. 

https://arxiv.org/abs/quant-ph/0306129

You can get the bib entry by clicking on 

image

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,149 @@
...
+    b_val: np.ndarray,
+    solver_name: str = "SCS",
+) -> float:
+    r"""Return the upper bound for the maximum violation(Tsirelson Bound) for a given bipartite Bell inequality.
+
+    This computes the upper bound for the maximum value of a given bipartite Bell inequality using an SDP.
+    The method is from :cite:`Navascues_2014_Characterization` and the implementation is based on :cite:`QETLAB_link`.
+    This is useful for various tasks in device independent quantum information processing.
+
+    The function formulates the problem as a SDP problem in the following format
```
purva-thakre, line 26:
```suggestion
    The function formulates the problem as a SDP problem in the following format for the $W$-state.
```

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,149 @@
...
+    if len(a_val) != 2 or len(b_val) != 2:
+        raise ValueError("This script is only capable of handling Bell inequalities with two outcomes.")
+
+    tot_dim = 2 ** (2 * m + 2)
+    obj_mat = np.zeros((tot_dim, tot_dim), dtype=float)
+
+    # Nested loops to compute the objective matrix.
+    for a in range(2):  # a = 0 to 1
+        for b in range(2):  # b = 0 to 1
+            # Indicies below are adjusted to account for Python-MATLAB difference.
```
purva-thakre, line 101:
```suggestion
            # Indices below are adjusted to account for Python-MATLAB difference.
```

purva-thakre (approved):
LGTM. Great work @EthanFeld!!

I requested some minor changes.

toqito/state_opt/bell_inequality_max.py
```diff
@@ -0,0 +1,149 @@
...
+    b_val: np.ndarray,
+    solver_name: str = "SCS",
+) -> float:
+    r"""Return the upper bound for the maximum violation(Tsirelson Bound) for a given bipartite Bell inequality.
+
+    This computes the upper bound for the maximum value of a given bipartite Bell inequality using an SDP.
+    The method is from :cite:`Navascues_2014_Characterization` and the implementation is based on :cite:`QETLAB_link`.
+    This is useful for various tasks in device independent quantum information processing.
+
+    The function formulates the problem as a SDP problem in the following format for the $W$-state.
```
purva-thakre, line 26:
```suggestion
    The function formulates the problem as a SDP problem in the following format for the :math:`W`-state.
```

purva-thakre:
@EthanFeld Are you interested in GSoC? Could you please email us so that we can include you in all communications involving GSoC? Our emails are available in the wiki https://github.com/vprusso/toqito/wiki/GSoC-2025-Projects

PR #1032, opened 2025-03-17
fix #1031

### LINKED ISSUE #1031: `random_unitary` doesn't sample from the Haar distribution
## Describe the bug
The `random_unitary` function of the `rand` module doesn't produce a unitary/orthogonal matrix according to the Haar distribution. The culprit is probably these lines

https://github.com/vprusso/toqito/blob/3a67e7b6e2a151d78e931098e541ac53ca8a22e5/toqito/rand/random_unitary.py#L104-L105

The last line samples from the uniform distribution on $[0, 1]$ rather than the normal distribution. Replacing this line by
```python
gin = gen.standard_normal((dim[0], dim[1]))
```
should probably fix this.

A test should probably be added to see whether the distribution reproduces the first moments of the Haar measure. See below for an example.

## To Reproduce
The following code shows that the average density matrix of a state sampled using this method isn't close to the maximally mixed one, which should be the case if sampling from the Haar distribution.

```python
from toqito.rand import random_unitary
import numpy as np

res = np.zeros((2, 2), dtype="complex")
N = 100_000

for i in range(N):
    state = random_unitary(2)[:, 0].reshape(-1, 1)
    res += state @ state.T.conj()

# Should be close to the identity over 2, but off-diagonal elements aren't close to zero
print(res / N)
```
This effect is way more noticeable when passing the `is_real=True` argument, as seen using the following code.

```python
from toqito.rand import random_unitary
import numpy as np

res = np.zeros((2, 2))
N = 100_000

for i in range(N):
    state = random_unitary(2, is_real=True)[:, 0].reshape(-1, 1)
    res += state @ state.T

# Should be close to the identity over 2, but off-diagonal elements aren't close to zero
print(res / N)
```

Abhijeet213:
## Description
fix #1031

## Changes

  -  [x] Changed random to standard

## Checklist

### DIFF
--- a/toqito/rand/random_unitary.py
+++ b/toqito/rand/random_unitary.py
@@ -102,7 +102,7 @@ def random_unitary(
         raise ValueError("Unitary matrix must be square.")
 
     # Construct the Ginibre ensemble.
-    gin = gen.random((dim[0], dim[1]))
+    gin = gen.standard_normal((dim[0], dim[1]))
 
     if not is_real:
         gin = gin + 1j * gen.standard_normal((dim[0], dim[1]))


vprusso:
Hi @Abhijeet213 . As suggested by @tnemoz , we would want to add in some additional tests to ensure that the distribution being used here is correct.

vprusso:
@Abhijeet213 can you also alter the title of this issue to be more descriptive to the task that it is solving (as this is easier for tracking purposes).

vprusso:
Hi @Abhijeet213 ,

Just following up to see if you are still planning on completing this. If not, I can assign to another individual. Let me know in either case, and thank you!

vprusso:
Closing as @Abhijeet213 is unresponsive.

PR #1036, opened 2025-03-18
Cycle benchmarking feature

Wilson123d:
## Description
This pull request adds a Jupyter Notebook implementing Cycle Benchmarking.
- Includes an example of Cycle Benchmarking applied to quantum channels.
- Uses Toqito functions to analyze quantum states.
- Demonstrates benchmarking techniques for quantum systems.

Let me know if any modifications are needed!

### DIFF
--- a/cycle_benchmarking.py
+++ b/cycle_benchmarking.py
@@ -1,193 +1,96 @@
-{
- "cells": [
-  {
-   "cell_type": "code",
-   "execution_count": 1,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "import numpy as np\n",
-    "import matplotlib.pyplot as ply\n",
-    "import random\n",
-    "from numpy import linalg as LA\n",
-    "from scipy.optimize import curve_fit"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": 2,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "X = np.matrix('0 1; 1 0')\n",
-    "Z = np.matrix('1 0; 0 -1')\n",
-    "Y = np.matrix('0 -1j; 1j 0') \n",
-    "I = np.matrix('1 0; 0 1')\n",
-    "cnot=np.matrix('1 0 0 0; 0 1 0 0; 0 0 0 1; 0 0 1 0')\n",
-    "RZ=[[np.exp(1j*22.5),0],[0, np.exp(-1j*22.5)]]\n",
-    "RY=[[np.cos(45),-np.sin(45)], [np.sin(45), np.cos(45)]]\n",
-    "H=[[1/np.sqrt(2), 1/np.sqrt(2)], [1/np.sqrt(2), -1/np.sqrt(2)]]\n",
-    "a = np.matrix('1;0')\n",
-    "b=np.matrix('0;1')\n",
-    "c=np.matrix('1;0')\n",
-    "d = 1/np.sqrt(2)*np.matrix('1;1')\n",
-    "Z1=np.matrix('1 0; 0 1j')\n",
-    "I1=np.kron(I,I)\n",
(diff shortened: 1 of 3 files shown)


purva-thakre:
Hi @Wilson123d, I think there has been some miscommunication. When I suggested add a PR, I did not mean the jupyter notebook but a .`py` file.

vprusso:
Seconding @purva-thakre 's comment about using Python files. As another question, I wonder whether cycle benchmarking in toqito makes sense. Is there an associated issue for this?

purva-thakre:
>  As another question, I wonder whether cycle benchmarking in toqito makes sense. Is there an associated issue for this?

I believe there was some communication from you about adding this to the 'board'. Perhaps you meant this as an issue and not a PR?

Wilson123d:
> Hi @Wilson123d, I think there has been some miscommunication. When I suggested add a PR, I did not mean the jupyter notebook but a .`py` file.

Thanks for your message and suggestions, Purva. I have exported .ipynb to .py and added here. I am sorry that this is my first time to use poll request. For cycle benchmarking topic, if it is converted into toqito language, I may need to take a look of toqito as current codes are without toqito package. I could also added another two topics of "superposition of quantum channels" which can be used for calculating coherent and private information through superposition of two identical channels but I am not sure if it will be used very frequently. One more topic can be "optimization of rabi frequencies of motional modes", maybe two ions firstly, it will return the rabi frequency with error symmetry so that the phase trajectories are all 0 and the accumulated phase reaches pi/4(corresponds to maximally entangled state as initial state). Thanks!

vprusso:
Hi @Wilson123d 

It seems that you have changed the extension of the file from `.ipynb` to `.py`. What @purva-thakre is not just to change the file extension, but the actual content of the file itself. You may want to consult YouTube (or any other resource) for an introductory crash course on Python.

vprusso:
Additionally, you would want to undo your change in the `.gitignore` file (as we want to ignore notebooks) and to remove the `.ipynb` file (which still appears on your MR). @Wilson123d

cycle_benchmarking.py
```diff
@@ -0,0 +1,96 @@
...
+        elif (k[i]==X).all() == True:
+            k1.append(np.matmul(X,g(P2,p)))
+        elif (k[i]==Y).all() == True:
+            k1.append(np.matmul(Y,g(P2,p)))
+        else:
+            k1.append(np.matmul(Z,g(P2,p)))
+    return [k,k1]
+
+
+def fc(L,m,N,p1,p2):##for CNOT, L is the number of sequences, m is the number of cycles, p1 is the error value set for single Pauli gates inside the m cycle, p2 is an error set for CNOT gate, N is number of qubits
```
purva-thakre, line 58: Hi @Wilson123d, I am not too familiar with cycle benchmarking. Could you please explain how this code works? There are no function docstrings that could maybe help me understand it.

Also, can you describe how this will be helpful in toqito?

vprusso: Hi @Wilson123d . 

I appreciate the enthusiasm, but I'm having trouble seeing how what you're proposing here would be applicable to toqito. I would perhaps suggest looking at some of the issues on the board and contributing to one of those. I don't think we would be able to realistically review and incorporate this work in the PR.

Wilson123d: Thanks for your suggestions. Can I try the issues posted on boards which have been assigned to someone already as most issues have been assigned/discussed already? Thanks!

vprusso: The majority of issues are unassigned. If there are issues that have an assigned person who hasn't made progress,  you can feel free to supersede and take from them if no activity has happened in say, at least the last week.

vprusso:
@Wilson123d FYI, I'll be closing this MR, but feel free to pick up one of the issues on the board (or propose a new issue if you see something that can be improved or added!)

Issue #989, opened 2025-03-03
Add functions to generate quantum channels


purva-thakre:
Some channels are already defined in `toqito/channels`. We want to add a few more:

- [x] Amplitude damping channel
- [ ] Bit flip channel or Phase flip channel or Bit and Phase flip channel
- [x] Phase damping channel

RajdeepAher:
please assign

vprusso:
> please assign

Done!

purva-thakre:
@RajdeepAher Correct me if I am wrong, but you did not add the option of a phase flip option in #990.

Do you mind adding a new function or adding more options to your bitflip channel function? I'll leave it up to you on which one to choose. The only thing that is different between the two channels is bit flip is due to Pauli X while phase flip is due to Pauli Z. 

But please make sure to add a new unit test verifying you can create a bit and phase flip channel from the two.

RajdeepAher:
Yes, @purva-thakre, you're absolutely right—I overlooked the Phase Flip channel. I was considering working on this issue:  https://github.com/vprusso/toqito/issues/1060 and referring to the Pauli channel :![Image](https://github.com/user-attachments/assets/e133bc5f-4167-4ef5-b3f0-26f5300baa50) Since this is a more general case, I was thinking of implementing the Phase Flip channel by importing the Pauli channel and setting `p2` and `p3` to `0`.

purva-thakre:
> Since this is a more general case, I was thinking of implementing the Phase Flip channel by importing the Pauli channel and setting p2 and p3 to 0

This could work. 

A combined bit and phase flip channel can be then obtained by setting $p_2$ and $p_4$ to zero.

RajdeepAher:
Do I add Phase Flip channel now that Pauli Channel is implemented?

vprusso:
As @purva-thakre is the one who wrote this up, I'll ask if there is a reason for the "or" clause between "bit-flip" and "phase-flip" channels. Can having one of the two be considered to be completed or did you want to include bit-flip *and* phase-flip?

purva-thakre:
>  I'll ask if there is a reason for the "or" clause between "bit-flip" and "phase-flip" channels

When I wrote the issue, I thought both `or` and `and` options would be easy to implement

- bit flip $p_x, p_I \neq 0$
- phase flip $p_z, p_I \neq 0$
- bit and phase flip $p_y, p_I \neq 0$

purva-thakre:
[...]

PR #1037, opened 2025-03-18
feat: Add Generalized Amplitude Damping Channel

RajdeepAher:
## Description

Partially closes https://github.com/vprusso/toqito/issues/989

Implemented Generalized Amplitude Damping Channel which will act as a pre-cusor to implementing Amplitude Damping Channel (setting `p = 1`)
## Changes

  -  [x] Added `generalized_amplitude_damping.py` to simulate Generalized Amplitude Damping Channel.
  -  [x] Added corresponding coverage tests in `test_generalized_amplitude_damping.py`.
  -  [x] Changed `__init__.py` in `toqito/channels`.

## Checklist


toqito/channels/generalized_amplitude_damping.py
```diff
@@ -0,0 +1,81 @@
...
+    :param prob: The probability of energy loss, a float between 0 and 1.
+    :return: The evolved quantum state after applying the generalized amplitude damping channel.
+             If `input_mat` is `None`, returns the list of Kraus operators.
+
+    """
+    if not (0 <= prob <= 1):
+        raise ValueError("Probability must be between 0 and 1.")
+
+    if not (0 <= gamma <= 1):
+        raise ValueError("Probability must be between 0 and 1.")
```
vprusso, line 62: The value error message here needs to be changed for gamma.

toqito/channels/generalized_amplitude_damping.py
```diff
@@ -0,0 +1,81 @@
...
+
+    k0 = np.sqrt(prob) * np.array([[1, 0], [0, np.sqrt(1 - gamma)]])
+    k1 = np.sqrt(prob) * np.sqrt(gamma) * np.array([[0, 1], [0, 0]])
+    k2 = np.sqrt(1 - prob) * np.array([[np.sqrt(1 - gamma), 0], [0, 1]])
+    k3 = np.sqrt(1 - prob) * np.sqrt(gamma) * np.array([[0, 0], [1, 0]])
+
+    if input_mat is not None and input_mat.shape != (2, 2):
```
vprusso, line 69: The amplitude damping channel should not be restricted to just 2x2 cases though, it should be parameterized by a user-input dimension.

RajdeepAher: According to page 8 of [this paper](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.102.012401), the amplitude damping channel is typically defined as a qubit-to-qubit channel. I'm not sure how it would generalize to higher dimensions, but I may be short on domain knowledge. Apologies if my understanding is incomplete.

toqito/channels/generalized_amplitude_damping.py
```diff
@@ -0,0 +1,81 @@
...
+
+import numpy as np
+
+
+def generalized_amplitude_damping(
+    input_mat: np.ndarray | None = None,
+    gamma: float = 0,
+    prob: float = 0,
+) -> np.ndarray:
+    r"""Apply the generalized amplitude damping channel to a quantum state.
```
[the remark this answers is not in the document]
RajdeepAher: apply

toqito/channels/__init__.py
```diff
@@ -7,3 +7,4 @@
 from toqito.channels.partial_transpose import partial_transpose
 from toqito.channels.realignment import realignment
 from toqito.channels.reduction import reduction
+from toqito.channels.generalized_amplitude_damping import generalized_amplitude_damping
```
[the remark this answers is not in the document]
RajdeepAher: Thank you for the suggestion! To clarify, the "generalized" in this context specifically refers to the inclusion of the hyperparameter `p` (which controls the damping strength), not to a multi-qubit generalization. The standard amplitude damping channel typically assumes `p=1`, and our implementation extends this by allowing p to vary.

I have built my understanding from :
1. [quantumchannelzoo.org/person/generalized_amplitude](https://www.quantumchannelzoo.org/person/generalized_amplitude)

toqito/channels/generalized_amplitude_damping.py
vprusso: Hmm, no I think you're correct. I was under the assumption that the channel was generalized to higher qubits (and perhaps it is and can be), but most of the other definitions of and usages of the channel appear to be in the qubit case. 

Perhaps we can explicitly mention this somewhere in the docstring? What do you think?

RajdeepAher: Sure! I will include the updated docstrings in my next commit, along with all the other fixes you suggested.

Regarding the higher-dimensional channel you mentioned, it might be referring to this: [Multi-Level Amplitude Damping](https://www.quantumchannelzoo.org/person/multilvl_amplitude_damping). However, I’m not entirely sure.

Thank you for your time and understanding, @vprusso !

vprusso: Sounds good, I'll resolve this thread then. Thank you for looking into it!

toqito/channels/__init__.py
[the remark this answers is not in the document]
vprusso: Yes, I suppose this might be a detail to call out in the comments. By convention, the existing depolarizing channel, for instance, also takes a parameter `p`, but we call this "depolarizing" and not the "generalized depolarizing" in toqito. Perhaps we should do the same here and use just "amplitude_damping"?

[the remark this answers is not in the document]
RajdeepAher: I think setting the default value of `p` as `1` :
```python
def generalized_amplitude_damping(
    input_mat: np.ndarray | None = None,
    gamma: float = 0,
    prob: float = 1,
) -> np.ndarray:
```
and add the following in the docstring:
```python
"""To also include the standard implementation of amplitude damping, we have set `prob = 1` as the default implementation."""
```

would provide a suitable and equivalent implementation. However, I’m open to any suggestions for improvements.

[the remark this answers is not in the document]
vprusso: That sounds reasonable to me!

vprusso (approved):
This LGTM, but I'll await @purva-thakre 's review here as well

RajdeepAher:
@purva-thakre I have successfully resolved all the conflicts. Please review and let me know if everything looks good.

Once this feature is merged, I will address the conflicts in [PR #1042](https://github.com/vprusso/toqito/pull/1042) to ensure it can be merged smoothly as well.

Thank you!

docs/refs.bib
```diff
@@ -801,6 +801,21 @@ @misc{Philip_2023_Schrodinger
   primaryClass={quant-ph}
 }
 
+@article{PhysRevA.102.012401,
```
vprusso, line 804: This bibtex key is not formatted consistently as the other keys in this file.

toqito/channels/generalized_amplitude_damping.py
```diff
@@ -0,0 +1,83 @@
+"""Generates the (generalized) amplitude damping channel."""
+
+import numpy as np
+
+
+def generalized_amplitude_damping(
```
vprusso, line 6: I would prefer to call this `amplitude_damping` instead of `generalized_amplitude_damping`. This would apply to the file names as well.

toqito/channels/tests/test_generalized_amplitude_damping.py
```diff
@@ -0,0 +1,113 @@
...
+    "gamma, error_message",
+    [
+        (-0.1, "Gamma (damping rate) must be between 0 and 1."),
+        (1.1, "Gamma (damping rate) must be between 0 and 1."),
+    ],
+)
+def test_invalid_gamma(gamma, error_message):
+    """Test that invalid gamma values raise an error."""
+    with pytest.raises(ValueError, match=re.escape(error_message)):
+        generalized_amplitude_damping(prob=0.5, gamma=gamma)
```
vprusso, line 84: These can likely be combined into testing invalid inputs instead of having them being two separate functions.

toqito/channels/tests/test_generalized_amplitude_damping.py
```diff
@@ -0,0 +1,113 @@
...
+        # Complex mixed state.
+        (np.array([[0.7, 0.2 + 0.1j], [0.2 - 0.1j, 0.3]], dtype=complex), 0.4, 0.6),
+    ],
+)
+def test_apply_to_states(rho, prob, gamma):
+    """Apply the channel to various input states rho."""
+    kraus_ops = generalized_amplitude_damping(None, prob=prob, gamma=gamma)
+
+    expected_output = np.zeros((2, 2), dtype=complex)
+    for k in kraus_ops:
```
vprusso, line 51: Could these two tests also be combined where the first input argument is `None` for the Kraus operator case and where it is the density matrix in the other case?

toqito/channels/tests/test_generalized_amplitude_damping.py
```diff
@@ -0,0 +1,113 @@
...
+        generalized_amplitude_damping(prob=0.5, gamma=gamma)
+
+
+@pytest.mark.parametrize(
+    "rho",
+    [
+        np.eye(3),  # 3x3 matrix.
+        np.array([[1, 0, 0], [0, 1, 0]]),  # 2x3 matrix.
+        np.array([1, 0]),  # 1D array.
+        np.array([[1, 0, 0, 0], [0, 1, 0, 0]]),  # 2x4 matrix.
```
vprusso, line 93:
```suggestion
        # 3x3 matrix.
        np.eye(3),
        # 2x3 matrix.
        np.array([[1, 0, 0], [0, 1, 0]]),
        # 1D array.
        np.array([1, 0]),
        # 2x4 matrix.
        np.array([[1, 0, 0, 0], [0, 1, 0, 0]]),
```

vprusso:
Hi @RajdeepAher 

I noticed another implementation of the "generalized" and "non-generalized" amplitude damping channels

generalized: https://github.com/sumeetkhatri/QuTIpy/blob/7b0596823bd275a0dec3f6111738aa8de927116e/qutipy/channels.py#L1070
non-generalized: https://github.com/sumeetkhatri/QuTIpy/blob/7b0596823bd275a0dec3f6111738aa8de927116e/qutipy/channels.py#L1032

Perhaps we want to keep the name `amplitude_damping` but include a parameter (similar to `N` in the implementation mentioned above in the generalized case)? 

I'm thinking we would want to combine both of these under the name "amplitude_damping" but be specific about which one we are returning based on what is provided as input. Thoughts?

RajdeepAher:
I could be mistaken, but doesn't our implementation and the addition of docstrings, as mentioned here, already address this issue?  
[[Reference](https://github.com/vprusso/toqito/pull/1037#discussion_r2006374663)](https://github.com/vprusso/toqito/pull/1037#discussion_r2006374663)

vprusso:
> I could be mistaken, but doesn't our implementation and the addition of docstrings, as mentioned here, already address this issue? [[Reference](https://github.com/vprusso/toqito/pull/1037#discussion_r2006374663)]([#1037 (comment)](https://github.com/vprusso/toqito/pull/1037#discussion_r2006374663))

I *think* it does. In the case where the `N` they use is our `prob`, I suppose it would be the same, right? Good sanity check in any case though, right?

RajdeepAher:
Yes, I agree with you @vprusso :)

RajdeepAher:
@purva-thakre can you please review this?

Thank you!

toqito/channels/amplitude_damping.py
```diff
@@ -0,0 +1,83 @@
...
+    r"""Apply the generalized amplitude damping channel to a quantum state.
+
+    The generalized amplitude damping channel is a quantum channel that models energy dissipation
+    in a quantum system, where the system can lose energy to its environment with a certain
+    probability. This channel is defined by two parameters: `gamma` (the damping rate) and `prob`
+    (the probability of energy loss).
+
+    To also include standard implementation of amplitude damping, we have set `prob = 1` as the default implementation.
+
+    Note: This channel is defined for qubit systems in the standard literature :cite:`Khatri_2020_Information`.
```
purva-thakre, line 20:
```suggestion
    .. note::
          This channel is defined for qubit systems in the standard literature :cite:`Khatri_2020_Information`.

```

purva-thakre (approved):

Issue #1001, opened 2025-03-07
Keep dependency versions unpinned to specific versions unless needed


vprusso:
Generally, we shouldn't be hard-pinning to dependencies in the `pyproject.toml` file unless we absolutely need to for a specific dependency. Ideally, we want all of these packages to be up-to-date with the latest updates and should ensure we use the `>=` syntax where appropriate.

ryanbijoy:
Do u mean like this?

From:

```
[tool.poetry.dependencies]
python = ">=3.11,<4"
cvxpy = "1.6.2"
more-itertools = "10.6.0"
numpy = "2.2.3"
scipy = "1.15.2"
scs = "3.2.7.post2"
picos = "2.6.0"

```

To:
```
[tool.poetry.dependencies]
python = ">=3.11,<4"
cvxpy = ">=1.6.2"
more-itertools = ">=10.6.0"
numpy = ">=2.2.3"
scipy = ">=1.15.2"
scs = ">=3.2.7"
picos = ">=2.6.0"
```

vprusso:
This is a good question, @ryanbijoy . I actually think we should use `^=` in favor of `>=`. The reason for this is that using `>=` could unintentionally install a future breaking version which may not be compatible. 

Do you agree with this, @purva-thakre ? I know that you previously had some input on whether we should hard-pin dependencies.

ryanbijoy:
Sure working on this. Also should i keep the `.post2` 
which should it be `scs = "^=3.2.7.post2"` or `scs = "^=3.2.7"`

<img width="261" alt="Image" src="https://github.com/user-attachments/assets/2ca3fb40-ff36-473b-ae15-0dc4b13d270f" />

vprusso:
> scs

Hmm, yes, that `.post2` seems to be a specific way in which the `scs` developers decided to version things. Let's aim for `scs ^= 3.2.7.post2` as this is how they have the most recent released version on the PyPi server:
https://pypi.org/project/scs/

Issue #487, opened 2024-02-24
New conda environment with Python 3.12 leads to installation issues


purva-thakre:
If I try to use `poetry` in a new environment for `Python 3.12`, there are some installation issues where the project does not install at all.

```
conda create -n mytoqitoenv python=3.12
conda activate mytoqitoenv
pip install poetry
cd toqito
~/toqito$ poetry lock
~/toqito$ poetry install
```

Going down the rabbit hole of following what the error message says causes more dependency errors. 

```
  • Installing pyflakes (3.2.0)
  • Installing qdldl (0.1.7.post0): Failed

  ChefBuildError

  Backend subprocess exited when trying to invoke build_wheel
  
  running bdist_wheel
  running build
  running build_ext
  Traceback (most recent call last):
    File "<string>", line 81, in build_extensions
    File "local_path/anaconda3/envs/mytoqitoenv/lib/python3.12/subprocess.py", line 466, in check_output
      return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "local_path/anaconda3/envs/mytoqitoenv/lib/python3.12/subprocess.py", line 548, in run
      with Popen(*popenargs, **kwargs) as process:
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "local_path/anaconda3/envs/mytoqitoenv/lib/python3.12/subprocess.py", line 1026, in __init__
      self._execute_child(args, executable, preexec_fn, close_fds,
    File "local_path/anaconda3/envs/mytoqitoenv/lib/python3.12/subprocess.py", line 1950, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  FileNotFoundError: [Errno 2] No such file or directory: 'cmake'
  
  During handling of the above exception, another exception occurred:
  
  Traceback (most recent call last):
    File "local_path/anaconda3/envs/mytoqitoenv/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
      main()
    File "local_path/anaconda3/envs/mytoqitoenv/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py", line 335, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1039, opened 2025-03-19
To support >=python3.11

### LINKED ISSUE #1038: Support for Python 3.13
We should increase our support to Python 3.13 (and perhaps deprecate Python 3.10). 

This would entail:

1. Updating the `pyproject.toml` file to include all Python 3.13 versions
2. Updating any residual documentation, README, etc. to indicate that we additional support 3.13 (and not 3.10, if we deprecate this).
3. Update the automated tests to run over 3.13

jyotiraj-code:
## Description

Closes #1038
_considering deprecation of `python=3.10` without any warning_

This PR increases the support to `python>=3.11` and `<=3.13` 

Some comments:
- `pytest` installation error in the local env due to some packages not installed after `pip install -e .` from `setup.py` and had to manually install
image
- Required to change `PyPI` in `https://pypi.org/project/toqito/` documentation to showcase the dependency change `python>=3.11 to <=3.13`
- `pytest` result summary concludes with `1706 passed, 2 skipped, 44 warnings in 171.13s (0:02:51)`

## Changes

  -  [x] Updated `bug_report.md` to support `python>=3.11`
  -  [x] Updated `build-test-actions.yml` to support `python>=3.11`to `<=3.13` (Expanding support matrix)
  -  [x] Updated `docs.yml` to support `python>=3.11`to `<=3.13` (Expanding support matrix from`python=3.12`)
  -  [x] Updated `README.md` text of supporting machines with `python>=3.11`
  -  [x] Updated text in `contributing.rst`
  -  [x] Updated text in `getting_started.rst` 
  -  [x] Updated `pyproject.toml` to support `python>=3.11`
  -  [x] Updated `setup.py` for python requirement `python>=3.11`

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,8 @@ information.
 
 toqito is available via [PyPi](https://pypi.org/project/toqito/) for Linux, and macOS, with support for Python 3.10 to 3.12.
 
-```console
-(venv) $ pip install toqito
+```sh
+pip install toqito
 ```
 
 The following code gives an example on the usage:
@@ -64,7 +64,6 @@ chsh.classical_value()
 # 0.75
 chsh.quantum_value()
 # 0.8535533
-
 ```
 
 **Detailed documentation on all available methods, options, and input formats is available at [ReadTheDocs](https://toqito.readthedocs.io/en/latest/).**
@@ -105,7 +104,7 @@ A BibTeX entry that you can use to cite `toqito` is provided here:
 ```bib
 @misc{toqito,
    author       = {Vincent Russo},
-   title        = {toqito: A {P}ython toolkit for quantum information, version 1.0.0},
+   title        = {toqito: A {P}ython toolkit for quantum information},
    howpublished = {\url{https://github.com/vprusso/toqito}},
    month        = May,
    year         = 2021,
@@ -117,7 +116,7 @@ A BibTeX entry that you can use to cite `toqito` is provided here:
 
 The `toqito` project has been used or referenced in the following works:
 
-- [![a](https://img.shields.io/static/v1?label=thesis&message=31639397&color=inactive&style=flat-square)]([https://arxiv.org/abs/2406.13430](https://www.proquest.com/openview/eb0021dd3eb463b5fb12b7fc71d920eb/1?cbl=18750&diss=y&pq-origsite=gscholar)) Philip, Aby
+- [![a](https://img.shields.io/static/v1?label=thesis&message=31639397&color=inactive&style=flat-square)](https://www.proquest.com/openview/eb0021dd3eb463b5fb12b7fc71d920eb/1?cbl=18750&diss=y&pq-origsite=gscholar) Philip, Aby
 "On Multipartite Entanglement and Its Use", (2024).
 
 - [![a](https://img.shields.io/static/v1?label=arXiv&message=2406.13430&color=inactive&style=flat-square)](https://arxiv.org/abs/2406.13430) Bandyopadhyay, Somshubhro and Russo, Vincent
@@ -149,8 +148,7 @@ All contributions, bug reports, bug fixes, documentation improvements,
 enhancements, and ideas are welcome.
 
 A detailed overview of how to contribute can be found in the
-[contributing guide](https://toqito.readthedocs.io/en/latest/getting_started.html#contributing).
-
+[contributing guide](https://toqito.readthedocs.io/en/latest/contributing.html#contrib-guide-reference-label).
 
 ## License
 


purva-thakre:
Hi @jyotiraj-code, we prefer to not use `pip` as it tends to cause issues with resolving dependencies. 

[10E-4 10E-3 10E-2](https://toqito.readthedocs.io/en/latest/contributing.html#contributing-guide)

You might need to figure out the versions for the other dependencies in the `toml` file when you bump up the python version. 
https://github.com/vprusso/toqito/blob/46259d2a187afb5e9c85f01f0a371d2d83b51c9c/pyproject.toml

jyotiraj-code:
Hi @purva-thakre thanks, here is what I checked after your comment:
- Created a new `python=3.13` conda env 
- Installed `poetry` using pip
- Used `poetry install` to install the dependencies, no dependency issues were found for `python=3.13` version without bumping their versions
How should I proceed next?

purva-thakre:
> Is the PR falling checks because I didn't update the poetry.lock file?
probably

jyotiraj-code:
Hmm, you were correct with the update of poetry.lock file. I checked the logs for the failed tests, and it seems related to the problem of building PEP517 for `cvxopt`.

purva-thakre:
I have run into a similar issue in the past. https://github.com/vprusso/toqito/issues/487#issuecomment-1962409730

vprusso:
@purva-thakre @jyotiraj-code I think part of the problem might be that we are hard-pinning to the cvx dependencies in the `pyproject.toml` file. We pin to 1.6.2, but they are on 1.6.4 ([here](https://pypi.org/project/cvxpy/)). We likely want to stay up-to-date with whatever the latest releases of these (and other) packages are. 

There is another issue to adapt this for all of the dependencies in this issue: https://github.com/vprusso/toqito/issues/1001

For now, perhaps it's worth trying to pump these dependencies up in the `pyproject.toml` file? I'm not sure if this will resolve this issue, but it may be worth trying?

purva-thakre:
> @purva-thakre @jyotiraj-code I think part of the problem might be that we are hard-pinning to the cvx dependencies in the `pyproject.toml` file. We pin to 1.6.2, but they are on 1.6.4 ([here](https://pypi.org/project/cvxpy/)). We likely want to stay up-to-date with whatever the latest releases of these (and other) packages are.
> 
> There is another issue to adapt this for all of the dependencies in this issue: #1001
> 
> For now, perhaps it's worth trying to pump these dependencies up in the `pyproject.toml` file? I'm not sure if this will resolve this issue, but it may be worth trying?

But it's not clear to me why we are seeing this failure on macos only?

vprusso:
> But it's not clear to me why we are seeing this failure on macos only?

@purva-thakre @jyotiraj-code My only (likely wrong/bad) guess is that this is an issue that was worked out in newer versions of the module. That could be incorrect, but I think it's worthwhile to check and see if that update fixes the macos build error.

jyotiraj-code:
Hi @purva-thakre and @vprusso, thanks for your insights! I committed the `pyproject.toml` and updated all the `dependencies`, `group.dev.dependencies`, `group.lint.dependencies`, and `group.docs.dependencies` to include unpinned versions. 

I have a Linux distribution, and it is passing all the `pytests` with minor warnings.

purva-thakre:
Doesn't look like cvxopt is compatible with 3.13 on macos.

vprusso:
Hmm, that's unfortunate. Maybe we aren't ready to support 3.13 until cvxpy catches up then?

purva-thakre:
Last release of cvxopt was approximately 2 years ago. https://github.com/cvxopt/cvxopt

image

vprusso:
That's not very encouraging. However, it looks like they've updated the repo much more recently than that:
https://github.com/cvxopt/cvxopt

jyotiraj-code:
Yes, I was wondering that too. What if we can change our dependency from `cvxopt` to resolve this issue for the new Python version?

On another note, I checked `cvxopt`s github workflow and they support the `macos` build for `python=3.13`, what is with the issue then?

https://github.com/cvxopt/cvxopt/blob/0a3c0e5d12f047b3c7caf05724669a9eb8b31d29/.github/workflows/macos_build.yml#L33

vprusso:
@jyotiraj-code I think all the dependencies that you found could be easily resolved by altering `"cvxopt"` to `"scs"` as initially suggested. There doesn't appear to be any hard dependencies from what I can tell. 

The docs suggest they support 3.13, but the released version does not. As suggested, you can pin the dependency to the GitHub as opposed to the PyPi release. Does this make sense to you?

jyotiraj-code:
Thanks, @vprusso, sorry I missed that part of the `scs` solver and removing dependency on `cvxopt` in the previous comment.

I will create an issue for removing the dependency from `cvxopt` along with a PR solving it after some local env checkups.

vprusso:
> Thanks, @vprusso, sorry I missed that part of the `scs` solver and removing dependency on `cvxopt` in the previous comment.
> 
> I will create an issue for removing the dependency from `cvxopt` along with a PR solving it after some local env checkups.

Great, sounds like a plan!

jyotiraj-code:
Due to some local issues, I had to close this PR. The files are saved on a different local repo branch, which I will soon commit with the new PR head as soon as issue #1040 gets resolved

vprusso:
No worries, it happens! Thank you for the transparency, @jyotiraj-code !

PR #1043, opened 2025-03-21
Add Multiple Approaches to Approximate BCS Games and Discuss Group-Theoretic Criteria

Wilson123d:
This pull request provides two different ways to approximate the commuting operator value of a binary constraint system (BCS) game in toqito and links the results to the group-theoretic condition from [arXiv:1606.02278](https://arxiv.org/abs/1606.02278) about the existence of a central element J != e. Actually, J is presented for -1. 
Method A: from_bcs_game
We define a function bcs(M, b) that turns each row of  M including a factor or +1 / -1 arises from (-1)^b[i] (for ith row), then the implementation of built tool NonlocalGame.from_bcs_game from Toqito is applied directly. The theory behind this is we could suppose the solution group is generated by g, for each row of M, we could write summation over j of (Mijxj) = b[i] into product form of gi^Mji where we set gi = (-1)^(xi), this is from definition 3 in the paper. 

Method B: Manual Construction of NonlocalGame
We explicitly create the probability distribution over (Alice question, Bob question) pairs and build the predicate tensor that encodes the valid answer pairs. This approach offers more transparency into how the constraints are mapped to a nonlocal game.
Both methods yield an upper bound on the commuting measurement value for the BCS game at the specified NPA level.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 .idea/
 .coverage
 *.dat
-*.ipynb
 *.hdf5
 *.ipynb_checkpoints
 *.jupyter_cache
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.hdf5
 *.ipynb_checkpoints
 *.jupyter_cache
+*.ipynb
 _build
 __pycache__/
 docs/build/
--- a/BCS-.py
+++ b/BCS-.py
@@ -1,0 +1,56 @@
+import numpy as np
+from toqito.nonlocal_games.nonlocal_game import NonlocalGame
+
+def bcs(M, b):
+    """
+    Constructs a NonlocalGame object for the given binary constraint system.
+    """
+
+    m, n = M.shape
+    aq = list(range(m))  # List of Alice's possible questions
+    bq = {}#Bob's questions
+    for i in range(m):
+        L = []
+        for j in range(n):
+            if M[i, j] == 1:
+                L.append(j)
+        bq[i] = L
+
+    aa = {#Alice's possible answers with sublists of length 2**len(bq[s])
+        s: [[int(bit) for bit in format(i, f'0{len(bq[s])}b')]
+            for i in range(2 ** len(bq[s]))]
+        for s in range(m)
+    }
+    
+    ba = {s: [0, 1] for s in range(m)}#Bob's assignment always either 0 or 1 on one variable
+    prob = np.zeros((m, n), dtype=float)
+    for i in range(m):
+        for j in range(n):
+            if M[i, j] == 1:
+                prob[i, j] = 1
+    prob /= np.sum(prob) #Normalized uniformly distributed to assign
+
+    pred = np.zeros((2 ** max(len(bq[s]) for s in range(m)), 2, m, n), dtype=float)#Alice's output, Bob's output, Alice's input, Bob's input 
+
+    for i in range(m):
+        for j in bq[i]:
+            for k in range(len(aa[i])):
+                G = aa[i][k]
+                if sum(G) % 2 == b[i]:
+                    for l in range(2):
+                        if G[bq[i].index(j)] == l:
+                            pred[k, l, i, j] = 1.0
+
+    return NonlocalGame(prob, pred, reps=1)#Each round run once only
+
+def approximate(M, b, level):
+    game = bcs(M, b)
+    return game.commuting_measurement_value_upper_bound(k=level)
+
+
+M = np.array([[1, 1, 1], [0, 1, 1]])
+b = np.array([1, 0])#here is x11+x12+x13=b1 mod 2; x22+x23=b2 
+
+val = approximate(M, b, level=2)
+print("Commuting measurement value upper bound:", val)
+
--- a/BSC.py
+++ b/BSC.py
@@ -1,0 +1,26 @@
+import numpy as np
+from toqito.nonlocal_games.nonlocal_game import NonlocalGame
+
+def bcs(M, b):
+    m, n = M.shape
+    constraints = []
+    for i in range(m):
+        rhs = (-1) ** b[i]  
+        constraint = np.concatenate((M[i], np.array([rhs])))
+        constraints.append(constraint)
+    return constraints
+
+def approximate(M, b, level):
+    constraints = bcs(M, b)
+    game = NonlocalGame.from_bcs_game(constraints, reps=1)
+    
+    val = game.commuting_measurement_value_upper_bound(k=level)
+    return val
+
+M = np.array([[1, 1, 1],[0, 1, 1]])
+b = np.array([1, 0])
+    
+val = approximate(M, b, 3)
+print("Commuting measurement value upper bound:", val)
+
+
--- a/cycle_benchmarking.ipynb
+++ b/cycle_benchmarking.ipynb
@@ -1,0 +1,193 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import matplotlib.pyplot as ply\n",
+    "import random\n",
+    "from numpy import linalg as LA\n",
+    "from scipy.optimize import curve_fit"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "X = np.matrix('0 1; 1 0')\n",
+    "Z = np.matrix('1 0; 0 -1')\n",
+    "Y = np.matrix('0 -1j; 1j 0') \n",
+    "I = np.matrix('1 0; 0 1')\n",
+    "cnot=np.matrix('1 0 0 0; 0 1 0 0; 0 0 0 1; 0 0 1 0')\n",
+    "RZ=[[np.exp(1j*22.5),0],[0, np.exp(-1j*22.5)]]\n",
+    "RY=[[np.cos(45),-np.sin(45)], [np.sin(45), np.cos(45)]]\n",
+    "H=[[1/np.sqrt(2), 1/np.sqrt(2)], [1/np.sqrt(2), -1/np.sqrt(2)]]\n",
+    "a = np.matrix('1;0')\n",
+    "b=np.matrix('0;1')\n",
+    "c=np.matrix('1;0')\n",
+    "d = 1/np.sqrt(2)*np.matrix('1;1')\n",
+    "Z1=np.matrix('1 0; 0 1j')\n",
+    "I1=np.kron(I,I)\n",
+    "toffoli=np.matrix('1 0 0 0 0 0 0 0; 0 1 0 0 0 0 0 0; 0 0 1 0 0 0 0 0; 0 0 0 1 0 0 0 0; 0 0 0 0 1 0 0 0; 0 0 0 0 0 1 0 0; 0 0 0 0 0 0 0 1; 0 0 0 0 0 0 1 0')"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "P2=[I,X,Y,Z]\n",
+    "P3=[np.kron(I,I),np.kron(I,X),np.kron(I,Y),np.kron(I,Z),np.kron(X,I),np.kron(X,X),np.kron(X,Y),np.kron(X,Z),np.kron(Y,I),np.kron(Y,X),np.kron(Y,Y),np.kron(Y,Z),np.kron(Z,I),np.kron(Z,X),np.kron(Z,Y),np.kron(Z,Z)]\n",
+    "P4=[np.kron(I,np.kron(I,I)),np.kron(I,np.kron(I,X)),np.kron(I,np.kron(I,Y)),np.kron(I,np.kron(I,Z)),np.kron(I,np.kron(X,I)),np.kron(I,np.kron(X,X)),np.kron(I,np.kron(X,Y)),np.kron(I,np.kron(X,Z)),np.kron(I,np.kron(Y,I)),np.kron(I,np.kron(Y,X)),np.kron(I,np.kron(Y,Y)),np.kron(I,np.kron(Y,Z)),np.kron(I,np.kron(Z,I)),np.kron(I,np.kron(Z,X)),np.kron(I,np.kron(Z,Y)),np.kron(I,np.kron(Z,Z)),np.kron(X,np.kron(I,I)),np.kron(X,np.kron(I,X)),np.kron(X,np.kron(I,Y)),np.kron(X,np.kron(I,Z)),np.kron(X,np.kron(X,I)),np.kron(X,np.kron(X,X)),np.kron(X,np.kron(X,Y)),np.kron(X,np.kron(X,Z)),np.kron(X,np.kron(Y,I)),np.kron(X,np.kron(Y,X)),np.kron(X,np.kron(Y,Y)),np.kron(X,np.kron(Y,Z)),np.kron(X,np.kron(Z,I)),np.kron(X,np.kron(Z,X)),np.kron(X,np.kron(Z,Y)),np.kron(X,np.kron(Z,Z)),np.kron(Y,np.kron(I,I)),np.kron(Y,np.kron(I,X)),np.kron(Y,np.kron(I,Y)),np.kron(Y,np.kron(I,Z)),np.kron(Y,np.kron(X,I)),np.kron(Y,np.kron(X,X)),np.kron(Y,np.kron(X,Y)),np.kron(Y,np.kron(X,Z)),np.kron(Y,np.kron(Y,I)),np.kron(Y,np.kron(Y,X)),np.kron(Y,np.kron(Y,Y)),np.kron(Y,np.kron(Y,Z)),np.kron(Y,np.kron(Z,I)),np.kron(Y,np.kron(Z,X)),np.kron(Y,np.kron(Z,Y)),np.kron(Y,np.kron(Z,Z)),np.kron(Z,np.kron(I,I)),np.kron(Z,np.kron(I,X)),np.kron(Z,np.kron(I,Y)),np.kron(Z,np.kron(I,Z)),np.kron(Z,np.kron(X,I)),np.kron(Z,np.kron(X,X)),np.kron(Z,np.kron(X,Y)),np.kron(Z,np.kron(X,Z)),np.kron(Z,np.kron(Y,I)),np.kron(Z,np.kron(Y,X)),np.kron(Z,np.kron(Y,Y)),np.kron(Z,np.kron(Y,Z)),np.kron(Z,np.kron(Z,I)),np.kron(Z,np.kron(Z,X)),np.kron(Z,np.kron(Z,Y)),np.kron(Z,np.kron(Z,Z))]\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def g(l,p):\n",
+    "    return random.choices(l,weights=((1-p)*1000,1000*p/3,1000*p/3,1000*p/3),k=1)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def g1(l,p):\n",
+    "    return random.choices(l,weights=((1-p)*1000,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15,1000*p/15),k=1)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def g2(l,p):\n",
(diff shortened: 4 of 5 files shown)


Wilson123d:
> @Wilson123d it seems you added files unintentionally from the cycle benchmarking issue. We want to remove those from this diff. It's also not clear to me how or where in your code that you are checking for whether the strategy is perfect or not. It seems like you are just computing the commuting measurement operator value of a BCS (which we already have the functionality to do).

Thanks for your message. I have removed cycle_benchmarking.ipynb. what I thought is I encode the binary constraint system Mx=b into a nonlocal game by setting up the inputs (questions) for Alice and Bob, and the corresponding winning conditions (constraints). These winning conditions reflect the equations from Mx=b. Then obtained an upper bound on the maximum winning probability that quantum players can achieve when their measurement operators commute. If the computed value is 1, it means that there is a strategy such that the players can win the game with probability 1. According to Theorem 4 in the paper, a winning probability of 1 is equivalent to the existence of a perfect commuting‑operator quantum strategy (which in turn means there exists an operator solution for Mx=b and the associated solution group has a nontrivial element 𝐽.

vprusso:
@Wilson123d you still have `cycle_benchmarking.py` which you need to remove. 

I would imagine that the resolution of this feature would include a function `has_perfect_commuting_measurement_strategy` (or similar) that is a property acting on a BCS game. This doesn't appear to be what you've provided, and what you're committing instead seem to be one-off scripts that don't appear to be integrated within toqito, have tests, conform to the style, etc. 

I encourage you to refer to the contribution guide for more information on how to contribute to the toqito project:
https://toqito.readthedocs.io/en/latest/contributing.html#contrib-guide-reference-label

BcS.py
```diff
@@ -0,0 +1,152 @@
...
+            m_B = total_constraints - m_A
+            aq = list(range(m_A))
+            bq = list(range(m_B))
+            prob = np.ones((m_A, m_B), dtype=float) / (m_A * m_B)
+            pred = np.ones((2 ** (total_constraints // 2), 2 ** m_B, m_A, m_B), dtype=float)#Here, set up to be 1 for simplicity
+            game = cls(prob, pred, reps=1)
+            game._perfect = True
+            return game
+
+# Unit Tests
```
purva-thakre, line 108: @Wilson123d As @vprusso suggested, it might be worth reading through the contributing guide in the documentation. 

Unit tests are added in a separate file. Please take a look at how `toqito` is organized.

test_BCS.py
```diff
@@ -0,0 +1,52 @@
+print("Starting tests in test_BCS.py...")
+
+import unittest
```
purva-thakre, line 3: @Wilson123d We use pytest not unittest

BCS.py
```diff
@@ -0,0 +1,104 @@
...
+            if A[aq[r], col] == 1:
+                A[aq[r]] = (A[aq[r]] + A[aq[row]]) % 2
+                b1[aq[r]] = (b1[aq[r]] + b1[aq[row]]) % 2
+        row += 1
+    for r in range(m):
+        if np.all(A[aq[r]] == 0) and b1[aq[r]] == 1:
+            return False
+    return True
+
+class NonlocalGame:
```
purva-thakre, line 39: There already exists a class with this name in toqito.

vprusso, line 39: This is defined in the `nonlocal_games/` module. Is there are reason you are defining it in this file?

BCS.py
```diff
@@ -0,0 +1,104 @@
...
+        self.reps = reps
+
+    def commuting_measurement_value_upper_bound(self, k) -> float:
+        if hasattr(self, '_perfect'):
+            return 1.0 if self._perfect else 0
+        return 0
+
+    def has_perfect_commuting_measurement_strategy(self, tol: float = 1e-6) -> bool:
+        comm_value = self.commuting_measurement_value_upper_bound(k=2)
+        return np.isclose(comm_value, 1.0, atol=tol)
```
vprusso, line 52: This doesn't seem to be doing what we want. We can already compute the commuting measurement value numerically. We want a method to let us know if a perfect strategy exists or not. In doing so, we specifically do not want to invoke the upper bound function, we instead want to use the group theoretic approach that is outlined in the issue.

test_BCS.py
```diff
@@ -0,0 +1,52 @@
+print("Starting tests in test_BCS.py...")
```
vprusso, line 1: Please refer to the contribution guide for how to write and where to put tests.

toqito/nonlocal_games/bcs.py
```diff
@@ -0,0 +1,119 @@
...
+                G.add_edge(k, l)
+    # Check for cycle in the induced subgraph
+    has_cycle = False
+    if len(nodes) > 0:
+        if len(nx.cycle_basis(G)) > 0:
+            has_cycle = True
+    # This is to check if the nontrivial J exists, consistent with lemma 9.
+    return has_cycle
+
+class BCSNonlocalGame(NonlocalGame):
```
vprusso, line 87: I wonder if defining this class (and also the `from_bcs_game` function is necessary. We already have such a method that is defined for any `NonlocalGame` object. 

Presumably, you could have something that looks like:

```py
from toqito.nonlocal_games import NonlocalGame

bcs_constraints = ...
game = NonlocalGame.from_bcs_game(bcs_constraints)

# Your function which would return either `True` or `False` depending on 
# whether it has a perfect strategy or not.
game.has_perfect_commuting_measurement_strategy()
```

Does the above make sense? It seems redundant to me to be redefining things that already exist in the `nonlocal_game.py` module.

Wilson123d: Thank you for your feedback! I have removed the BCSNonlocalGame class, as it is indeed unnecessary. To align with the existing NonlocalGame class from Toqito, I introduced functions to convert the input (M,b) into the required constraints format.

However, to fully integrate the functionality, we may need to modify NonlocalGame in Toqito by adding the _perfect attribute and defining a has_perfect_commuting_measurement_strategy() method that simply returns self._perfect. Additionally, within from_bcs_game, we might insert perfect = check_perfect_commuting_strategy(M, b) and update game._perfect accordingly—similar to what was previously handled in BCSNonlocalGame.

The updated tests all pass using check_perfect_commuting_strategy(M: np.ndarray, b: np.ndarray) -> bool. However, when using NonlocalGame directly from Toqito, an error occurs due to the missing _perfect attribute and has_perfect_commuting_measurement_strategy() method.

Lastly, I ensured that the constraints are stored in the solution group formalism for consistency.

vprusso: Thanks for the update @Wilson123d 

You may also want to take a peek at how to format comments on GitHub using markdown as they can sometimes be hard to follow:
https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax

toqito/nonlocal_games/test_bcs.py
```diff
@@ -0,0 +1,51 @@
...
+    #game = NonlocalGame.from_bcs_game(constraints, reps=1)
+    # Here, our game should have the perfect strategy flag set if no contradiction arises.
+    # For consistency, we check:
+    #assert game.has_perfect_commuting_measurement_strategy() is True
+
+def test_chsh_bcs():
+    M = np.array([[1, 1], [1, 1]], dtype=int)
+    b = np.array([0, 1], dtype=int)
+    constraints = bcs(M, b)
+    assert check_perfect_commuting_strategy(M, b) is False
```
purva-thakre, line 21:
```suggestion
    assert not check_perfect_commuting_strategy(M, b)
```
You don't really need `is True` or `is False` at the end.

toqito/nonlocal_games/test_bcs.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
```
purva-thakre, line 1: Check where this file is placed compared to other tests files in `nonlocal_games`. 

Please read the contributing guide. https://toqito.readthedocs.io/en/latest/contributing.html

You need to import your main function into the init file too.

toqito/nonlocal_games/test_bcs.py
Wilson123d: Thanks for your suggestion. I have modified it.

toqito/nonlocal_games/test_bcs.py
Wilson123d: Thanks for your suggestion. I have moved test_binary_constraint_games.py to the tests files.

toqito/nonlocal_games/bcs.py
```diff
@@ -0,0 +1,126 @@
...
+        h = 0
+        for j in range(n):
+            if M[i, j] == 1:
+                h |= (1 << j)
+        row.append(h)
+    # Parity bits list
+    parity=[]
+    for i in b:
+        parity.append(int(i))
+    return row, parity
```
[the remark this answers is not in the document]
Wilson123d: Thanks for your suggestion. Your code is completely correct. I have simplified it.

toqito/nonlocal_games/bcs.py
```diff
@@ -0,0 +1,126 @@
+import numpy as np
```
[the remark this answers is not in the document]
Wilson123d: Thanks for your suggestion. I have renamed both files.

toqito/nonlocal_games/bcs.py
```diff
@@ -0,0 +1,126 @@
...
+            if M[i, j] == 1:
+                h |= (1 << j)
+        row.append(h)
+    # Parity bits list
+    parity=[]
+    for i in b:
+        parity.append(int(i))
+    return row, parity
+
+def check_perfect_commuting_strategy(M: np.ndarray, b: np.ndarray) -> bool:
```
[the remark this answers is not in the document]
Wilson123d: Thanks for your suggestion. I  tried to simplified it.

toqito/nonlocal_games/test_bcs.py
```diff
@@ -0,0 +1,51 @@
...
+from toqito.nonlocal_games.nonlocal_game import NonlocalGame
+from BCS import bcs, check_perfect_commuting_strategy, nonlocal_game_from_constraints
```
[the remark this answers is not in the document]
Wilson123d: Thanks for your suggestion, yes, the error should be the name of file BCS. I used capital letter but github saved it as bcs, it should be fixed now.

toqito/nonlocal_games/test_bcs.py
```diff
@@ -0,0 +1,51 @@
...
+    M = np.array([
+        [1, 1, 1],
+        [1, 1, 0],
+        [0, 1, 1]
+    ], dtype=int)
+    b = np.array([1, 0, 0], dtype=int)
+    constraints = bcs(M, b)
+    assert check_perfect_commuting_strategy(M, b) is True
+    #game = nonlocal_game_from_constraints(constraints)
+    #assert game.has_perfect_commuting_measurement_strategy() is True
```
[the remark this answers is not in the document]
Wilson123d: > Yes, that line was commented out intentionally. The reason is that calling `has_perfect_commuting_measurement_strategy()` would raise an error because this method is not defined in the `NonlocalGame` class in Toqito.  
>  
> Previously, I created a subclass called `BCSNonlocalGame` specifically to run these tests. That subclass included both the `has_perfect_commuting_measurement_strategy()` method and the `_perfect` attribute to support this functionality.

toqito/nonlocal_games/binary_constraint_system_game.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def generate_solution_group(M: np.ndarray, b: np.ndarray):
+    """
+    Constructs the solution group structure for the BCS game in a more vectorized manner.
+    
+    Each row in M is converted into a bitmask by summing powers of 2 for columns
+    where M[i, j] == 1. The parity bits are taken directly from b.
+    
+    Returns:
```
purva-thakre, line 26: We have a different style for specifying params. Please take a look at the contributing guide in the documentation. 
https://toqito.readthedocs.io/en/latest/contributing.html#contrib-guide-reference-label

toqito/nonlocal_games/binary_constraint_system_game.py
```diff
@@ -0,0 +1,127 @@
+import numpy as np
+import networkx as nx
+from toqito.nonlocal_games.nonlocal_game import NonlocalGame
+
+def bcs(M, b):
```
vprusso, line 5: I think this function should be renamed to something other than `bcs`. Is there not some way to rectify the way in which a BCS game is constructed from the constraint list within the `nonlocal_game.py` file and this function?

toqito/nonlocal_games/binary_constraint_system_game.py
```diff
@@ -0,0 +1,127 @@
...
+def generate_solution_group(M: np.ndarray, b: np.ndarray):
+    """
+    Constructs the solution group structure for the BCS game in a more vectorized manner.
+    
+    Each row in M is converted into a bitmask by summing powers of 2 for columns
+    where M[i, j] == 1. The parity bits are taken directly from b.
+    
+    Returns:
+        row_masks (List[int]): Each integer is a bitmask representing one row of M.
+        parity (List[int]): Each element is 0 or 1, taken from b.
```
vprusso, line 28: You should be specifying the return types via type hints for the function (not in the doc-string).

toqito/nonlocal_games/binary_constraint_system_game.py
```diff
@@ -0,0 +1,127 @@
...
+        for i, u in enumerate(nodes)
+        for v in nodes[i+1:]
+        if row[u] & row[v]
+    ]
+    G.add_edges_from(edges)
+
+    # If there's a cycle, a perfect strategy exists
+    return bool(nx.cycle_basis(G))
+
+def nonlocal_game_from_constraints(constraints) -> NonlocalGame:
```
vprusso, line 101: You need to type-hint the arguments for your functions.

toqito/nonlocal_games/tests/test_binary_constraint_system_game.py
```diff
@@ -0,0 +1,53 @@
...
+from binary_constraint_system_game import bcs,check_perfect_commuting_strategy,nonlocal_game_from_constraints
+
+def test_classically_satisfiable_bcs():
+    M = np.array([[1, 0], [0, 1]], dtype=int)
+    b = np.array([0, 0], dtype=int)
+    constraints = bcs(M, b)
+    # Expecting a perfect strategy => True
+    assert check_perfect_commuting_strategy(M, b)
+    game = NonlocalGame.from_bcs_game(constraints, reps=1)
+    #assert game.has_perfect_commuting_measurement_strategy()
```
vprusso, line 12: Did you resolve the comments from before? I believe I previously asked about why these are commented out.

Wilson123d: > **Note:** I created a new file called `modified_nonlocal_game.py`, which is a duplicate of `nonlocal_game.py`.
> 
> I only changed the `from_bcs_game` method inside the `NonlocalGame` class for testing.
> 
> Everything should now be consistent. If this change looks good, I’ll delete `modified_nonlocal_game.py` and apply the final changes directly in `nonlocal_game.py`.
>
>  **All tests are now passing.**

toqito/nonlocal_games/binary_constraint_system_game.py
```diff
@@ -0,0 +1,184 @@
...
+    3. Checks for a contradiction of the form ``0 = 1``.
+    4. Constructs a graph of involved constraints and checks for cycles.
+       A cycle implies a nontrivial group element, which corresponds
+       to the existence of a perfect strategy.
+
+    Examples
+    --------
+    .. code-block:: python
+
+        import numpy as np
```
vprusso, line 117: Please refer to other docstrings in toqito to see how we represent examples.

toqito/nonlocal_games/binary_constraint_system_game.py
```diff
@@ -0,0 +1,184 @@
...
+    1. Converts ``M`` and ``b`` into bitmask representation via
+       :func:`generate_solution_group`.
+    2. Performs Gaussian elimination over :math:`\mathrm{GF}(2)`.
+    3. Checks for a contradiction of the form ``0 = 1``.
+    4. Constructs a graph of involved constraints and checks for cycles.
+       A cycle implies a nontrivial group element, which corresponds
+       to the existence of a perfect strategy.
+
+    Examples
+    --------
```
vprusso, line 114: Please refer to the toqito docs to see how represent the subject headings.

toqito/nonlocal_games/binary_constraint_system_game.py
```diff
@@ -0,0 +1,184 @@
...
+    powers = 1 << np.arange(M.shape[1])  # e.g. if n=3 => [1, 2, 4]
+    
+    return (M * powers).sum(axis=1).astype(int).tolist(), b.astype(int).tolist()
+
+def check_perfect_commuting_strategy(M: np.ndarray, b: np.ndarray) -> bool:
+    r"""
+    Determine whether a perfect commuting-operator strategy exists for a BCS game.
+
+    This function checks if there is a perfect commuting-operator strategy
+    for the system of binary constraints ``M x = b``. The logic is based on
```
vprusso, line 99: The phrase M x = b should not be represented by "code", it is a "math"-ematical statement (and should use the appropriate rendering environment.

toqito/nonlocal_games/binary_constraint_system_game.py
```diff
@@ -0,0 +1,184 @@
...
+
+    :param M: A binary matrix of shape ``(m, n)`` specifying which variables appear
+              in each of the ``m`` constraints. Each entry is either 0 or 1.
+    :param b: A binary vector of length ``m``. For each row ``i``, the exponent
+              :math:`b[i]` determines whether the right-hand side is +1 or -1,
+              via :math:`(-1)^{b[i]}`.
+    :return: A list of 1D NumPy arrays, where each array is of length ``n+1``.
+             The first ``n`` entries correspond to a row of ``M``, and the last
+             entry is :math:`\pm 1` depending on ``b[i]``.
+    :rtype: List[np.ndarray]
```
vprusso, line 44: Please do not put in explicit return types. We know what these should be based on the type-hinting.

vprusso (commented):
Thanks for the updates, @Wilson123d 

I'm not sure I follow the `modified_nonlocal_game.py` approach, but if you think that helps you in scaffolding things, that's okay as a temporary thing.

purva-thakre:
@Wilson123d What's the status on this?

Wilson123d:
> @Wilson123d What's the status on this?

> Thanks for your message!

Last time, I modified some design recipes for the function and updated the `from_bcs_game` method inside the `NonlocalGame` class for testing. All tests were passing at that point.

I also created a new file called `modified_nonlocal_game.py`, which is currently a duplicate of `nonlocal_game.py`.

If we're going to modify `nonlocal_game.py` directly, then I can delete `modified_nonlocal_game.py`.

vprusso:
Hi @Wilson123d ,

Yes, ideally we would be modifying the existing file to include this logic instead of having some auxiliary file. If you could proceed with the MR, we can continue to offer our feedback if helpful. Thank you, @Wilson123d

vprusso:
Hi @Wilson123d 

Are you still intending on working on this MR? It seems like there are still some outstanding comments, conflicts, etc. Thank you!

Wilson123d:
> Hi @Wilson123d
> 
> Are you still intending on working on this MR? It seems like there are still some outstanding comments, conflicts, etc. Thank you!

Hi @vprusso, thanks for the reminder!

> I originally thought the MR was working fine apart from the `RuntimeWarning` related to `dependent_variables`.

> After digging deeper, I realized that some constraints could result in uniform tensors (all ones), which caused `np.diff(..., axis=i).any()` to return `False` across all axes — leading to a division-by-zero in `p_y = dependent_variables[j] / num_dependent_vars`.

> I’ve now fixed this by initializing each constraint tensor with:
> ```python
> constraint = np.full((2,) * n, fill_value=-rhs, dtype=int)
> constraint[idx] = rhs
> ```
> This guarantees variation in the tensor and ensures `dependent_variables` are detected correctly.

All tests now pass without warnings (`test_magic_square_bcs`, `test_special_case`, etc.), and I’ll push the updated changes and resolve the remaining comments shortly.

Thanks again for your patience!

toqito/nonlocal_games/nonlocal_game.py
```diff
@@ -81,6 +84,29 @@ def from_bcs_game(cls, constraints: list[np.ndarray], reps: int = 1) -> "Nonloca
         :param reps: Number of parallel repetitions to perform. Default is 1.
         :return: A NonlocalGame object arising from the variables and constraints that define the game.
         """
+        # Detect if constraints are in 1D form (each a vector of length n+1).
```
vprusso, line 87: I don't think the placement of this makes sense. In the event that the user wants to invoke `from_bcs`, we don't want to simultaneously also do a check to see if the strategy is perfect. Instead, I would anticipate a separate class method in the `NonlocalGame` class called `is_bcs_perfect_commuting_strategy`. This function can then be conditionally invoked on the `NonlocalGame` object to return a boolean. 

Currently, there is too much conflation between the two items and this makes the computation for conversion longer as well as includes an interface for BCS that goes against the interface for all other NLG objects. Does this make sense to you?

Wilson123d: >>> @vprusso Thanks for pointing this out. I have revised the design so that `from_bcs_game` no longer attempts to detect whether a perfect commuting‑operator strategy exists during the game’s construction. 

I introduced a separate instance method on `NonlocalGame`:

```python
def is_bcs_perfect_commuting_strategy(self) -> bool:
    ...
This method can be invoked on the constructed 'NonlocalGame' object at any time to check for a perfect commuting‑operator strategy. It handles the conversion of constraints (if necessary) into the form required by 'check_perfect_commuting_strategy', but it no longer conflates game building with strategy checking.

Comment (Wilson123d) [reply toqito/nonlocal_games/nonlocal_game.py]: > @vprusso Thanks for pointing this out. I have revised the design so that `from_bcs_game` no longer attempts to detect whether a perfect commuting‑operator strategy exists during the game’s construction. 

I introduced a separate instance method on `NonlocalGame`:

```python
def is_bcs_perfect_commuting_strategy(self) -> bool:
    ...
```
This method can be invoked on the constructed `NonlocalGame` object at any time to check for a perfect commuting‑operator strategy. It handles the conversion of constraints (if necessary) into the form required by `check_perfect_commuting_strategy`, but it no longer conflates game building with strategy checking.

purva-thakre:
@Wilson123d Is this PR ready for a review?

Wilson123d:
Thanks for your message. Yes I think so. I included it in the submission of proposal. Thanks!

Best,
Yangchen Zhou
________________________________
发件人: Purva Thakre ***@***.***>
发送时间: 2025年5月7日 3:59
收件人: vprusso/toqito ***@***.***>
抄送: Yangchen Zhou ***@***.***>; Mention ***@***.***>
主题: Re: [vprusso/toqito] Add Group-Theoretic Verification for Perfect Commuting Strategies in BCS Games (PR #1043)

[https://avatars.githubusercontent.com/u/66048318?s=20&v=4]purva-thakre left a comment (vprusso/toqito#1043)<https://github.com/vprusso/toqito/pull/1043#issuecomment-2855787041>

@Wilson123d<https://github.com/Wilson123d> Is this PR ready for a review?

―

PR #1044, opened 2025-03-21
fix: `random_unitary` doesn't sample from the Haar distribution

### LINKED ISSUE #1031: `random_unitary` doesn't sample from the Haar distribution

jyotiraj-code:
## Description

Fixes #1031 

This bug was noticed, and a probable solution was put forward by @tnemoz; thanks!

The bug was related to the `rand` module not producing the unitary/orthogonal matrix according to the Harr distribution.

## Changes

  -  [x] Fixed the bug by changing `random distribution` with a 'standard distribution` causing the issue in `rand/random_unitary.py`
  -  [x] Updated the `test` files  `test_random_density_matrix.py`, `test_random_orthonormal_basis.py` and `test_random_unitary.py`

## Checklist

### DIFF
--- a/toqito/rand/random_density_matrix.py
+++ b/toqito/rand/random_density_matrix.py
@@ -34,8 +34,8 @@ def random_density_matrix(
     >>> from toqito.rand import random_density_matrix
     >>> complex_dm = random_density_matrix(2)
     >>> complex_dm # doctest: +SKIP
-    array([[0.83378447+0.j        , 0.25937396-0.12958422j],
-           [0.25937396+0.12958422j, 0.16621553+0.j        ]])
+    array([[ 0.53822849+0.j        , -0.26155866+0.02081311j],
+           [-0.26155866-0.02081311j,  0.46177151+0.j        ]])
 
     We can verify that this is in fact a valid density matrix using the :code:`is_denisty` function from :code:`toqito`
     as follows
@@ -49,8 +49,8 @@ def random_density_matrix(
     >>> from toqito.rand import random_density_matrix
     >>> real_dm = random_density_matrix(2, is_real=True)
     >>> real_dm # doctest: +SKIP
-    array([[0.3686455 , 0.48243625],
-           [0.48243625, 0.6313545 ]])
+    array([[0.47783773, 0.45763467],
+           [0.45763467, 0.52216227]])
 
 
     Again, verifying that this is a valid density matrix can be done as follows.
@@ -65,8 +65,8 @@ def random_density_matrix(
     >>> from toqito.rand import random_density_matrix
     >>> bures_mat = random_density_matrix(2, distance_metric="bures")
     >>> bures_mat # doctest: +SKIP
-    array([[0.73969055+0.j        , 0.09494426-0.15279342j],
-           [0.09494426+0.15279342j, 0.26030945+0.j        ]])
+    array([[ 0.41427711+0.j       , -0.15503543+0.2405496j],
+           [-0.15503543-0.2405496j,  0.58572289+0.j       ]])
 
 
     As before, we can verify that this matrix generated is a valid density matrix.
--- a/toqito/rand/random_orthonormal_basis.py
+++ b/toqito/rand/random_orthonormal_basis.py
@@ -16,16 +16,16 @@ def random_orthonormal_basis(dim: int, is_real: bool = False, seed: int | None =
 
     >>> from toqito.rand import random_orthonormal_basis
     >>> random_orthonormal_basis(4, is_real = True) # doctest: +SKIP
-    [array([0.52188745, 0.4983613 , 0.69049811, 0.04981832]),
-    array([-0.48670459,  0.58756912, -0.10226756,  0.63829658]),
-    array([ 0.23965404, -0.58538248,  0.187136  ,  0.75158061]),
-    array([ 0.658269  ,  0.25243989, -0.69118291,  0.158815  ])]
+    [array([ 0.18609797,  0.12416167,  0.28230062, -0.93287608]),
+    array([-0.48238484,  0.72089168,  0.4763625 ,  0.14387088]),
+    array([-0.66230111,  0.06548609, -0.6711639 , -0.32650855]),
+    array([ 0.54224502,  0.67868302, -0.49287336,  0.04935131])]
 
     It is also possible to add a seed for reproducibility.
 
     >>> from toqito.rand import random_orthonormal_basis
     >>> random_orthonormal_basis(2, is_real=True, seed=42)
-    [array([0.66954664, 0.74277002]), array([-0.74277002,  0.66954664])]
+    [array([0.37621414, 0.92653274]), array([-0.92653274,  0.37621414])]
 
 
     References
--- a/toqito/rand/random_unitary.py
+++ b/toqito/rand/random_unitary.py
@@ -102,7 +102,7 @@ def random_unitary(
         raise ValueError("Unitary matrix must be square.")
 
     # Construct the Ginibre ensemble.
-    gin = gen.random((dim[0], dim[1]))
+    gin = gen.standard_normal((dim[0], dim[1]))
 
     if not is_real:
         gin = gin + 1j * gen.standard_normal((dim[0], dim[1]))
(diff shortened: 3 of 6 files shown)


vprusso:
@jyotiraj-code  looks like some failing jobs due to the white space in the doc-string. You can look at other examples in the toqito project for reference in terms of how best to format output matrices in doc-strings. Hope that helps, but let me know if that's not clear.

jyotiraj-code:
Yes, @vprusso thanks for pointing out! I am using autopep8 to remove the white space issue in doc-string.

jyotiraj-code:
Ah, I see the remaining issues related to the `non-updation` of doc-tests after the update of the `random_unitary.py` methods.

vprusso:
> Ah, I see the remaining issues related to the `non-updation` of doc-tests after the update of the `random_unitary.py` methods.

Right, they appear to be due to the same thing where the matrices generated before used a different distribution:
https://github.com/vprusso/toqito/actions/runs/13996404372/job/39192552603?pr=1044#step:5:3058

Should be a matter of regenerating those matrices with the seed and updated distribution and then updating the docstring accordingly.

jyotiraj-code:
Hi @purva-thakre and @vprusso can you please review the code! (I am unable to tag under the reviewers section)
image

vprusso:
> Hi @purva-thakre and @vprusso can you please review the code! (I am unable to tag under the reviewers section) ![image]

You should be able to. If you click on the "Reviewers" title in the screenshot you provided, our names should appear. 

I added us as reviewers here, but in general, you should be able to add us and I'm not aware of any permissions restrictions. Perhaps the UI is just a bit counterintuitive though :)

toqito/rand/random_density_matrix.py
```diff
@@ -34,8 +34,8 @@ def random_density_matrix(
     >>> from toqito.rand import random_density_matrix
     >>> complex_dm = random_density_matrix(2)
     >>> complex_dm # doctest: +SKIP
```
vprusso, line 36: Is there a reason that we are skipping the doctest here and in other places? I think we want to remove the `+SKIP` here in anywhere is it is used (unless there is a compelling reason to keep it in).

vprusso (approved):
Minor comment about skipping some of the doc-string linting, but aside from that LGTM!

toqito/rand/random_density_matrix.py
jyotiraj-code: Actually I don't see any reason. When I was changing the doc-test some of the tests were like this i.e `+SKIP` from before, skipping the doc-tests. Do you want me to include those tests too? 

Although they were `+SKIP`, I changed them anyway to match with the recent change results.

jyotiraj-code:
> > Hi @purva-thakre and @vprusso can you please review the code! (I am unable to tag under the reviewers section) ![image]
> 
> You should be able to. If you click on the "Reviewers" title in the screenshot you provided, our names should appear.
> 
> I added us as reviewers here, but in general, you should be able to add us and I'm not aware of any permissions restrictions. Perhaps the UI is just a bit counterintuitive though :)

Thank you @vprusso! I don't know, I tried clicking on the "reviewers" title, but didn't show any options.

toqito/rand/random_density_matrix.py
vprusso: Right, in that case, I would remove the `+SKIP` as we should definitely be checking these and not ignoring them in the docs (unless there is some reason why we should, but I'm not aware of any). So yes, please remove, and thank you!

vprusso:
> Thank you @vprusso! I don't know, I tried clicking on the "reviewers" title, but didn't show any options. 

Weird. I'm not aware of any restrictions on reviewer request so it might be worthwhile to review the GitHub docs on this. Other users have been able to request us as reviewers so I'm wondering why you might be having issues with that.

toqito/rand/random_density_matrix.py
purva-thakre: As the example is not using `seed`, we are skipping the doctest check because a randomly generated function will generate a newer value for each run. You will see doctest failures otherwise.

vprusso: Yeah, I'm suggesting that we include the ability to feed in a seed. In priciniple, any function that is generating anything random should be taking one in any case. Does that ake sense, @jyotiraj-code ?

purva-thakre: > I think I'll close this one for now, but feel free to re-open if you disagree.

If you scroll down and expand the diff even more, there's already an example utilizing a seed.

vprusso: Think you might be responding to another comment, @purva-thakre 

The takeaway is that any function that computes something random should take an optional argument for the seed which will allow us to properly test (both in the unit tests and the docs) that specific matrices are generated with a fixed seed.

purva-thakre: > Think you might be responding to another comment

@vprusso Copy paste issues :sweat_smile:

> The takeaway is that any function that computes something random should take an optional argument for the seed 

Yes, what I am saying is that all the functions in this PR already do. 

image

image

image

There are two sets of examples in each docstring. One that uses seed and one without it. So, it is fine to use `SKIP` for those that do not use seed.

vprusso: Ah, I understand now. Yep, that makes sense, thanks for spelling it out for me :P

purva-thakre:
> Weird. I'm not aware of any restrictions on reviewer request so it might be worthwhile to review the GitHub docs on this. Other users have been able to request us as reviewers so I'm wondering why you might be having issues with that.

@vprusso If you or I have not reviewed a PR or made some comments on it, then a new contributor does not have access to the request review option. Once a comment is made by us, the recycle symbols show up in the `Reviewers` section.

vprusso:
> > Weird. I'm not aware of any restrictions on reviewer request so it might be worthwhile to review the GitHub docs on this. Other users have been able to request us as reviewers so I'm wondering why you might be having issues with that.
> 
> @vprusso If you or I have not reviewed a PR or made some comments on it, then a new contributor does not have access to the request review option. Once a comment is made by us, the recycle symbols show up in the `Reviewers` section.

Lame, GitHub! 

Good to know for future me!

jyotiraj-code:
@purva-thakre and @vprusso, thank you so much for your insights; I clearly missed the point that the `+SKIP` doc-tests didn't have seed to check the results in the first place. I hope the issue is resolved the way it is right now.

Is there any point I am missing out?

vprusso:
> @purva-thakre and @vprusso, thank you so much for your insights; I clearly missed the point that the `+SKIP` doc-tests didn't have seed to check the results in the first place. I hope the issue is resolved the way it is right now.
> 
> Is there any point I am missing out?

No worries at all, it was mostly me confusing both of you with my questions :) 

I think this LGTM, and I'll approve (unless @purva-thakre has any reservations).

vprusso (approved):

purva-thakre (approved):
LGTM too!

PR #1041, opened 2025-03-20
Removing explicit cvxopt dependency

jyotiraj-code:
## Description

Fixes #1040 

## Changes

  -  [x] Removed explicit installation of cvxopt from `setup.py`, `pyproject.toml`
  -  [x] Added the usage of `solver = SCS` instead of `solver=CVXOPT` for a `problem` method defined in `cvxpy` in file `toqito/channel_metrics/channel_fidelity.py`

## Checklist
Before marking your PR ready for review, make sure you checked the following locally. If this is your first PR, you might be notified of some workflow failures after a maintainer has approved the workflow jobs to be run on your PR. 

Additional information is available in the [documentation](https://toqito.readthedocs.io/en/latest/contributing.html#testing).

  -  [x] Use `ruff` for errors related to code style and formatting.
  -  [x] Verify all previous and newly added unit tests pass in `pytest`.
  -  [x] Check the documentation build does not lead to any failures. `Sphinx` build can be checked locally for any failures related to your PR
  -  [x] Use `linkcheck` to check for broken links in the documentation
  -  [x] Use `doctest` to verify the examples in the function docstrings work as expected.

Issue #1040, opened 2025-03-20
Remove dependencies from CVXOPT


jyotiraj-code:
The current version of `toqito(1.1.1)` create conflicts for `macos` in `python=3.13` due to dependency on CVXOPT. After a discussion in #1039 comments, it is decided to remove file dependencies of solver from `cvxopt` to `scs` with the motivation of:

1. Removing conflicts for macos users with python=3.13
2. Increase toqito package support to python=3.13 (Solves #1038)

vprusso:
Thanks, @jyotiraj-code . I think it is sensible to split these issues up as you did here.

jyotiraj-code:
Thanks @vprusso, while running tests for `state_metrics/fidelity_of_separability.py`. I came across the issue of `ValueError: The option 'solver' does not accept the value 'scs'.`

Digging up the log file (below):

[test_summary.log](https://github.com/user-attachments/files/19375950/test_summary.log)

I believe `picos` module is causing the `ValueError`. I came to know `picos` doesn't support `scs` as solver, as evident from the ([website](https://pypi.org/project/PICOS/))

Am I doing something wrong, or should we resort to another solver?

jyotiraj-code:
Ah, I see! `cvxopt` comes as extra with `cvxpy`. I will remove the explicit dependencies of CVXOPT without breaking the current files and tests.

vprusso:
Yeah, exactly. I likely could have made that more clear than I did, but exactly that.

PR #1063, opened 2025-03-22
Extend support to python=3.13 (deprecate 3.10)

### LINKED ISSUE #1001: Keep dependency versions unpinned to specific versions unless needed

jyotiraj-code:
## Description

_Removes support (deprecate) from 3.10 without any warning_ as illustrated as a probable solution in #1038 

Fixes #1001 and #1038 ( in a previously closed PR for it #1039 ) 

In the past PR, there was an issue with the code-build in macos versions due to non-updation of `cvxopt` wheels.

After the removal of explicit dependency on `cvxopt` with the PR #1040, it seems viable to extend to `python=3.13` version. (Hopefully after the build tests). The issue #1001 highlighted the usage of ">=" to keep the dependencies updated with the recent package developments, which seemed very relevant to solving this issue.

Some further comments:

- Required to change `PyPI` in `https://pypi.org/project/toqito/` documentation to showcase the dependency change `python>=3.11 to <=3.13`
- `pytest` result summary concludes with `1706 passed, 2 skipped, 44 warnings in 171.13s (0:02:51)`

## Changes

  -  [x] Updated `bug_report.md` to support `python>=3.11`
  -  [x] Updated "=" (pinned to specific versions) to ">=" while keeping the builds in the mind
  -  [x] Updated `build-test-actions.yml` to support `python>=3.11`to `<=3.13` (Expanding support matrix)
  -  [x] Updated `docs.yml` to support `python>=3.11`to `<=3.13` (Expanding support matrix from`python=3.12`)
  -  [x] Updated `README.md` text of supporting machines with `python>=3.11`
  -  [x] Updated text in `contributing.rst`
  -  [x] Updated text in `getting_started.rst` 
  -  [x] Updated `pyproject.toml` to support `python>=3.11`
  -  [x] Updated `setup.py` for python requirement `python>=3.11`

## Checklist

### DIFF
--- a/toqito/channel_props/__init__.py
+++ b/toqito/channel_props/__init__.py
@@ -8,3 +8,4 @@ from toqito.channel_props.choi_rank import choi_rank
 from toqito.channel_props.is_trace_preserving import is_trace_preserving
 from toqito.channel_props.is_quantum_channel import is_quantum_channel
 from toqito.channel_props.is_unitary import is_unitary
+from toqito.channel_props.is_extremal import is_extremal
--- a/toqito/channel_props/is_extremal.py
+++ b/toqito/channel_props/is_extremal.py
@@ -1,0 +1,94 @@
+"""Determines whether a quantum channel is extremal."""
+
+import numpy as np
+from numpy.linalg import matrix_rank
+
+from toqito.channel_ops.choi_to_kraus import choi_to_kraus
+
+
+def is_extremal(phi: np.ndarray | list[np.ndarray | list[np.ndarray]], tol: float = 1e-9) -> bool:
+    r"""Determine whether a quantum channel is extremal.
+
+    (Section 2.2.4: Extremal Channels from :cite:`Watrous_2018_TQI`).
+
+    Theorem 2.31 in :cite:`Watrous_2018_TQI` provides the characterization of extremal
+    quantum channels as a channel :math:`\Phi` is an extreme point of the convex set
+    of quantum channels if and only if the collection:
+
+    .. math::
+        \{ A_i^\dagger A_j \}_{i,j=1}^{r}
+
+    is linearly independent.
+
+    The channel can be provided in one of the following representations:
+
+    - A Choi matrix, representing the quantum channel in the Choi representation. It will
+      be converted internally to a set of Kraus operators.
+    - A list of Kraus operators, representing the channel in Kraus form.
+    - A nested list of Kraus operators, which will be flattened automatically.
+
+    Examples
+    ==========
+
+    The following demonstrates an example of an extremal quantum channel from Example 2.33
+    in :cite:`Watrous_2018_TQI`.
+
+    >>> import numpy as np
+    >>> from toqito.channel_props import is_extremal
+    >>> kraus_ops = [
+    ...     (1 / np.sqrt(6)) * np.array([[2, 0], [0, 1], [0, 1], [0, 0]]),
+    ...     (1 / np.sqrt(6)) * np.array([[0, 0], [1, 0], [1, 0], [0, 2]])
+    ... ]
+    >>> is_extremal(kraus_ops)
+    np.True_
+
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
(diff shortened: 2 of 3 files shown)


jyotiraj-code:
It seems @vprusso, @purva-thakre that we are unable to extend the support to `python=3.13` until `cvxopt` resolves the issue or we completely remove (explicits removed) `cvxopt` as a dependency.

pyproject.toml
```diff
@@ -10,40 +10,40 @@ repository = "https://github.com/vprusso/toqito"
...
 [tool.poetry.dependencies]
-python = ">=3.10,<4"
-cvxpy = "1.6.2"
-more-itertools = "10.6.0"
-numpy = "2.2.3"
-scipy = "1.15.2"
-scs = "3.2.7.post2"
-picos = "2.6.0"
+python = ">=3.11,<4"
+cvxpy = ">=1.6.2"
```
vprusso, line 14: Do we want to do `^=` instead of `>=` for these dependencies?

Using `>=` is the most permissive and can allow breaking changes if the major version changes. I'd say for safety, we want to opt for `^=` over `>=` for all of the third party package dependencies.

vprusso:
> It seems @vprusso, @purva-thakre that we are unable to extend the support to `python=3.13` until `cvxopt` resolves the issue or we completely remove (explicits removed) `cvxopt` as a dependency.

But wasn't cvxopt dropped in https://github.com/vprusso/toqito/pull/1041?

jyotiraj-code:
> > It seems @vprusso, @purva-thakre that we are unable to extend the support to `python=3.13` until `cvxopt` resolves the issue or we completely remove (explicits removed) `cvxopt` as a dependency.
> 
> But wasn't cvxopt dropped in #1041?

The explicit usage of it has been dropped along with direct dependency installations in `pyproject.toml` . However, we are still dependent on 'cvxopt` as a solver in various files across the project, under the `extra-package` from `cvxpy` installations.

Yes, I think that might be a valid way to do it.

pyproject.toml
jyotiraj-code: Thank you for the clarification @vprusso ; yes, I will change it for the third-party packages.

vprusso:
Posted an issue on their workboard here: https://github.com/cvxopt/cvxopt/issues/258

(FYI @jyotiraj-code)

jyotiraj-code:
> Posted an issue on their workboard here: [cvxopt/cvxopt#258](https://github.com/cvxopt/cvxopt/issues/258)
> 
> (FYI @jyotiraj-code)

Thank you so much! As per the issues:

2. I have already tried the usage of "SCS" instead of "CVXOPT" in the closed issue #1041, however, it throws ValueError because problems defined in various `.py modules` use `picos` to create the `problem variable` which is doesn't support "SCS" sadly :( ([Link](https://pypi.org/project/PICOS/))

For 1. I am a bit skeptical, but might be worthwhile to try?

vprusso:
> For 1. I am a bit skeptical, but might be worthwhile to try?

I think it's worth a try, is there a reason you are skeptical? It looks like they support MacOS builds for Python 3.13 (it's just not released on PyPi). So I think pinning the dependency to the GitHub *should* (hopefully) do the trick.

jyotiraj-code:
@vprusso I was directly using:

`cvxopt = { git = "https://github.com/cvxopt/cvxopt.git", branch = "master" }` while re-writing the poetry lock

But the `poetry install` fails to update:

image

vprusso:
> @vprusso I was directly using:
> 
> `cvxopt = { git = "https://github.com/cvxopt/cvxopt.git", branch = "master" }` while re-writing the poetry lock
> 
> But the `poetry install` fails to update:
> 
> image

Ah that is unfortunate. I also tried some older commit hashes and it appears to be the same story. 

As another alternative, instead of using SCS, for Picos, we could try either of the other two solvers that are supported that do SDPs and that are free. According to this table, that would be QICS and SMCP:
https://gitlab.com/picos-api/picos

PR #1071, opened 2025-03-24
Adding Kraus to channel 

### LINKED ISSUE #14: Feature: Convert kraus representation to channel (superoperator)
Given a collection of Kraus operators, convert them into the corresponding quantum channel (i.e. the superoperator). 

This would be placed under `channel_ops/kraus_to_channel.py`. Corresponding tests would also need to be created in the `tests/test_channel_ops/test_kraus_to_channel.py`file. Updating the docs would also need to be done as well.

Shivansh20128:
## Description

This PR is a work I progress for adding the function kraus_to_channel.

#Resolves #14

## Changes

  -  [ ] Added code for kraus_to_channel function.
  -  [ ] Added tests for the function.

## Checklist


toqito/channel_ops/tests/test_kraus_to_channel.py
```diff
@@ -0,0 +1,28 @@
+"""Tests for kraus to channel."""
+
+import numpy as np
+import pytest
+
+import toqito.state_ops
```
purva-thakre, line 6: Can you add a unit test where the superoperator acts on a state?

See Section 2.2 of https://arxiv.org/pdf/1509.02921 for more info.

It would also be pretty easy to construct a superoperator from the channels in Section 2.1.1

Shivansh20128: Before going into that, can you tell me if the function has been implemented correctly. Because I was not very sure about it. If you read the previous discussion on the issue, you will see that the two results, one from using the function on a vector, and other from directly applying a series of kraus operators were coming out to be transpose of each other. 
To tackle that, I modified the implementation of the function a little by changing the order of A and B ( from `(A, B.conj())` to  `(B, A.conj())`) But I am not sure about it. 

`
   super_op = sum(tensor(B, A.conj()) for A, B in kraus_list)`

vprusso: Hi @Shivansh20128 ,

Thanks for your response. From what I can tell (and what I recall) I believe your approach here is sensible. 

Perhaps one way to sanity check your approach would be to convert a set of Kraus operators to a quantum channel. Then, if you can use some of the existing functionality in `toqito` to check that you can either go from the channel back to the operators, or apply the channel in a way that gives the same result as a quantum channel that was arrived at through another means, perhaps that would be serve as some good sanity checks? 

I understand that some of the above was a bit vague, but in general, do you think that's a sensible approach?

vprusso: You might want to try other values of `rho`, but I think this is definitely the correct sanity check approach for what you are working on, and it seems right to me! Nice approach, @Shivansh20128 !

Shivansh20128: Okay. I will add more values of `rho` in the test cases.

toqito/channel_ops/kraus_to_channel.py
```diff
@@ -0,0 +1,54 @@
...
+    >>> kraus_list = [(kraus_1, kraus_1), (kraus_2, kraus_2)]
+    >>> kraus_to_channel(kraus_list)
+    array([[1, 0, 0, 1],
+           [0, 0, 0, 0],
+           [0, 0, 0, 0],
+           [0, 0, 0, 0]])
+
+    See Also
+    ========
+    choi_to_kraus, kraus_to_choi
```
purva-thakre, line 42:
```suggestion
    func:`.choi_to_kraus`, func:`.kraus_to_choi`
```

toqito/channel_ops/tests/test_kraus_to_channel.py
```diff
@@ -0,0 +1,87 @@
...
+    (np.sqrt(1 - 3 * p_2 / 4) * np.eye(2), np.sqrt(1 - 3 * p_2 / 4) * np.eye(2)),  # Identity
+    (np.sqrt(p_2 / 4) * np.array([[0, 1], [1, 0]]), np.sqrt(p_2 / 4) * np.array([[0, 1], [1, 0]])),  # X
+    (np.sqrt(p_2 / 4) * np.array([[0, -1j], [1j, 0]]), np.sqrt(p_2 / 4) * np.array([[0, -1j], [1j, 0]])),  # Y
+    (np.sqrt(p_2 / 4) * np.array([[1, 0], [0, -1]]), np.sqrt(p_2 / 4) * np.array([[1, 0], [0, -1]])),  # Z
+]
+
+kraus_operators_3 = [
+    (np.array([[1, 0], [0, 0]]), np.array([[1, 0], [0, 0]])),  # Projection onto |0⟩
+    (np.array([[0, 1], [0, 0]]), np.array([[0, 1], [0, 0]])),  # Projection onto |1⟩
+]
```
purva-thakre, line 44: We have commonly predefined channels in `toqito/channels`.

toqito/channel_ops/tests/test_kraus_to_channel.py
```diff
@@ -0,0 +1,87 @@
...
+
+vector = np.random.randint(-3, 3, (dim, 1))
+dm = toqito.matrix_ops.to_density_matrix(vector)
+vec_dm = toqito.matrix_ops.vec(dm)
+
+# Random quantum test states (density matrices)
+rho_1 = np.array([[0.5, 0.5], [0.5, 0.5]])
+rho_2 = np.array([[0.5, 0], [0, 0.5]])
+rho_3 = np.array([[1, 0], [0, 0]])
+rho_4 = np.array([[0, 0], [0, 1]])
```
purva-thakre, line 20: Look if we have ran

toqito/channel_ops/tests/test_kraus_to_channel.py
```diff
@@ -0,0 +1,87 @@
...
+
+A = np.random.rand(2, 2) + 1j * np.random.rand(2, 2)
+rho_5 = A @ A.conj().T
+rho_5 /= np.trace(rho_5)  # Normalize trace to 1
+
```
purva-thakre, line 25: We have quite a few functions to generate random things. See if those are useful, instead of manually defining your own randomly generated input.

https://toqito--1071.org.readthedocs.build/en/1071/autoapi/rand/index.html

toqito/channel_ops/tests/test_kraus_to_channel.py
```diff
@@ -0,0 +1,87 @@
...
+    rho_after_super_op = quantum_channel @ rho_vec
+    rho_after_super_op = rho_after_super_op.reshape(2, 2, order="F")  # Reshape back
+
+    # Compare both methods
+    print("Using apply_channel:\n", rho_after_channel)
+    print("Using superoperator:\n", rho_after_super_op)
+    print("Difference:\n", rho_after_channel - rho_after_super_op)
+
+    # The difference should be close to zero
+    assert np.allclose(rho_after_channel, rho_after_super_op), "Mismatch in quantum channel application!"
```
purva-thakre, line 87:
```suggestion
    assert np.allclose(rho_after_channel, rho_after_super_op)
```

We don't need this.

toqito/channel_ops/tests/test_kraus_to_channel.py
Shivansh20128: I will work on this later.
Thanks

toqito/channel_ops/tests/test_kraus_to_channel.py
Shivansh20128: I will work on this later.
Thanks

vprusso: Yes, I think specifically in this case the [`random_density_matrix`](https://github.com/vprusso/toqito/blob/master/toqito/rand/random_density_matrix.py) function would be useful!

toqito/channel_ops/kraus_to_channel.py
```diff
@@ -0,0 +1,54 @@
...
+    >>> kraus_list = [(kraus_1, kraus_1), (kraus_2, kraus_2)]
+    >>> kraus_to_channel(kraus_list)
+    array([[1, 0, 0, 1],
+           [0, 0, 0, 0],
+           [0, 0, 0, 0],
+           [0, 0, 0, 0]])
+
+    See Also
+    ========
+    func:`.choi_to_kraus`, func:`.kraus_to_choi`
```
purva-thakre, line 42: @Shivansh20128 something has gone wrong with the formatting for this. I was trying to make these clickable similar to what I did in #1091 

image

purva-thakre:
@Shivansh20128 Are you planning to work on this?

PR #1072, opened 2025-03-24
Renamed diamond_norm to diamond_distance, use picos instead of cvxopt, to compute the completely bounded trace norm and define diamond_distance as a wrapper of the latter

### LINKED ISSUE #1070: Possible enhancements on the `diamond_norm` function?
The [`channel_metrics.diamond_norm`](https://toqito.readthedocs.io/en/latest/autoapi/channel_metrics/diamond_norm/index.html) function currently takes two arguments which are the Choi representations of the Quantum channels. However, it should be possible to compute the diamond norm of a given quantum channel. This shouldn't require too much refactor given that the current code computes the difference between the inputs and then computes the diamond norm, as seen here:

https://github.com/vprusso/toqito/blob/51474ab73e45eef4eeb9deefa56f8bf2c526a50b/toqito/channel_metrics/diamond_norm.py#L65

If computing the diamond norm in another context of the distance between two quantum channels doesn't make sense, maybe should this function be renamed `diamond_distance`?

Another possible enhancement would be to move from `cvxpy` to `picos`, which would allow the use of other solvers like mosek.

Are these relevant issues?

**Describe the solution you'd like**

- Rename the current function `diamond_distance`
- Create a function `diamond_norm` if its use is relevant
- Use `picos` instead of `cvxpy`

tnemoz:
## Description

Fixes #1070.

## Changes

  -  `diamond_norm` has been renamed `diamond_distance`
  - `diamond_distance` is now a wrapper around `completely_bounded_trace_norm`
  - `completely_bounded_trace_norm` now uses `picos` to allow for other solvers

## Checklist


vprusso:
Ah, actually, it looks like some of the tests in the pipeline appear to be failing. @tnemoz

tnemoz:
The error comes from this line:
```python
from channel_metrics.completely_bounded_trace_norm import completely_bounded_trace_norm
```
though the logs don't show it. It's at least the one that triggers the error. I tried to change it in several ways but couldn't get something working, not to mention I don't see why it should fail in `test_is_separable.py` of all files...

I've tried to replace it by:
```python
from .completely_bounded_trace_norm import completely_bounded_trace_norm
```
and
```python
from toqito.channel_metrics import completely_bounded_trace_norm
```
(this one actually fails even individually and I don't know why) and
```python

```
but to no avail. "Funnily" enough, the error only shows up when running the whole test suite with `pytest toqito` (except when using the second import), all tests pass when running `pytest toqito/channel_metrics/tests/test_diamond_distance.py` or `pytest toqito/state_props/tests/test_is_separable.py` individually.

Would you have any guidance on this? I must say I'm quite out of ideas right now :')

vprusso:
Hi @tnemoz ,

Yes, I do believe I know what might be happening here, and this is likely occurring for the same reason as it is doing so in:
https://github.com/vprusso/toqito/pull/1077#issuecomment-2751037697

The tl;dr is that there is likely some type of circular import within the `channels/` module that is causing this cryptic stack trace to occur. There is a direct fix for this, which would be to follow the comment response in the linked issue I provided. While this will work, it would be good, as also mentioned in that thread, to find a proper fix as opposed to what I envision as just a band-aid solution. 

Let me know if that makes sense and if that approach works for you! Thank you for flagging this as well!

tnemoz:
> Hi @tnemoz ,
> 
> Yes, I do believe I know what might be happening here, and this is likely occurring for the same reason as it is doing so in: [#1077 (comment)](https://github.com/vprusso/toqito/pull/1077#issuecomment-2751037697)
> 
> The tl;dr is that there is likely some type of circular import within the `channels/` module that is causing this cryptic stack trace to occur. There is a direct fix for this, which would be to follow the comment response in the linked issue I provided. While this will work, it would be good, as also mentioned in that thread, to find a proper fix as opposed to what I envision as just a band-aid solution.
> 
> Let me know if that makes sense and if that approach works for you! Thank you for flagging this as well!

Thanks for your input! By moving the import of `completely_bounded_trace_distance` within the function, it does work as expected indeed. It's kind of frustrating though, I haven't managed to find a cleaner way :(

vprusso:
> Thanks for your input! By moving the import of `completely_bounded_trace_distance` within the function, it does work as expected indeed. It's kind of frustrating though, I haven't managed to find a cleaner way :(

@tnemoz, yes, it is quite frustrating to be sure. I've opened up the following issue to track, and optimistically, to eventually find a more optimal manner of resolving:
https://github.com/vprusso/toqito/issues/1081

For now, this approach you have will work, but it is indeed a band-aid solution. Thank you again for your input here!

purva-thakre:
@tnemoz Do you mind providing a reference for where you picked this up? Maybe I am not looking at the correct paper.

image

tnemoz:
@purva-thakre Unfortunately I can't, that's a mistake on my part! Either the docstring must be changed to remove "half of", or the function should add a $\frac12$ factor, in which case the tests must also be changed.

I could see arguments in both cases. Adding the $\frac12$ factor has a nice operational meaning, as it would return the maximal probability of distinguishing between the two quantum channels, but I think people are kind of used to channels being maximally distinguishable if their diamond distance is equal to 2 (like at the end of [this answer](https://cstheory.stackexchange.com/a/4920/74658)).

That's on me though, sorry for that!

PR #1075, opened 2025-03-24
Update docstrings in matrix_ops/vec.py 

### LINKED ISSUE #1073: Inconsistencies in matrix_ops/vec.py docstrings
I found some issues in the documentation of `vec.py` that could confuse users regarding the behavior of the `vec` operator:

- **The very first docstring reads**:  
  > "Vec operation is used to perform the **unvec** operation on a matrix."  
  This is misleading because the function is intended to perform the **vec** operation (converting a matrix into a column vector), not the unvec operation.

- **The docstring inside the function says**:  
  > "Stacks the rows of the matrix on top of each other to obtain the 'vec' representation of the matrix."  
  But the code uses:
  
  ```python
  return mat.reshape((-1, 1), order="F")
  ```
  
  which stacks the columns using Fortran ordering. The subsequent sentence that claims it "converts each row to column, and then continually stacks the columns on top of each other" is confusing. It would be clearer to state:  
  > "Reorders the matrix into a column vector by stacking the columns of the matrix sequentially."

- **Mathematical example inconsistency:**  
  The docstring provides a mathematical example. It displays the matrix:
  
  ```
  A = \begin{pmatrix}
          1 & 2 \\
          3 & 4
      \end{pmatrix}
  ```
  
  and then says:
  
  ```
  \text{vec}(A) = \left[1, 2, 3, 4 \right]^{T}.
  ```
  
  This notation implies that the rows are stacked (i.e., `[1, 2, 3, 4]^T`). But using `order="F"` actually produces:
  
  ```
  array([[1],
         [3],
         [2],
         [4]])
  ```
  
  which confirms that the columns are stacked (first column: `1, 3`; then second column: `2, 4`). The mathematical example should be updated to clearly show the column-major stacking.

---

The code itself is fine I think, only the documentation needs to be updated so that readers don't get confused.

---

CreativeBinBag:
## Description
This PR fixes the vec operator documentation in matrix_ops/ so it correctly describes the function’s behavior (column stacking using Fortran order) and updates the math example accordingly.

Fixes #1073 

## Changes
- [x] Removed the misleading "unvec" reference.
- [x] Updated the description to indicate that the function stacks columns, not rows.
- [x] Corrected the example math to reflect the proper output: `[1, 3, 2, 4]^T`.

## Checklist
 - [x] Used ruff for errors related to code style and formatting. All checks were passed.
 - [x] Verified all previous and newly added unit tests pass in pytest.
 - [x] The Sphinx build completed successfully. 
 - [x] Used linkcheck to check for broken links in the documentation
 - [x] The Sphinx doctest build completed successfully.

### DIFF
--- a/toqito/matrix_ops/vec.py
+++ b/toqito/matrix_ops/vec.py
@@ -1,4 +1,4 @@
-"""Vec operation is used to perform the unvec operation on a matrix."""
+"""Performs the vec operation on a matrix."""
 
 import numpy as np
 
@@ -8,12 +8,7 @@ def vec(mat: np.ndarray) -> np.ndarray:
 
     For more info, see Section: The Operator-Vector Correspondence from :cite:`Watrous_2018_TQI`.
 
-    Stacks the rows of the matrix on top of each other to
-    obtain the "vec" representation of the matrix.
-
-    The vec function is a linear mapping that in essence converts each row to
-    column, and then continually stacks the columns on top of each other.
-    An example is helpful.
+    The function reorders the given matrix into a column vector by stacking the columns of the matrix sequentially.
 
     For instance, for the following matrix:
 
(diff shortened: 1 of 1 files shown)


vprusso (approved):
Great job and catch, @CreativeBinBag ! I'll go ahead and merge this in. Thank you!

Issue #1060, opened 2025-03-22
Add pauli channel


purva-thakre:
https://github.com/nathanieljohnston/QETLAB/blob/0359eb36d73ebac1986879ec644dd684a1f62b6a/PauliChannel.m

RajdeepAher:
Please assign this to me, I would like to work on it!

vprusso:
Sounds good, @RajdeepAher ! Assigned!

vprusso:
@RajdeepAher there are two cases in the QETLAB implementation:

1. It's a CVX variable. In this case the `sum(p) > 1+tol` part of the if statement checks it
2. It's not a CVX variable. In this case the `RandomProbabilities` function is used. If you look at the definition (and by the name) it will generate a set of random probabilities, which by definition (and indeed this can be verified by looking at the function) sum to 1.

So the QETLAB version appears to be okay to me, unless I'm not understanding what you're saying here.

RajdeepAher:
1. Yes, you are correct on the first point

2. Regarding the second point, I reviewed the code and noticed that `RandomProbabilities` is not being triggered, as it is located in a different branch of the `if-else` statement.
```matlab
if(len_p == 1) % P is a scalar, so generate a random P-qubit Pauli channel
    q = p;
    len_p = 4^q;
    p = RandomProbabilities(len_p);  %<------ HERE
    
% Otherwise, do some error checking to make sure that P is a probability
% vector.
else
    % Let the user enter either a single vector P or a 4-by-4-by-...-by-4
    % array P.
    p = reshape(p,len_p,1);
    
    tol = len_p*eps^(3/4);
    if(isa(p,'cvx') == 0 && (min(p) < -tol || sum(p) > 1+tol))  %<------ HERE
        error('PauliChannel:InvalidP','P must be a probability vector: its entries must be non-negative and sum to 1.');
    end
```

3. Also in the example of docs I mentioned [above](https://github.com/vprusso/toqito/issues/1060#issuecomment-2747057224), I still don't feel convinced:

> It can also be observed in their [docs](https://qetlab.com/PauliChannel) too:
> 
> >> Phi = full(PauliChannel([1,2,3,4]/15))
> 
> Phi =
> 
>     0.3333         0         0   -0.2000
>          0    0.3333   -0.0667         0
>          0   -0.0667    0.3333         0
>    -0.2000         0         0    0.3333
> Here the vector `[1,2,3,4]/15` does not sum to 1.
> 
[...]

Issue #677, opened 2024-07-13
Move channel dim to channel props


vprusso:
It seems as if the `helper/channel_dim.py` file should live instead in `channel_props/channel_dim.py` as this is a property of a given quantum channel.

purva-thakre:
This is a duplicate of #244 

The failures in #680 are identical to the attempts I made to resolve the issue. Still don't know the cause behind `TypeError: 'module' object is not callable`

vprusso:
Okay, so I think I fixed this issue. Changes the import pattern a bit, but it works. Let me know what you think, @purva-thakre

purva-thakre:
Yes, but shouldn't we be consistent with all the other input patterns? I think the issue I ran into was the problem persists if we apply the same input pattern to all of them. If we use ruff for formatting, it will catch these differences. 

I say we keep this or the other issue open to understand why one input pattern works while the other doesn't.

vprusso:
Yes, I agree. It's definitely an awkward hack that is not consistent with the way we do things elsewhere. It's a difficult bug to diagnose as running the tests for `channel_props/` works fine without altering the pattern but then trying to run *all* the tests seems to fail if that pattern is not used. It's probably some "circular-import-type-thing" but it's not obvious to me what it is.  

Agree with keeping the issue open and attempting to fix properly.

PR #1077, opened 2025-03-25
feat:  Added Pauli Channel under toqito/channels

RajdeepAher:
## Description

Closes https://github.com/vprusso/toqito/issues/1060

## Changes

  -  [x] Added `pauli_channel.py` to simulate Phase Damping Channel.
  -  [x] Added corresponding coverage tests in `test_pauli_channel.py`.
  -  [x] Changed `__init__.py` in `toqito/channels`.
 
## Checklist


toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,112 @@
+"""Generates and applies Pauli Channel to a matrix."""
+
+from typing import Optional, Union
```
vprusso, line 3: [nit]: We don't use these in toqito and prefer using native types for typing. 
```suggestion
```

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,112 @@
...
+
+from toqito.channel_ops import kraus_to_choi
+from toqito.helper import update_odometer
+from toqito.matrices import pauli
+
+
+def pauli_channel(
+    prob: Union[float, np.ndarray] = 0,
+    kraus_ops: bool = False,
+    input_mat: Optional[np.ndarray] = None,
```
vprusso, line 16:
```suggestion
    input_mat: np.ndarray | None = None,

```

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if kraus_ops:
+        if input_mat is not None:
+            return Phi, output_mat, kraus_operators
+        else:
+            return Phi, kraus_operators
+    elif input_mat is not None:
+        return Phi, output_mat
+    else:
+        return Phi
```
vprusso, line 112: This may be simplified a bit I think (but please test this to ensure I didn't mess something up here). I think this is easier to follow:

```py
output_mat = None
if input_mat is not None:
    output_mat = sum(k @ input_mat @ k.conj().T for k in kraus_operators)

if kraus_ops:
    return (Phi, output_mat, kraus_operators) if input_mat is not None else (Phi, kraus_operators)

return (Phi, output_mat) if input_mat is not None else Phi
```

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,112 @@
...
+    Phi = sparse.csc_matrix((4**q, 4**q), dtype=complex)
+
+    kraus_operators = []
+    ind = np.zeros(q, dtype=int)
+
+    for j in range(len_p):
+        pauli_op = pauli(ind.tolist())
+        kraus_operators.append(np.sqrt(prob[j]) * pauli_op)
+        Phi += prob[j] * kraus_to_choi([[pauli_op, pauli_op.conj().T]])
+        ind = update_odometer(ind, 4 * np.ones(q, dtype=int))
```
vprusso, line 95: I've made things a bit easier to follow (I think), but let me know if you agree or not:

```py
# Convert prob to ndarray if it's not already
if np.isscalar(prob):
    q = int(prob)
    len_p = 4**q
    prob = np.random.rand(len_p)
    prob /= np.sum(prob)
else:
    prob = np.asarray(prob)

# Validate that prob is a proper probability vector
if np.any(prob < 0) or not np.isclose(np.sum(prob), 1):
    raise ValueError("Probabilities must be non-negative and sum to 1.")

len_p = len(prob)
q = int(round(np.log2(len_p) / 2))

if len_p != 4**q:
    raise ValueError("The length of the probability vector must be 4^Q for some integer Q (number of qubits).")

# Initialize Choi matrix and Kraus operator list
Phi = sparse.csc_matrix((len_p, len_p), dtype=complex)
kraus_operators = []

ind = np.zeros(q, dtype=int)

for j, p in enumerate(prob):
    pauli_op = pauli(ind.tolist())
    k_op = np.sqrt(p) * pauli_op
    kraus_operators.append(k_op)
    Phi += p * kraus_to_choi([[pauli_op, pauli_op.conj().T]])
    ind = update_odometer(ind, 4 * np.ones(q, dtype=int))
```

Also, please do check to make sure it doesn't inadvertently break something, as I'm free-handing this here :)

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,161 @@
...
+    [
+        # Single qubit.
+        (1, 4),
+        # Two qubits.
+        (2, 16),
+        # Three qubits.
+        (3, 64),
+    ],
+)
+def test_pauli_channel_dimensions(Q, expected_dim):
```
vprusso, line 47: [nit]: I think where possible and within reason, we want to avoid variable names like `Q` and `p`, especially for input arguments.

RajdeepAher:
I’m getting `pytest` failures in parts of the file that I haven’t modified. I’ve tried various fixes but haven’t been able to resolve this:
```
FAILED toqito/matrix_props/tests/test_is_block_positive.py::test_is_block_positive[input_mat1-True-False] - TypeError: 'module' object is not callable
FAILED toqito/matrix_props/tests/test_is_block_positive.py::test_is_block_positive[input_mat2-True-False] - TypeError: 'module' object is not callable
FAILED toqito/matrix_props/tests/test_is_block_positive.py::test_is_block_positive[input_mat3-True-False] - TypeError: 'module' object is not callable
FAILED toqito/matrix_props/tests/test_is_block_positive.py::test_is_block_positive[input_mat4-True-None] - TypeError: 'module' object is not callable
FAILED toqito/matrix_props/tests/test_is_block_positive.py::test_dim_None[input_mat1-True] - TypeError: 'module' object is not callable
FAILED toqito/matrix_props/tests/test_is_block_positive.py::test_dim_None[input_mat2-True] - TypeError: 'module' object is not callable
FAILED toqito/matrix_props/tests/test_sk_norm.py::test_s1_norm_werner[3-0.5] - TypeError: 'module' object is not callable
FAILED toqito/matrix_props/tests/test_sk_norm.py::test_s1_norm_werner[3--0.5] - TypeError: 'module' object is not callable
FAILED toqito/state_props/tests/test_is_separable.py::test_entangled_realignment_criterion - TypeError: 'module' object is not callable
FAILED toqito/state_props/tests/test_is_separable.py::test_separable_closeness_to_maximally_mixed_state - TypeError: 'module' object is not callable
FAILED toqito/state_props/tests/test_is_separable.py::test_separable_small_rank1_perturbation_of_maximally_mixed_state - TypeError: 'module' object is not callable
FAILED toqito/state_props/tests/test_is_separable.py::test_separable_schmidt_rank - TypeError: 'module' object is not callable
FAILED toqito/state_props/tests/test_is_separable.py::test_entangled_symmetric_extension - TypeError: 'module' object is not callable
```


 @vprusso, if you have some time, could you kindly take a look? Extremely sorry if this turns out to be too trivial. Thanks in advance!

vprusso:
> @vprusso, if you have some time, could you kindly take a look? Extremely sorry if this turns out to be too trivial. Thanks in advance!

Hi @RajdeepAher

So in this case, I know *what* the issue is and I am aware of a band-aid fix, but I'm not sure of the best way in which to solve the problem. 

The tl;dr is that if you want the tests to pass, you will need to move the `from toqito.channel_ops import kraus_to_choi` line to somewhere inside of the function (instead of outside the function). If you do this, it will work. 

However, that solution is rather unsatisfactory. The overarching reason as to why this occurs is due to some type of circular import (the error you get is not particularly insightful to this fact though). Since `kraus_to_choi` is used in other places, the order in which these things get imported becomes a bit of a web of dependencies. 

That being said, the solution I proposed will work, but if you now know the problem, if you are able to devise a better approach, I am all ears!

Does that make sense and help at all, @RajdeepAher ?

RajdeepAher:
Thank you @vprusso for your response, I do understand you and the fix works. I also feel this has something to with cyclical imports and deadlocks being formed during it. Will try to propose a solution for it, until then I will commit the `band-aid` fix

purva-thakre:
@RajdeepAher @vprusso The import issue is something we have noticed in the past. https://github.com/vprusso/toqito/issues/677#issuecomment-2229478388

The problem is if you apply the same pattern everywhere, the bug comes back again.

toqito/channels/pauli_channel.py
RajdeepAher: I have added a change for this in https://github.com/vprusso/toqito/pull/1077/commits/38f85982aaf73440d32f6437cd0f94e47db9a723

toqito/channels/pauli_channel.py
RajdeepAher: fixed in https://github.com/vprusso/toqito/pull/1077/commits/38f85982aaf73440d32f6437cd0f94e47db9a723

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,104 @@
...
+    :raises ValueError: If length of probability vector is not ``4^Q`` for some integer ``Q``.
+    :return: The Choi matrix of the channel.
+         If ``input_mat`` is provided, also returns the output matrix.
+         If ``kraus_ops`` is ``True``, returns Kraus operators as well.
+
+    """
+    if not isinstance(prob, np.ndarray):
+        if np.isscalar(prob):
+            q = int(prob)
+            len_p = 4**q
```
vprusso, line 69: We define `len_p` here, but then below it gets overwritten.

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,104 @@
...
+
+from toqito.channel_ops.kraus_to_choi import kraus_to_choi
+from toqito.helper.update_odometer import update_odometer
+from toqito.matrices.pauli import pauli
+
+
+def pauli_channel(
+    prob=0,
+    kraus_ops=False,
+    input_mat=None,
```
vprusso, line 14: We will want to type-hint these.

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,104 @@
...
+    :raises ValueError: If probabilities are negative or don't sum to 1.
+    :raises ValueError: If length of probability vector is not ``4^Q`` for some integer ``Q``.
+    :return: The Choi matrix of the channel.
+         If ``input_mat`` is provided, also returns the output matrix.
+         If ``kraus_ops`` is ``True``, returns Kraus operators as well.
+
+    """
+    if not isinstance(prob, np.ndarray):
+        if np.isscalar(prob):
+            q = int(prob)
```
vprusso, line 68: Do we want to convert this to an int if it is a float?

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,104 @@
...
+    the probability vector. For a given probability vector :math:`(p_1, \ldots, p_{4^Q})`, the
+    channel is defined as:
+
+    .. math::
+       \Phi(\rho) = \sum_{i=1}^{4^Q} p_i P_i \rho P_i^*
+
+    where :math:`P_i` are Pauli operators and :math:`Q` is the number of qubits.
+    If :code:`prob` is a scalar, it generates a random :code:`prob`-qubit Pauli channel.
+    The length of the probability vector (if provided) must be :math:`4^Q` for some
+    integer :math:`Q` (number of qubits).
```
vprusso, line 29: If you use `q` as a variable, you should also use "q" in the docstring for consistency (or vice-versa).

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Examples
+    ========
+
+    Generate a random single-qubit Pauli channel:
+
+    >>> from toqito.channels import pauli_channel
+    >>> choi_matrix = pauli_channel(prob = 1)
```
vprusso, line 37:
```suggestion
    >>> choi_matrix = pauli_channel(prob=1)

```

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,104 @@
...
+from scipy import sparse
+
+from toqito.channel_ops.kraus_to_choi import kraus_to_choi
+from toqito.helper.update_odometer import update_odometer
+from toqito.matrices.pauli import pauli
+
+
+def pauli_channel(
+    prob=0,
+    kraus_ops=False,
```
vprusso, line 13: It might be confusing to call this input `kraus_ops` as it might confuse the user that it takes as input a set of kraus ops. In this case, this is a flag as to whether to return the Kraus operators. Maybe we should be more explicit with the name?

toqito/state_props/is_separable.py
```diff
@@ -45,8 +47,8 @@ def is_separable(state: np.ndarray, dim: None | int | list[int] = None, level: i
     On the other hand, a random density matrix will be an entangled state (a separable state).
 
     >>> import numpy as np
-    >>> from toqito.rand import random_density_matrix
-    >>> from toqito.state_props import is_separable
+    >>> from toqito.rand.random_density_matrix import random_density_matrix
+    >>> from toqito.state_props.is_separable import is_separable
```
vprusso, line 51: Why are these changed here? (similarly for the other imports) is this to resolve the "circular import" issue?

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,104 @@
...
+        else:
+            prob = np.array(prob)
+
+    if np.any(prob < 0) or not np.isclose(np.sum(prob), 1):
+        raise ValueError("Probabilities must be non-negative and sum to 1.")
+
+    len_p = len(prob)
+    q = int(np.round(np.log2(len_p) / 2))
+
+    if len_p != 4**q:
```
vprusso, line 82:
```suggestion
    q = int(np.round(np.log2(len(prob)) / 2))

    if len(prob) != 4**q:

```

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,104 @@
...
+        raise ValueError("The length of the probability vector must be 4^Q for some integer Q (number of qubits).")
+
+    Phi = sparse.csc_matrix((4**q, 4**q), dtype=complex)
+
+    kraus_operators = []
+    ind = np.zeros(q, dtype=int)
+
+    for j in range(len_p):
```
vprusso, line 90:
```suggestion
    for j in range(len(prob)):

```

toqito/state_props/is_separable.py
RajdeepAher: Yes, it was to resolve the test errors caused due to circular imports

toqito/channels/pauli_channel.py
RajdeepAher: I plan to fix it by:
```python
if not isinstance(prob, np.ndarray):
        if np.isscalar(prob):
            q = int(prob)
            prob = np.random.rand(4**q)
            prob /= np.sum(prob)
        else:
            prob = np.array(prob)
```

vprusso: Think that's reasonable!

toqito/channels/pauli_channel.py
RajdeepAher: I plan to fix it by:
```python
def pauli_channel(
    prob: float | int | np.ndarray, 
    kraus_ops: bool = False, 
    input_mat: np.ndarray | None = None
) -> np.ndarray:
```

toqito/channels/pauli_channel.py
RajdeepAher: Ah I see, if we convert `1.9` to `1`, that would be a problem and we must add check for it! Thank you for pointing that out 

We can do this instead of raising an error for non-integer float everytime:
```python
def pauli_channel(
    prob: int | np.ndarray, 
    kraus_ops: bool = False, 
    input_mat: np.ndarray | None = None
) -> np.ndarray:
```

toqito/channels/pauli_channel.py
RajdeepAher: Yes, I we should definitely use more suggestive variable names:

```python

def pauli_channel(
    prob: int | np.ndarray, 
    return_kraus_ops: bool = False, 
    input_mat: np.ndarray | None = None
) -> np.ndarray:

```

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def pauli_channel(
+    prob: int | np.ndarray, return_kraus_ops: bool = False, input_mat: np.ndarray | None = None
+) -> np.ndarray:
+    r"""Generate and apply a Pauli channel to a matrix.
+
+    Generates the Choi matrix of a Pauli channel with given probabilities and optionally applies it
+    to an input matrix. The Pauli channel is defined by the set of Pauli operators weighted by
+    the probability vector. For a given probability vector :math:`(p_1, \ldots, p_{4^Q})`, the
```
vprusso, line 18:
```suggestion
    the probability vector. For a given probability vector :math:`(p_1, \ldots, p_{4^q})`, the

```

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,99 @@
...
+    for j in range(len(prob)):
+        pauli_op = pauli(ind.tolist())
+        kraus_operators.append(np.sqrt(prob[j]) * pauli_op)
+        Phi += prob[j] * kraus_to_choi([[pauli_op, pauli_op.conj().T]])
+        ind = update_odometer(ind, 4 * np.ones(q, dtype=int))
+
+    if input_mat is not None:
+        output_mat = sum(k @ input_mat @ k.conj().T for k in kraus_operators)
+    else:
+        output_mat = None
```
vprusso, line 94:
```suggestion
    output_mat = None
    if input_mat is not None:
        output_mat = sum(k @ input_mat @ k.conj().T for k in kraus_operators)
```

toqito/channels/pauli_channel.py
vprusso: Is the response to this comment perhaps for another comment? I'm not sure how the code and the comment are related in this case.

vprusso (commented):
A few extra comments, but I think after those it's looking close to being done. 

Would also be good for @purva-thakre to take a look as well when she is able to do so.

toqito/channels/pauli_channel.py
RajdeepAher: Sorry for not being clear, in this fix, I restrict `prob` to `int` and not let it be `float`

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,98 @@
...
+) -> np.ndarray:
+    r"""Generate and apply a Pauli channel to a matrix.
+
+    Generates the Choi matrix of a Pauli channel with given probabilities and optionally applies it
+    to an input matrix. The Pauli channel is defined by the set of Pauli operators weighted by
+    the probability vector. For a given probability vector :math:`(p_1, \ldots, p_{4^q})`, the
+    channel is defined as:
+
+    .. math::
+       \Phi(\rho) = \sum_{i=1}^{4^q} p_i P_i \rho P_i^*
```
purva-thakre, line 22: @RajdeepAher Can you add some information about the probabilities? As noticed in #1085, there can be some ambiguity about what's assumed...

RajdeepAher: I can mention the following

 ```python
 """\( p_i \) are **probabilities** satisfying \( p_i \geq 0 \) and \( \sum_i p_i = 1 \), ensuring the channel is trace-preserving and completely positive (valid quantum channel)."""
 ```
 
as I could not find an explicit mention of the definition of `p_i`  but I can add page 174 as the [1106.1445](https://arxiv.org/pdf/1106.1445) reference for `q = 1`, so that there won't be any confusion to the person referring to it.

purva-thakre: Since you allow two options for the probability, add information about the order in which $p_0$ is applied, and so on. 

See equation 197 of https://www.dias.ie/wp-content/uploads/2012/06/lecture-3-wilde.pdf

RajdeepAher: I'm sorry, I might be a bit worked up at the moment, but I’m afraid I don’t fully understand what you mean.  

Are you asking if the docstring should be:
```
:math:`P_i` are the Pauli operators for :math:`q`-qubit systems. These are tensor products of the single-qubit Pauli matrices :math:`\{I, X, Y, Z\}`, ordered lexicographically (e.g., for :math:`q=1`: :math:`I, X, Y, Z`).

:math:`p_i` are probabilities satisfying :math:`p_i \geq 0` and :math:`\sum_i p_i = 1`, ensuring the channel is trace-preserving and completely positive (valid quantum channel).

The index :math:`i` runs over all :math`:4^q` possible Pauli operators for :math:`q`-qubits.

```

vprusso (approved):
I'm sorry if I'm approving this twice, but modulo @purva-thakre 's comments, this LGTM!

toqito/channels/pauli_channel.py
purva-thakre: @RajdeepAher One of the inputs allowed by your function is a numpy array of the probabilities. As Pauli channel is defined using 4 different probabilities, you should make a note of how the provided probabilities are used. 

Is $p_0 = p_I$ or $p_0 = p_X$ $p_0 = p_Y$ $p_0 = p_Z$? What about $p_1$, $p_2$ or $p_3$? The only constraint is that $p_0 + p_1 + p_2 + p_3 = 1$.

If a user is interested in applying a pauli channel to their input state, they should know how to provide the probability array. Makes sense?

RajdeepAher: Thank you @purva-thakre for pointing that out!

I initially assumed that when we evaluate `pauli([0,1])`, it would intuitively evaluate to $\mathbf{I} \otimes \mathbf{X}$, as the [documentation](https://toqito.readthedocs.io/en/latest/autoapi/matrices/pauli/index.html) does not specify what happens when we pass a list. However, upon further inspection of `matrices/pauli.py`, we see that the following loop is responsible for its behavior:

```python
num_qubits = len(ind)  # ind = [0,1] in our case
pauli_mats = []
for i in range(num_qubits - 1, -1, -1):
    pauli_mats.append(pauli(ind[i], is_sparse))  # appends pauli[1] first, then pauli[0]
return tensor(pauli_mats)
```
This evaluates to $\mathbf{X} \otimes \mathbf{I}$ instead. I would prefer a lexicographically increasing sequence for increasing $i$, viz., $\mathbf{I}, \mathbf{X}, \mathbf{Y}, \mathbf{Z}, \mathbf{II}, \mathbf{IX}, \mathbf{IY}, \ldots$. The above claim was also verified through experiments, as shown below:

```bash
ind is set to [0, 1]
pauli for ind  is
 [[0. 1. 0. 0.]
 [1. 0. 0. 0.]
 [0. 0. 0. 1.]
 [0. 0. 1. 0.]]
********************************
ind is set to [0, 2]
pauli for ind  is
 [[0.+0.j 0.-1.j 0.+0.j 0.-0.j]
 [0.+1.j 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.-0.j 0.+0.j 0.-1.j]
 [0.+0.j 0.+0.j 0.+1.j 0.+0.j]]
```
  
  ---
  So, 
  1. Should I add `pauli_op = pauli(list(reversed(ind.tolist())))` in my existing code to enforce a lexicographically increasing order, or
  2. Should I stick with `QETLAB` conventions?

purva-thakre: > Should I add pauli_op = pauli(list(reversed(ind.tolist()))) in my existing code to enforce a lexicographically increasing order

What if we instead make this change in `pauli.py`? As well as update the documentation for `pauli.py`?

@vprusso Your thoughts?

vprusso: Yep, that makes sense to me if that aligns with @RajdeepAher as well. So long as we are clear about the intended behavior, I think it makes sense!

RajdeepAher: Sure, I will raise an issue and PR for doing the said changes in `pauli.py`, and then I can update the docstrings here.

Does that sound good @vprusso and @purva-thakre ?

purva-thakre: @RajdeepAher You can use this PR to make these changes.

RajdeepAher: That works too, thanks!

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,103 @@
...
+
+def pauli_channel(
+    prob: int | np.ndarray, return_kraus_ops: bool = False, input_mat: np.ndarray | None = None
+) -> np.ndarray:
+    r"""Generate and apply a Pauli channel to a matrix.
+
+    Generates the Choi matrix of a Pauli channel with given probabilities and optionally applies it
+    to an input matrix. The Pauli channel is defined by the set of Pauli operators weighted by
+    the probability vector. For a given probability vector :math:`(p_0, \ldots, p_{4^q -1 })`, the
+    channel is defined as:
```
purva-thakre, line 19:
```suggestion
    channel is defined as shown below. Here, $q$ is the number of qubits. 

```

toqito/channels/pauli_channel.py
```diff
@@ -0,0 +1,103 @@
...
+
+    .. math::
+       \Phi(\rho) = \sum_{i=0}^{4^q - 1} p_i P_i \rho P_i^*
+
+    where :math:`P_i` are Pauli operators generated by a lexographically increasing sequence of pauli operators of
+    length strictly equal to :math:`q`, and :math:`p_i` is the corresponding probability of that operator where,
+    :math:`q` is the number of qubits. For example, when :math:`q = 2`;
```
purva-thakre, line 26:
```suggestion
    length strictly equal to :math:`q`, and :math:`p_i` is the corresponding probability of that operator. For example, when :math:`q = 2`, 

```

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,167 @@
...
+        (2, 16),
+        # Three qubits.
+        (3, 64),
+    ],
+)
+def test_pauli_channel_dimensions(num_qubits, expected_dim):
+    """Test Pauli channel generation for different numbers of qubits."""
+    p = np.ones(expected_dim) / expected_dim
+    Phi = pauli_channel(prob=p)
+    assert Phi.shape == (expected_dim, expected_dim), f"Incorrect matrix dimensions for {num_qubits} qubits"
```
purva-thakre, line 51: We don't need these strings. 
```suggestion
    assert Phi.shape == (expected_dim, expected_dim)

```

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,167 @@
...
+
+    - The output matrix remains Hermitian.
+    - The output matrix is positive semidefinite.
+    - The trace of the input matrix is preserved in the output.
+
+    """
+    prob = prob / np.sum(prob)
+    _, output_mat = pauli_channel(prob=prob, input_mat=input_mat)
+
+    assert np.allclose(output_mat, output_mat.conj().T), "Output matrix is not Hermitian."
```
purva-thakre, line 98:
```suggestion
    assert np.allclose(output_mat, output_mat.conj().T)

```

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,167 @@
...
+
+    eig_v = eigvalsh(output_mat)
+    assert np.all(eig_v >= -1e-10), "Output matrix is not positive semidefinite."
```
purva-thakre, line 101:
```suggestion
    assert np.all(eig_v >= -1e-10)
```

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,167 @@
...
+
+    in_trace = np.trace(input_mat)
+    out_trace = np.trace(output_mat)
+    assert np.isclose(out_trace, in_trace), "Trace is not preserved in the output matrix."
```
purva-thakre, line 105:
```suggestion
    assert np.isclose(out_trace, in_trace)
```

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,167 @@
...
+    - Its partial trace over the output system yields the identity matrix of the input Hilbert space.
+    - Its total trace equals the input dimension `(2**q)`.
+
+    """
+    Phi, _ = pauli_channel(prob=num_qubits, return_kraus_ops=True)
+    Phi = np.array(Phi)
+    expected_choi_dim = 4**num_qubits
+    assert Phi.shape == (expected_choi_dim, expected_choi_dim), (
+        f"Expected shape {(expected_choi_dim, expected_choi_dim)}, got {Phi.shape}"
+    )
```
purva-thakre, line 131:
```suggestion
    assert Phi.shape == (expected_choi_dim, expected_choi_dim)

```

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,167 @@
...
+
+    eigenvalues = eigvalsh(Phi)
+    assert np.all(eigenvalues >= -1e-10), f"Phi is not positive semidefinite for {num_qubits} qubits"
```
purva-thakre, line 134:
```suggestion
    assert np.all(eigenvalues >= -1e-10)

```

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,167 @@
...
+
+    input_dim = 2**num_qubits
+    dims = [input_dim, input_dim]
+    pt_output = partial_trace(Phi, sys=1, dim=dims)
+    identity_input = np.eye(input_dim)
+    assert np.allclose(pt_output, identity_input), f"Partial trace does not equal identity for {num_qubits} qubits"
```
purva-thakre, line 140:
```suggestion
    assert np.allclose(pt_output, identity_input)

```

toqito/channels/tests/test_pauli_channel.py
```diff
@@ -0,0 +1,167 @@
...
+
+    total_trace = np.trace(Phi)
+    assert np.isclose(total_trace, input_dim), (
+        f"Total trace ({total_trace}) does not equal input dimension ({input_dim}) for {num_qubits} qubits"
+    )
```
purva-thakre, line 145:
```suggestion
    assert np.isclose(total_trace, input_dim)

```

purva-thakre (approved):

purva-thakre:
@RajdeepAher I requested some minor changes.

If you go to the `files changed` tab, you can commit all the requested changes in a single batch. 

image

PR #1109, opened 2025-03-31
feat: Add Operator Sinkhorn

### LINKED ISSUE #848: Feature: OperatorSinkhorn function
One of the prerequisites to be able to generate a set of random quantum channels (quantum superoperators) will be to implement the `operator_sinkhorn.py` file. 

For a MATLAB reference of an implementation of this issue, refer to [OperatorSinkhorn](https://github.com/nathanieljohnston/QETLAB/blob/master/OperatorSinkhorn.m) from the QETLAB package.

NV-Karthik:
## Description
Implements the Iterative Sinkhorn algorithm. Closes #848

## Changes
  - [x] Added the Operator Sinkhorn function
  - [x] Borrowed testcases from PR #849 and added some more
  - [x] Added documentation and examples 

## Checklist
  -  [x] Use `linkcheck` to check for broken links in the documentation - no links added with this PR


NV-Karthik:
Hi @vprusso @purva-thakre, Please have a look at the Sinkhorn Operator implementation. I added some more tests to what's there in PR #849 to improve coverage.

purva-thakre:
@NV-Karthik Still looks like not all lines are covered by unit tests. 

https://app.codecov.io/gh/vprusso/toqito/pull/1109?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=Vincent+Russo

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+    iterations = 0
+    rho2 = rho.copy().astype(np.complex128)
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=RuntimeWarning)
+        while it_err > tol:
+
+            if iterations > max_iterations:
+                raise RuntimeError(f"operator_sinkhorn did not converge within {max_iterations} iterations.")
+
```
purva-thakre, line 145: I don't think you tested for this.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+    # Perform the operator Sinkhorn iteration.
+    it_err = 1.0
+    max_cond = 0
+    max_iterations = 10000
+    iterations = 0
+    rho2 = rho.copy().astype(np.complex128)
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=RuntimeWarning)
+        while it_err > tol:
```
vprusso, line 141: I'm not sure if we want to necessarily use the `warning` module for this behavior.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+    Prho = [np.eye(d, dtype=np.complex128) / d for d in dim]  # Force complex128 type for Prho
+    F = [np.eye(d, dtype=np.complex128) * tr_rho_p for d in dim]  # Force complex128 for F
+    ldim = [np.prod(dim[:j]) for j in range(num_sys)]
+    rdim = [np.prod(dim[j+1:]) for j in range(num_sys)]
+
+    # Perform the operator Sinkhorn iteration.
+    it_err = 1.0
+    max_cond = 0
+    max_iterations = 10000
+    iterations = 0
```
vprusso, line 136: Should these instead be provided as optional parameters to the function?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+
+                    max_cond = max(max_cond, np.linalg.cond(F[j]))
+
+            except Exception:
+                error_flag_in_iteration = True
+
+            # Check the condition number to ensure invertibility.
+            if (error_flag_in_iteration) or (max_cond >= 1 / tol) or (np.isinf(max_cond)):
+                raise ValueError("The operator Sinkhorn iteration does not converge for RHO. "
+                             "This is often the case if RHO is not of full rank.")
```
vprusso, line 188: Seems like there is some odd indentation going on with the multi-line strings here.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+
+            iterations += 1
+
+    # Stabilize the output for numerical reasons.
+    sigma = (rho2 + rho2.conj().T) / 2
+    sigma = tr_rho * sigma / np.trace(sigma)
+
+    # handle cases where trace(sigma) can be near zero: TODO
```
vprusso, line 196: Just flagging this as something that needs to be done.

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,131 @@
...
+def test_operator_sinkhorn_valid_single_dim():
+    """Test operator Sinkhorn when a single valid number is passed as `dim`."""
+    rho = random_density_matrix(9)  # 9-dimensional density matrix
+
+    # The dimension `3` divides evenly into `9`, so no error should occur
+    sigma, F = operator_sinkhorn(rho, dim=[3])
+
+    # Check that sigma is a valid density matrix with trace equal to 1
+    np.testing.assert_almost_equal(np.trace(sigma), 1)
+
```
vprusso, line 97: Be sure to be mindful of spacing between functions (2 carriage returns) as is done elsewhere.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
+"""Sinkhorn operator."""
+
+import warnings
+
+import numpy as np
+import scipy.linalg
```
vprusso, line 6: For consistency, let's `import scipy as sp` and then use `sp.linalg.sqrtm` when needed.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+    :param dim: `None` list containing dimensions of each subsystem.
+                Assumes 2 subsystems with equal dimensions as default.
+    :param tol: `np.sqrt(np.finfo(float).eps)` Convergence tolerance of the iterative Sinkhorn Algorithm.
+                Assumes square root of numpy eps as default.
+    :raises: ValueError: if input density matrix is not a square matrix.
+    :raises: ValueError: if the product of dimensions provided/assumed does not match the dimension of density matrix.
+    :raises: RuntimeError: if the sinkhorn algorithm doesnot converge before 10000 iterations.
+    :raises: ValueError: if the density matrix provided is singular (or is not of full rank).
+
+    """ # noqa: E501
```
vprusso, line 92: Why is there a `noqa` on this line?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+    # Run checks on the input density matrix
+    rho = np.asarray(rho)
+    if rho.ndim != 2 or rho.shape[0] != rho.shape[1]:
```
vprusso, line 95: We already have a function that checks whether a matrix is square in `matrix_props.is_square`.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+        raise ValueError("Input 'rho' must be a square matrix.")
+
+    rho = rho.astype(np.complex128)  # Ensure complex128 type for rho
+
+    dX = rho.shape[0]
+    sdX = int(round(np.sqrt(dX)))
+    tr_rho = np.trace(rho)
```
vprusso, line 102: I don't think we need a variable to house this, we can just compute the trace when we need it.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+    if dim is None:
+        dim = [sdX, sdX]
+
+    # allow the user to enter a single number for dim.
+    if len(dim) == 1:
+        dim.append(dX / dim[0])
+
+        if abs(dim[1] - round(dim[1])) >= (2 * dX * np.finfo(float).eps):
+            raise ValueError("If DIM is a scalar, X must be square and DIM must evenly divide length(X); "
+                             "please provide the DIM array containing the dimensions of the subsystems.")
```
vprusso, line 114: This message seems like it was taken directly from QETLAB. Since we are using different variable names, conventions, etc., we want to make sure that we are abiding by that in the error messages here.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+    # dimensions check before starting iterations
+    if np.prod(dim) != dX:
+        raise ValueError(f"Product of dimensions {dim} does not match rho dimension {dX}.")
+
+    # precompute trace for all iterations
+    tr_rho_p = np.power(tr_rho, 1.0 / (2.0 * num_sys)) if tr_rho != 0 else 1.0
+
+    # Prepare the iteration.
+    Prho = [np.eye(d, dtype=np.complex128) / d for d in dim]  # Force complex128 type for Prho
+    F = [np.eye(d, dtype=np.complex128) * tr_rho_p for d in dim]  # Force complex128 for F
```
vprusso, line 128: The "force" comments aren't particularly helpful, as we can tell this by what the line of code is doing. Also be sure to put the comments above the line in question to which they are referring to.

toqito/matrices/operator_sinkhorn.py
NV-Karthik: If you look at line 67, the python output of F is being flagged as a "line-too-long error". I cannot make edits to it since it would flag a doctest error. So I had to force ruff to ignore line too long error for that docstring.

That was the solution I could find, Please let me know if there are better ways to handle these

toqito/matrices/operator_sinkhorn.py
NV-Karthik: Yes, I tried to. But I didn't find any valid density matrices which take more than 10000 iterations to converge.

I included that check just for additional safety. But making max_iterations an input argument can solve this.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+    [array([[ 1.33132101+0.00313432j, -0.38927801+0.14985847j],
+           [-0.37199021-0.14296461j,  1.24635929-0.00300407j]]), array([[ 1.16317427+0.00765735j, -0.16925696+0.09054309j],
+           [-0.18974418-0.09028502j,  0.93359381-0.00777829j]])]
+
+    Similarly to perform sinkhorn operation on multipartite state where the
+    subsystems are of dimensions 2, 3 and 4.
+
+    >>> # rho_random = random_density_matrix(24)
+    >>> # sigma, F = operator_sinkhorn(rho=rho_random, dim=[2, 3, 4])
+    >>> # print(sigma, F)
```
vprusso, line 75: Why are these commented?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,198 @@
...
+     [-0.02275152-0.05104798j  0.1039554 +0.0350973j   0.15215814+0.j
+      -0.09278034-0.00551919j]
+     [ 0.20443565+0.12511156j  0.02275152+0.05104798j -0.09278034+0.00551919j
+       0.34784186+0.j        ]]
+
+    >>> print("F", F, sep='\n')
+    F
+    [array([[ 1.33132101+0.00313432j, -0.38927801+0.14985847j],
+           [-0.37199021-0.14296461j,  1.24635929-0.00300407j]]), array([[ 1.16317427+0.00765735j, -0.16925696+0.09054309j],
+           [-0.18974418-0.09028502j,  0.93359381-0.00777829j]])]
```
vprusso, line 68: It's possible to print these in a way where the output does not overflow onto the side. Also, is there a reason you do `print("sigma", sigma, sep='\n')` instead of just `print(sigma)`? Same for the subsequent print as well.

toqito/matrices/operator_sinkhorn.py
vprusso: Yes, but you can print out this content without it spilling over onto the next line. One way, since it's a list of arrays, is just to print out each element of the list as a separate print statement (for example). I don't think we should be `noqa`-ing things unless there is a very good reason to do so.

vprusso:
Hi @NV-Karthik . Are you still intending to work on this MR? There appear to still be some outstanding comments that have yet to be resolved and the pipeline also appears to be failing with the current state of the code.

NV-Karthik:
Hi. Yes I'm still figuring out a better fix for the failing tests. Will push "with the fix with imports inside the function" by tomorrow if I couldn't figure out one. Sorry I didn't update you on this.

vprusso:
No problem, thank you for the update, @NV-Karthik !

toqito/matrices/operator_sinkhorn.py
NV-Karthik: Makes sense. I tried to replicate QETLAB's behavior, but errors/warnings are now better handled with the try/except. Removed that part now

toqito/matrices/operator_sinkhorn.py
NV-Karthik: I have updated variable names with toqito specific ones. Let me know if further changes are needed.

toqito/matrices/operator_sinkhorn.py
NV-Karthik: My reasoning behind this was to give an example on tripartite systems with large density matrices. But including its results would clutter the docs. So I commented it. 

But yes, commented code doesn't look like helping here. Will better remove this example.

toqito/matrices/operator_sinkhorn.py
NV-Karthik: It was just to improve readability. But yes, `print(sigma)` looks better. Will change it back

vprusso: Got it, thank you!

toqito/matrices/operator_sinkhorn.py
vprusso: Makes sense to me, thank you @NV-Karthik !

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,179 @@
...
+def test_operator_sinkhorn_max_iterations():
+    """Test operator sinkhorn on insufficient iteration limit."""
+    # function should raise a RuntimeError
+
+    rho_random = random_density_matrix(4, seed=42)
+    try:
+        operator_sinkhorn(rho=rho_random, dim=[2, 2], max_iterations=20)
+    except RuntimeError as e:
+        expected_msg = "operator_sinkhorn did not converge within 20 iterations."
+        assert str(e) == expected_msg
```
purva-thakre, line 159: most of these raises error tests can be condensed into one. If you would like an example, search through the existing unit tests. 

https://docs.pytest.org/en/7.1.x/how-to/parametrize.html

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,179 @@
...
+    dim1 = [4, 3, 2]  # 4*3*2 != 8
+
+    rho = random_density_matrix(8)  # 8-dimensional density matrix
+
+    # The dimension `4` does not divide evenly into `8`, so we expect an error
+    try:
+        operator_sinkhorn(rho, dim=dim1)
+    except ValueError as e:
+        expected_msg = f"Product of dimensions {dim1} does not match rho dimension {dX}."
+        assert str(e) == expected_msg
```
purva-thakre, line 100: same as my other comment about parametrize

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,179 @@
...
+    """Test operator Sinkhorn on a maximally entangled bipartite state. Should be invariant."""
+    # function should return the inintial state since it already satisfies the trace property
+
+    u0 = bell(0)
+    rho = u0 @ u0.conj().T
+
+    sigma, F = operator_sinkhorn(rho, dim=[2])
+
+    # Check that rho is invariant after sinkhorn operation
+    np.testing.assert_almost_equal(sigma, rho)
```
purva-thakre, line 135: If multiple unit tests follow the same pattern, it's better to condense them into one through parameterized.

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Expected partial trace should be proportional to identity matrix.
+    expected_identity = np.eye(3) * (1 / 3)
+
+    # Partial trace on the first subsystem.
+    pt = partial_trace(sigma, 0, [3, 3])
+    pt_rounded = np.around(pt, decimals=2)
+
+    # Check that partial trace matches the expected identity.
+    np.testing.assert_array_almost_equal(pt_rounded, expected_identity, decimal=2)
```
purva-thakre, line 38: Vincent recently removed `assert_array_almost_equal` in #1167 

Read the PR description for why this is a bad choice.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,195 @@
...
+    [[ 1.33132101+0.00313432j -0.38927801+0.14985847j]
+     [-0.37199021-0.14296461j  1.24635929-0.00300407j]]
+    >>> print(F[1])
+    [[ 1.16317427+0.00765735j -0.16925696+0.09054309j]
+     [-0.18974418-0.09028502j  0.93359381-0.00777829j]]
+
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
```
purva-thakre, line 77: If you are not citing a reference, you do not need this section.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,195 @@
...
+      0.39298553+0.j        ]]
+
+    >>> sigma, F = operator_sinkhorn(rho=rho_random, dim=[2, 2])
+
+    This returns the result density matrix along with the operations list F.
+    SIGMA has all of its (single-party) reduced density matrices
+    proportional to the identity, and SIGMA = Tensor(F)*RHO*Tensor(F)'. In
+    other words, F contains invertible local operations that demonstrate
+    that RHO and SIGMA are locally equivalent.
+
```
purva-thakre, line 51: Most of these symbols should be in latex. See existing examples in toqito, if you are new to latex.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,195 @@
...
+from toqito.channels.partial_trace import partial_trace
+from toqito.matrix_props.is_square import is_square
+
+
+def operator_sinkhorn(
+    rho: np.ndarray, dim: list[int] = None, tol: float = np.sqrt(np.finfo(float).eps), max_iterations: int = 10000
+) -> tuple[np.ndarray, list[np.ndarray]]:
+    r"""Perform the operator Sinkhorn iteration.
+
+    This function implements the iterative Sinkhorn algorithm to find a density matrix
```
purva-thakre, line 17: What is the Sinkhorn algorithm? Describe it briefly. 

Add a reference to this too.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,195 @@
...
+    that is locally equivalent to a given bipartite or multipartite density matrix
+    `rho`, but has both of its partial traces proportional to the identity.
```
purva-thakre, line 19:
```suggestion
    `rho`, but has both of its partial traces proportional to the identity.
```

Either use latex for `rho` or use double `

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,195 @@
...
+
+    Examples
+    ==========
+
+    To Perform Operator Sinkhorn on a random 2-Qubit Bi-partite state:
```
purva-thakre, line 24:
```suggestion
    To Perform Operator Sinkhorn on a random 2-Qubit Bipartite state:
```

docs/refs.bib
```diff
@@ -465,6 +465,21 @@ @article{Gupta_2024_Optimal
   year={2024}
 }
 
+@article{GURVITS2004448,
```
purva-thakre, line 468: take a look at the other keys in this file and format it correctly. 

Keys are arranged alphabetically in this file.

NV-Karthik: Do you mean, I have to rename the key to something more meaningful?, I kept the default key in this case

toqito/matrices/tests/test_operator_sinkhorn.py
NV-Karthik: all repetitive tests which can be parameterized are now clubbed into one function now. only the ones which call the base function differently are in seperate functions. Let me know if its good now

toqito/matrices/tests/test_operator_sinkhorn.py
NV-Karthik: Fixed. now all instances where two complex arrays are compared will use `assert_allclose`

docs/refs.bib
purva-thakre: look at the entry right above your addition. how is the key formatted?

docs/refs.bib
```diff
@@ -466,18 +466,18 @@ @article{Gupta_2024_Optimal
...
+  number={3},
+  pages={448-484},
+  year={2004},
+  note={Special Issue on STOC 2003},
+  issn={0022-0000},
+  doi={https://doi.org/10.1016/j.jcss.2004.06.003},
+  url={https://www.sciencedirect.com/science/article/pii/S0022000004000893},
+  author={Leonid Gurvits},
+  keywords={Entanglement, Complexity, Determinant}
 }
```
purva-thakre, line 481: @NV-Karthik I am referring to the key, the rest was fine. 

image

NV-Karthik: Please check it now. Got what you were referring to now.

docs/refs.bib
```diff
@@ -981,6 +996,19 @@ @misc{Sikora_2019_Semidefinite
   url={https://sites.google.com/site/jamiesikora/teaching/sdps-in-quantum-theory}
 }
 
+@article{Sinkhorn_1964_ARelationship,
```
vprusso, line 999: Convention is to drop articles (i.e. "a", "the", "an", etc.) and only keep the first word. 

```suggestion
@article{Sinkhorn_1964_Relationship,
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+
+    To Perform Operator Sinkhorn on a random 2-Qubit Bipartite state:
+
+    >>> import numpy as np
+    >>> from toqito.matrices import operator_sinkhorn
+    >>> from toqito.rand import random_density_matrix
+
+    >>> rho_random = random_density_matrix(4, seed=42)
+
+    returns a random 4x4 complex matrix like this
```
vprusso, line 42: I don't think you need this comment here, it is implicit what is being returned by the function above.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+     [0.11523158+0.02017518j 0.14478708+0.02004061j 0.11824612+0.j
+      0.09088248-0.05551508j]
+     [0.18524981+0.13636277j 0.22496295+0.11837381j 0.09088248+0.05551508j
+      0.39298553+0.j        ]]
+
+    >>> sigma, F = operator_sinkhorn(rho=rho_random, dim=[2, 2])
+
+    This returns the result density matrix along with the operations list F.
+    `sigma` (:math:`\sigma`) has all of its (single-party) reduced density matrices
+    proportional to the identity, while satisfying
```
vprusso, line 58: Perhaps this description should preceede the example code?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+    return the list of Filtering Operations which can be directly applied on :math:`\rho` to arrive at :math:`\sigma`.
+    Such converted forms of density matrices are useful to analyse entangled systems and to study effects of operations
+    on each subsystems. (as discussed in :cite:`Gurvits_2004_Classical`)
+
+    Examples
+    ==========
+
+    To Perform Operator Sinkhorn on a random 2-Qubit Bipartite state:
+
+    >>> import numpy as np
```
vprusso, line 36: We now use `..jupyter-execute` for code blocks (you can see an example by looking in the `toqito` project for where this is currently used.

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,182 @@
...
+# The dimension `3` divides evenly into `9`, so no error should occur
+dim_mixed = [3]
+
+"""Test operator Sinkhorn on a maximally entangled bipartite state. Should be invariant."""
+# function should return the inintial state since it already satisfies the trace property
+u0 = bell(0)
+rho_ent = u0 @ u0.conj().T
+dim_ent = [2]
+
+params2 = [([rho_mixed, dim_mixed]), ([rho_ent, dim_ent])]
```
vprusso, line 171: Shouldn't these be parameters of `pytest.parameterize` instead of free-floating variables that live above a test? IOW, you can define them within the `parameterize` call. You can refer to other tests that use `pytest.parameterize` in the toqtio project for reference.

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,182 @@
...
+
+    try:
+        operator_sinkhorn(rho, dim=dim1)
+    except ValueError as e:
+        assert str(e) == expected_msg
+
+
+def test_operator_sinkhorn_max_iterations():
+    """Test operator sinkhorn on insufficient iteration limit."""
+    # function should raise a RuntimeError
```
vprusso, line 128:
```suggestion
```

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,182 @@
...
+    pt = partial_trace(sigma, target_subsys, dim=dim1)
+    pt_rounded = np.around(pt, decimals=2)
+
+    # Check that partial trace matches the expected identity.
+    np.testing.assert_allclose(pt_rounded, expected_identity, rtol=1e-2)
+
+
+# Test Error handling in different scenarios
+
+"""Test operator Sinkhorn with a singular matrix that triggers LinAlgError."""
```
vprusso, line 76: These test cases should be within the paramtereize call, not defined outside of it. Refer to something that uses `pytest.paramterize` within the project (one such example, out of many, could be `test_is_pseudo_hermitian.py` which uses the convention we want to adopt here.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+
+                max_cond = max(max_cond, np.linalg.cond(F[j]))
+
+        except Exception:
+            error_flag_in_iteration = True
+
+        # Check the condition number to ensure invertibility.
+        if (error_flag_in_iteration) or (max_cond >= 1 / tol) or (np.isinf(max_cond)):
+            raise ValueError(
+                "The operator Sinkhorn iteration does not converge for RHO. "
```
vprusso, line 192: In our case, can we use `rho` in place of `RHO`?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+        if abs(dim[1] - round(dim[1])) >= (2 * dX * np.finfo(float).eps):
+            raise ValueError(
+                "If dim is of size 1, rho must be square and dim[0] must evenly divide length(rho); "
+                "please provide the dim array containing the dimensions of the subsystems."
+            )
+
+        dim[1] = round(dim[1])
+
+    num_sys = len(dim)
+    # dimensions check before starting iterations
```
vprusso, line 132:
```suggestion
    # Dimension check before starting iterations.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+        dim = [sdX, sdX]
+
+    # allow the user to enter a single number for dim.
+    if len(dim) == 1:
+        dim.append(dX / dim[0])
+
+        if abs(dim[1] - round(dim[1])) >= (2 * dX * np.finfo(float).eps):
+            raise ValueError(
+                "If dim is of size 1, rho must be square and dim[0] must evenly divide length(rho); "
+                "please provide the dim array containing the dimensions of the subsystems."
```
vprusso, line 126: `length` is a MATLAB convention. We should adopt this to be applicable to Python.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+    rho = rho.astype(np.complex128)
+
+    dX = rho.shape[0]
+    sdX = int(round(np.sqrt(dX)))
+
+    # set optional argument defaults: dim=sqrt(length(rho)), tol=sqrt(eps).
+    if dim is None:
+        dim = [sdX, sdX]
+
+    # allow the user to enter a single number for dim.
```
vprusso, line 119:
```suggestion
    # Allow the user to enter a single number for dim.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+    :param tol: `np.sqrt(np.finfo(float).eps)` Convergence tolerance of the iterative Sinkhorn Algorithm.
+                Assumes square root of numpy eps as default.
+    :param max_iterations: Number of iterations after which the solver terminates with a convergence error.
+    :raises: ValueError: if input density matrix is not a square matrix.
+    :raises: ValueError: if the product of dimensions provided/assumed does not match the dimension of density matrix.
+    :raises: RuntimeError: if the sinkhorn algorithm doesnot converge before `max_iterations` iterations.
+    :raises: ValueError: if the density matrix provided is singular (or is not of full rank).
+
+    """
+    # Run checks on the input density matrix
```
vprusso, line 105:
```suggestion
    # Run checks on the input density matrix.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,209 @@
...
+from toqito.channels.partial_trace import partial_trace
+from toqito.matrix_props.is_square import is_square
+
+
+def operator_sinkhorn(
+    rho: np.ndarray, dim: list[int] = None, tol: float = np.sqrt(np.finfo(float).eps), max_iterations: int = 10000
+) -> tuple[np.ndarray, list[np.ndarray]]:
+    r"""Perform operator Sinkhorn iteration (iterative sinkhorn algorithm) on a quantum system's density matrix.
+
+    This function relies on Sinkhorn's theorem :cite:`Sinkhorn_1964_ARelationship` which states "for any
```
vprusso, line 17: Somewhere in this we should be citing QETLAB to indicate that the reference implementation upon which we are basing it is coming from that implementation.

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,179 @@
...
+    rho = random_density_matrix(9)  # 9-dimensional density matrix
+
+    # The dimension `3` divides evenly into `9`, so no error should occur
+    sigma, F = operator_sinkhorn(rho, dim=[3])
+
+    # Check that sigma is a valid density matrix with trace equal to 1
+    np.testing.assert_almost_equal(np.trace(sigma), 1)
+
+
+def test_operator_sinkhorn_max_mixed():
```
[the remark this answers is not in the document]
NV-Karthik: Somehow, I was not able to see your old comments till yesterday. But anyways, all tests which call the function in the similar ways are now parameterized.

NV-Karthik:
@vprusso Regarding your suggestion to change 

`To Perform Operator Sinkhorn on a random 2-Qubit Bipartite state:` to
`To obtain the Sinkhorn operator of a random 2-qubit bipartite state:`

Please correct me if I'm wrong, but this function does not return any sinkhorn operator. What it does is perform the "sinkhorn operator scaling algorithm" to convert the input density matrix into a locally normalized form.

If at all required, only the operators inside the list `F` applied one upon another can be considered close to a sinkhorn operator.

Thats the reason I went with a title consistent to what QETLAB used...
`OPERATORSINKHORN    Performs the operator Sinkhorn iteration`. 

I did fix the capitalization and all, but do let me know if you still prefer to change the title of example.

docs/refs.bib
vprusso: Yes, I would! Thank you!

vprusso:
> @vprusso Regarding your suggestion to change
> 
> `To Perform Operator Sinkhorn on a random 2-Qubit Bipartite state:` to `To obtain the Sinkhorn operator of a random 2-qubit bipartite state:`
> 
> Please correct me if I'm wrong, but this function does not return any sinkhorn operator. What it does is perform the "sinkhorn operator scaling algorithm" to convert the input density matrix into a locally normalized form.
> 
> If at all required, only the operators inside the list `F` applied one upon another can be considered close to a sinkhorn operator.
> 
> Thats the reason I went with a title consistent to what QETLAB used... `OPERATORSINKHORN Performs the operator Sinkhorn iteration`.
> 
> I did fix the capitalization and all, but do let me know if you still prefer to change the title of example.

You are correct, and thank you!

vprusso:
Thanks for the update @NV-Karthik , and feel free to ping when you are ready for a re-review!

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+    :param rho: input density matrix of a multipartite system
+    :param dim: `None` list containing dimensions of each subsystem.
+                Assumes 2 subsystems with equal dimensions as default.
+    :param tol: `np.sqrt(np.finfo(float).eps)` Convergence tolerance of the iterative Sinkhorn Algorithm.
+                Assumes square root of numpy eps as default.
+    :param max_iterations: Number of iterations after which the solver terminates with a convergence error.
+    :raises: ValueError: if input density matrix is not a square matrix.
+    :raises: ValueError: if the product of dimensions provided/assumed does not match the dimension of density matrix.
+    :raises: RuntimeError: if the sinkhorn algorithm doesnot converge before `max_iterations` iterations.
+    :raises: ValueError: if the density matrix provided is singular (or is not of full rank).
```
vprusso, line 83:
```suggestion
    :raises: ValueError: if the density matrix provided is singular (or is not full rank).
    :raises: RuntimeError: if the Sinkhorn algorithm does not converge before :code:`max_iterations` iterations.    
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+    ..jupyter-execute::
+        print(F[0])
+
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
+    :param rho: input density matrix of a multipartite system
+    :param dim: `None` list containing dimensions of each subsystem.
```
vprusso, line 75: The wording of this is a little weird. The default argument is `None`, but it could also be a list as you describe. 
```suggestion
    :param dim: `None` list containing dimensions of each subsystem.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+    F here will be the list of 2 local filtering operators, which are
+
+    ..jupyter-execute::
+        print(F[0])
+
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
```
vprusso, line 73: You need to describe the `:return` in the docstring as well.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+        sigma, F = operator_sinkhorn(rho=rho_random, dim=[2, 2])
+
+        print(sigma)
+
+        print(len(F))
+
+    F here will be the list of 2 local filtering operators, which are
+
+    ..jupyter-execute::
+        print(F[0])
```
vprusso, line 67:
```suggestion
    In other words, :code:`F` contains invertible local operations that demonstrate
    that :code:`rho` and :code:`sigma` are locally equivalent. Here, :code:`F` will be the list of 2 local filtering operators:

    ..jupyter-execute::
        sigma, F = operator_sinkhorn(rho=rho_random, dim=[2, 2])
        print(f"{sigma=}")
        print(f"{F=}")
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+        if abs(dim[1] - round(dim[1])) >= (2 * dX * np.finfo(float).eps):
+            raise ValueError(
+                "If dim is of size 1, rho must be square and dim[0] must evenly divide rho.shape[0]; "
+                "please provide the dim array containing the dimensions of the subsystems."
+            )
+
+        dim[1] = round(dim[1])
+
+    num_sys = len(dim)
+    # Dimensions check before starting iterations
```
vprusso, line 113:
```suggestion
    # Dimensions check before starting iterations.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+    if dim is None:
+        dim = [sdX, sdX]
+
+    # Allow the user to enter a single number for dim.
+    if len(dim) == 1:
+        dim.append(dX / dim[0])
+
+        if abs(dim[1] - round(dim[1])) >= (2 * dX * np.finfo(float).eps):
+            raise ValueError(
+                "If dim is of size 1, rho must be square and dim[0] must evenly divide rho.shape[0]; "
```
vprusso, line 106: This needs to be in an `f`-string, otherwise, these arguments won't render properly.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+                it_err += np.linalg.norm(Prho[j] - Prho_tmp)
+                Prho[j] = Prho_tmp.astype(np.complex128)
+
+                # Apply filter with explicit complex128
+                T_inv = np.linalg.inv(Prho[j]).astype(np.complex128)
+
+                T = sp.linalg.sqrtm(T_inv) / np.sqrt(dim[j])
+                T = T.astype(np.complex128)
+
+                # enforce hermiticity for numerical stability
```
vprusso, line 155:
```suggestion
                # Enforce hermiticity for numerical stability.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+                T = (T + T.conj().T) / 2.0
+
+                eye_ldim = np.eye(int(ldim[j]), dtype=np.complex128)
+                eye_rdim = np.eye(int(rdim[j]), dtype=np.complex128)
```
vprusso, line 159: Do these need to be redefined each time in the loop?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+
+        try:
+            for j in range(num_sys):
+                # Compute the reduced density matrix on the j-th system.
+                Prho_tmp = partial_trace(rho2, list(set(range(num_sys)) - {j}), dim)
+                Prho_tmp = (Prho_tmp + Prho_tmp.conj().T) / 2.0  # for numerical stability
+                it_err += np.linalg.norm(Prho[j] - Prho_tmp)
+                Prho[j] = Prho_tmp.astype(np.complex128)
+
+                # Apply filter with explicit complex128
```
vprusso, line 149:
```suggestion
                # Apply filter with explicit complex128.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+
+    while it_err > tol:
+        if iterations > max_iterations:
+            raise RuntimeError(f"operator_sinkhorn did not converge within {max_iterations} iterations.")
+
+        it_err = 0.0
+        max_cond = 0.0
+
+        error_flag_in_iteration = False
+
```
vprusso, line 140:
```suggestion
        it_err, max_cond, error_flag_in_iteration = 0.0, 0.0, False
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,190 @@
...
+    # handle cases where trace(sigma) can be near zero:
+    # this can happen
+    # 1. if input trace is near zero (singular matrix) - Algorithm will not converge
+    # 2. if there was some numerical instability and result drifted off - Warn users about
+    # about a potentially wrong answer
+    sigma = np.trace(rho) * sigma / np.trace(sigma)
+    if np.trace(sigma) < 10 * np.finfo(float).eps:
+        warnings.warn("Final trace is near zero, but initial trace was not. Result may be unreliable.", RuntimeWarning)
+
+    return sigma, F
```
vprusso, line 190: Is it possible to call the `F` variable something else that is more descriptive?

toqito/matrices/operator_sinkhorn.py
NV-Karthik: Hi @vprusso, my intention here was to convey directly that first element of the list `dim` array should divide first element of `rho.shape`.

so do you want me to put numbers from the input instead of giving a general explanation of error. like...
` "If dim is of size 1, rho must be square and 3 must evenly divide 8; "`

I feel that the current message will cause less confusion to the users. If not edit the message to display like
` "If dim is of size 1, rho must be square and dim[0] (i.e, 3) must evenly divide rho.shape[0] (i.e, 8); "`

toqito/matrices/operator_sinkhorn.py
NV-Karthik: yes. because `ldim[j]` and `rdim[j]` change each time in the loop

vprusso: Wow, I am blind. Indeed, you are correct, sorry!

toqito/matrices/operator_sinkhorn.py
vprusso: Ah, I see what you're trying to do here. Then sure, I agree that your error message (the way you have written it here) is likely more clear/better. Thanks for pointing that out!

toqito/matrices/operator_sinkhorn.py
NV-Karthik: @vprusso I have renamed `F` to `local_ops` (for local operators). Please let me know if its good enough

toqito/matrices/operator_sinkhorn.py
NV-Karthik: @vprusso Edited it now. Please let me know if its good to go.

docs/refs.bib
```diff
@@ -981,6 +996,19 @@ @misc{Sikora_2019_Semidefinite
   url={https://sites.google.com/site/jamiesikora/teaching/sdps-in-quantum-theory}
 }
 
+@article{Sinkhorn_1964_Relationship,
+  author={Richard Sinkhorn},
+  title={{A relationship between arbitrary positive matrices and doubly stochastic matrices}},
```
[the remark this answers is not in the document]
NV-Karthik: Apoologies for this. dont know how it missed my sight!

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,203 @@
...
+        rho_random = random_density_matrix(4, seed=42)
+
+        print(rho_random)
+
+    :code:`operator_sinkhorn` returns the result density matrix along with the operations list :code:`local_ops`.
+    :code:`sigma` (:math:`\sigma`) has all of its (single-party) reduced density matrices proportional
+    to the identity, while satisfying
+
+    .. math::
+        \sigma \, = \, F \, \cdot \, \rho \, \cdot \, F^{\dagger}
```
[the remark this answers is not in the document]
NV-Karthik: I just included them to make it more readable. Removed them now.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,203 @@
...
+    To obtain a locally normalized density matrix of a random 2-qubit bipartite state using operator Sinkhorn:
+
+    ..jupyter-execute::
+        import numpy as np
+        from toqito.matrices import operator_sinkhorn
+        from toqito.rand import random_density_matrix
+
+        rho_random = random_density_matrix(4, seed=42)
+
+        print(rho_random)
```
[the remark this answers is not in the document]
NV-Karthik: Hi @vprusso, I wrote explanations between the code blocks to provide context. should I just move the line where I call the operator sinkhorn to the above code block?  (which gives this)

```
..jupyter-execute::
        import numpy as np
        from toqito.matrices import operator_sinkhorn
        from toqito.rand import random_density_matrix

        rho_random = random_density_matrix(4, seed=42)

        print(rho_random)

        sigma, local_ops = operator_sinkhorn(rho=rho_random, dim=[2, 2])

    :code:`operator_sinkhorn` returns the result density matrix along with the operations list :code:`local_ops`.
    :code:`sigma` (:math:`\sigma`) has all of its (single-party) reduced density matrices proportional
    to the identity, while satisfying

    .. math::
        \sigma = F \cdot \rho \cdot F^{\dagger}

    In other words, :code:`local_ops` contains invertible local operations that demonstrate that :code:`rho` and
    :code:`sigma` are locally equivalent. This can be checked by first obtaining the :math:`F` matrix using the
    elements of :code:`local_ops`

    for this example, :math:`F_1` = :code:`local_ops[0]`, and :math:`F_2` = :code:`local_ops[1]`
    and :math:`F` matrix is then

    .. math::
        F = \left( F_1 \otimes F_2 \right)

    and in general,

    .. math::
        F = \left( F_1 \otimes F_2 \otimes F_3 \otimes ... F_n \right)

    ..jupyter-execute::
        print(sigma)
        print(len(local_ops))

    :code:`local_ops` here will be the list of 2 local filtering operators, first of which is

    ..jupyter-execute::
        print(local_ops[0])
```

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,149 @@
...
+    pt = partial_trace(sigma, target_subsys, dim=dim1)
+    pt_rounded = np.around(pt, decimals=2)
+
+    # Check that partial trace matches the expected identity.
+    np.testing.assert_allclose(pt_rounded, expected_identity, rtol=1e-2)
+
+
+# Test error handling in different scenarios.
+@pytest.mark.parametrize(
+    "rho1, dim1, iters, error_type, expected_msg",
```
vprusso, line 60: Why is it called "rho1" and "dim1" and not just "rho" and "dim"?

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,149 @@
...
+        (
+            random_density_matrix(4, seed=42),
+            [2, 2],
+            20,
+            RuntimeError,
+            "operator_sinkhorn did not converge within 20 iterations.",
+        ),
+    ],
+)
+def test_operator_sinkhorn_errors(rho1, dim1, iters, error_type, expected_msg):
```
vprusso, line 104: Similar comment here for rho1 and dim1 vs. rho and dim

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,204 @@
...
+
+    .. math::
+        F = \left( F_1 \otimes F_2 \right)
+
+    and in general,
+
+    .. math::
+        F = \left( F_1 \otimes F_2 \otimes F_3 \otimes ... F_n \right)
+
+    ..jupyter-execute::
```
vprusso, line 71: I think it would make more sense to have the entire code example in one place instead of splitting it up as you are doing here.

toqito/matrices/operator_sinkhorn.py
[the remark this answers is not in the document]
vprusso: I would suggest putting the code example in one block.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,204 @@
...
+def operator_sinkhorn(
+    rho: np.ndarray, dim: list[int] = None, tol: float = np.sqrt(np.finfo(float).eps), max_iterations: int = 10_000
+) -> tuple[np.ndarray, list[np.ndarray]]:
+    r"""Perform operator Sinkhorn iteration (iterative sinkhorn algorithm) on a quantum system's density matrix.
+
+    This function is adapted from QETLAB. :cite:`QETLAB_link`.
+
+    This function relies on Sinkhorn's theorem :cite:`Sinkhorn_1964_Relationship` which states "for any
+    positive-definite square matrix, there exist diagonal matrices :math:`D_1` and :math:`D_2` such that
+    :math:`D_1 \, \cdot A \, \cdot D_2` is doubly stochastic.
```
vprusso, line 21: Again, I think the "\." are not necessary for formatting.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,204 @@
...
+
+    The iterative Sinkhorn algorithm alternately rescales the input density matrix of the quantum system along each
+    subsystem so that its marginals become uniform (i.e., partial traces or row/column sums become proportional to
+    identity), until convergence.
+
+    Upon convergence we end up with a density matrix `sigma` (:math:`\sigma`), which is locally equivalent to the given
+    multipartite density matrix `rho` (:math:`\rho`), but having maximally mixed subsystems. The algorithm will also
+    return the list of Filtering Operations which can be directly applied on :math:`\rho` to arrive at :math:`\sigma`.
```
vprusso, line 29: Why is "Filtering Operations" capitalized?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,204 @@
...
+    Such converted forms of density matrices are useful to analyse entangled systems and to study effects of operations
+    on each subsystems. (as discussed in :cite:`Gurvits_2004_Classical`)
```
vprusso, line 31:
```suggestion
    on each subsystems (as discussed in :cite:`Gurvits_2004_Classical`).
```

toqito/matrices/operator_sinkhorn.py
NV-Karthik: okay got it.

toqito/matrices/tests/test_operator_sinkhorn.py
NV-Karthik: Now that I think of it, it was just to differentiate when i pass the variables in function. Will change it as suggested.

toqito/matrices/operator_sinkhorn.py
NV-Karthik: sorry, was intending to highlight the term by capitalizing it. now it just looks odd. fixed it

vprusso (approved):
I think this LGTM, but I'll ping @purva-thakre if she has the capacity to review this as well. Thank you, @NV-Karthik ! Great work!

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,199 @@
...
+
+def operator_sinkhorn(
+    rho: np.ndarray, dim: list[int] = None, tol: float = np.sqrt(np.finfo(float).eps), max_iterations: int = 10_000
+) -> tuple[np.ndarray, list[np.ndarray]]:
+    r"""Perform operator Sinkhorn iteration (iterative Sinkhorn algorithm) on a quantum system's density matrix.
+
+    This function is adapted from QETLAB. :cite:`QETLAB_link`.
+
+    This function relies on Sinkhorn's theorem :cite:`Sinkhorn_1964_Relationship` which states "for any
+    positive-definite square matrix, there exist diagonal matrices :math:`D_1` and :math:`D_2` such that
```
purva-thakre, line 20:
```suggestion
    positive-definite square matrix :math:`A`, there exist diagonal matrices :math:`D_1` and :math:`D_2` such that
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,199 @@
...
+        import numpy as np
+        from toqito.matrices import operator_sinkhorn
+        from toqito.rand import random_density_matrix
+
+        rho = random_density_matrix(4, seed=42)
+        print(f"{rho=}")
+
+        sigma, local_ops = operator_sinkhorn(rho=rho_random, dim=[2, 2])
+        print(f"{sigma=}")
+        print(f"{local_ops[0]=}")
```
purva-thakre, line 49:
```suggestion
    ..jupyter-execute::
    
        import numpy as np
        from toqito.matrices import operator_sinkhorn
        from toqito.rand import random_density_matrix

        rho = random_density_matrix(4, seed=42)
        print(f"{rho=}")

        sigma, local_ops = operator_sinkhorn(rho=rho_random, dim=[2, 2])
        print(f"{sigma=}")
        print(f"{local_ops[0]=}")
```
Not working as expected. 
image

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,199 @@
...
+
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
+    :param rho: Input density matrix of a multipartite system.
+    :param dim: List containing dimensions of each subsystem.
+                :code:`None` is passed by default, which assumes two subsystems with equal dimensions.
+    :param tol: `np.sqrt(np.finfo(float).eps)` Convergence tolerance of the iterative Sinkhorn algorithm.
```
purva-thakre, line 83:
```suggestion
    :param tol:  Convergence tolerance of the iterative Sinkhorn algorithm.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,199 @@
...
+    :param max_iterations: Number of iterations after which the solver terminates with a convergence error.
+    :raises: ValueError: If input density matrix is not a square matrix.
+    :raises: ValueError: If the product of dimensions provided/assumed does not match the dimension of density matrix.
+    :raises: ValueError: If the density matrix provided is singular (or is not of full rank).
+    :raises: RuntimeError: If the Sinkhorn algorithm does not converge before :code:`max_iterations` iterations.
+    :return: A tuple of 2 items :code:`(sigma, local_ops)` where,
+
+    - :code:`sigma` is the locally normalized form of the input density matrix :code:`rho`.
+        (:code:`numpy.ndarray` of shape equal to :code:`rho`)
+    - :code:`local_ops` is the list of invertible local operators which can obtain :code:`sigma` from :code:`rho`.
```
purva-thakre, line 94: You can't have a list here. It gets formatted weirdly. 

image

purva-thakre (changes requested):

toqito/matrices/operator_sinkhorn.py
NV-Karthik: @purva-thakre thanks for this, I just learned how to compile and check the docs. This should be fixed now.

toqito/matrices/operator_sinkhorn.py
NV-Karthik: Reformatted the lists to appear under the return section.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+        import numpy as np
+        from toqito.matrices import operator_sinkhorn
+        from toqito.rand import random_density_matrix
+
+        rho = random_density_matrix(4, seed=42)
+        print(f"{rho=}")
+
+        sigma, local_ops = operator_sinkhorn(rho=rho, dim=[2, 2])
+        print(f"{sigma=}")
+        print(f"{local_ops[0]=}")
```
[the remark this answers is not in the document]
purva-thakre: take a look at https://jupyter-sphinx.readthedocs.io/en/latest/

docs/conf.py
```diff
@@ -44,6 +44,8 @@
     "sphinx.ext.napoleon",
     "sphinxcontrib.bibtex",
     "sphinx_copybutton",
+    "jupyter_sphinx",
+    "sphinx.ext.duration",
```
purva-thakre, line 48: rebase your PR. Don't add these as dependencies since they already exist in the main branch.

NV-Karthik: yes just realized. that the changes from jupyter-sphinx PR are now merged. The branch is now rebased.

toqito/matrices/operator_sinkhorn.py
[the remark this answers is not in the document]
NV-Karthik: Now it does appear in the build. I suppose it was an issue with jupyter sphinx not added to the master branch then.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def operator_sinkhorn(
+    rho: np.ndarray, dim: list[int] = None, tol: float = np.sqrt(np.finfo(float).eps), max_iterations: int = 10_000
+) -> tuple[np.ndarray, list[np.ndarray]]:
+    r"""Perform operator Sinkhorn iteration (iterative Sinkhorn algorithm) on a quantum system's density matrix.
+
+    This function is adapted from QETLAB. :cite:`QETLAB_link`.
+
+    This function relies on Sinkhorn's theorem :cite:`Sinkhorn_1964_Relationship` which states "for any
```
purva-thakre, line 19:
```suggestion
    The function relies on Sinkhorn's theorem :cite:`Sinkhorn_1964_Relationship`, which states that "for any
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+    The iterative Sinkhorn algorithm alternately rescales the input density matrix of the quantum system along each
+    subsystem so that its marginals become uniform (i.e., partial traces or row/column sums become proportional to
+    identity), until convergence.
+
+    Upon convergence we end up with a density matrix `sigma` (:math:`\sigma`), which is locally equivalent to the given
+    multipartite density matrix `rho` (:math:`\rho`), but having maximally mixed subsystems. The algorithm will also
+    return the list of filtering operations which can be directly applied on :math:`\rho` to arrive at :math:`\sigma`.
+    Such converted forms of density matrices are useful to analyse entangled systems and to study effects of operations
+    on each subsystems (as discussed in :cite:`Gurvits_2004_Classical`).
+
```
purva-thakre, line 32: Can you add a `See Also` section here and link to the following functions:

- is_positive_definite
- is_stochastic

Look at the other existing functions that have this subsection on how to format the function links.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+    The iterative Sinkhorn algorithm alternately rescales the input density matrix of the quantum system along each
+    subsystem so that its marginals become uniform (i.e., partial traces or row/column sums become proportional to
+    identity), until convergence.
+
+    Upon convergence we end up with a density matrix `sigma` (:math:`\sigma`), which is locally equivalent to the given
+    multipartite density matrix `rho` (:math:`\rho`), but having maximally mixed subsystems. The algorithm will also
+    return the list of filtering operations which can be directly applied on :math:`\rho` to arrive at :math:`\sigma`.
+    Such converted forms of density matrices are useful to analyse entangled systems and to study effects of operations
+    on each subsystems (as discussed in :cite:`Gurvits_2004_Classical`).
```
purva-thakre, line 31:
```suggestion
    Upon convergence we end up with a density matrix :math:`\sigma`, which is locally equivalent to the given
    multipartite density matrix :math:`\rho`, but having maximally mixed subsystems. The algorithm will also
    return the list of filtering operations which can be directly applied on :math:`\rho` to arrive at :math:`\sigma`.
    Such converted forms of density matrices are useful to analyse entangled systems and to study effects of operations
    on each subsystems (as discussed in :cite:`Gurvits_2004_Classical`).
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+    consider the following example:
+
+    .. jupyter-execute::
+
+        import numpy as np
+        from toqito.matrices import operator_sinkhorn
+        from toqito.rand import random_density_matrix
+
+        rho = random_density_matrix(4, seed=42)
+        print(f"{rho=}")
```
purva-thakre, line 46:
```suggestion
        print(f"{Random density matrix =}")
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+
+        sigma, local_ops = operator_sinkhorn(rho=rho, dim=[2, 2])
+        print(f"{sigma=}")
```
purva-thakre, line 49:
```suggestion
        print(f"{Locally normalized form of input matrix =}")
```

toqito/matrices/operator_sinkhorn.py
purva-thakre: Add a new line here such that the output from the function is printed on the next line.

toqito/matrices/operator_sinkhorn.py
purva-thakre: Add a new line here too.

toqito/matrices/operator_sinkhorn.py
purva-thakre, line 50:
```suggestion
        print(f"{Example local invertible operator responsible for returning a locally normalized form of the input matrix =}")
```
Add a new line here, too.

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+    :param rho: Input density matrix of a multipartite system.
+    :param dim: List containing dimensions of each subsystem.
+                :code:`None` is passed by default, which assumes two subsystems with equal dimensions.
+    :param tol: Convergence tolerance of the iterative Sinkhorn algorithm.
+                Assumes square root of numpy eps as default.
+    :param max_iterations: Number of iterations after which the solver terminates with a convergence error.
+    :raises ValueError: If input density matrix is not a square matrix.
+    :raises ValueError: If the product of dimensions provided/assumed does not match the dimension of density matrix.
+    :raises ValueError: If the density matrix provided is singular (or is not of full rank).
+    :raises RuntimeError: If the Sinkhorn algorithm does not converge before :code:`max_iterations` iterations.
```
purva-thakre, line 91:
```suggestion
    :raises RuntimeError: If the Sinkhorn algorithm does not converge before the provided iterations.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+
+    .. bibliography::
+        :filter: docname in docnames
+
+    :param rho: Input density matrix of a multipartite system.
+    :param dim: List containing dimensions of each subsystem.
+                :code:`None` is passed by default, which assumes two subsystems with equal dimensions.
+    :param tol: Convergence tolerance of the iterative Sinkhorn algorithm.
+                Assumes square root of numpy eps as default.
+    :param max_iterations: Number of iterations after which the solver terminates with a convergence error.
```
purva-thakre, line 87:
```suggestion
    :param max_iterations: Number of iterations after which the solver terminates with a convergence error. The default value is :math:`10000`
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+
+    The variable :code:`local_ops` here will be the list of 2 local filtering operators.
+
+    References
+    ==========
+
+    .. bibliography::
+        :filter: docname in docnames
+
+    :param rho: Input density matrix of a multipartite system.
```
purva-thakre, line 82: You say here that you want the density matrix of a multipartite system as an input. However, you don't check if the density matrix provided as input is multipartite in the function. 

What I provide an input $2 \times 2$ density matrix that is not even bipartite? Will this raise some error?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+
+    .. math::
+        F = \left( F_1 \otimes F_2 \right)
+
+    and in general,
+
+    .. math::
+        F = \left( F_1 \otimes F_2 \otimes F_3 \otimes \cdots F_n \right).
+
+    The variable :code:`local_ops` here will be the list of 2 local filtering operators.
```
purva-thakre, line 74: Shouldn't this be before the example?

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+        from toqito.rand import random_density_matrix
+
+        rho = random_density_matrix(4, seed=42)
+        print(f"{rho=}")
+
+        sigma, local_ops = operator_sinkhorn(rho=rho, dim=[2, 2])
+        print(f"{sigma=}")
+        print(f"{local_ops[0]=}")
+
+    :code:`operator_sinkhorn` returns the result density matrix :code:`sigma` along with the operations list
```
purva-thakre, line 52:
```suggestion
    :code:`operator_sinkhorn` returns the result density matrix :math:`sigma` along with the operations list
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+    :code:`local_ops`. :code:`sigma` (:math:`\sigma`) has all of its (single-party) reduced density matrices
```
purva-thakre, line 53:
```suggestion
    :code:`local_ops`. :math:`\sigma` has all of its (single-party) reduced density matrices
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+    proportional to the identity, while satisfying
+
+    .. math::
+        \sigma = F \cdot \rho \cdot F^{\dagger}.
+
+    In other words, :code:`local_ops` contains invertible local operations that demonstrate that :code:`rho` and
```
purva-thakre, line 59:
```suggestion
    In other words, :code:`local_ops` contains invertible local operations that demonstrate that :math:`rho` and
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+    :code:`sigma` are locally equivalent. This can be checked by first obtaining the :math:`F` matrix using the
```
purva-thakre, line 60:
```suggestion
    :math:`sigma` are locally equivalent. This can be checked by first obtaining the :math:`F` matrix using the
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+                Assumes square root of numpy eps as default.
+    :param max_iterations: Number of iterations after which the solver terminates with a convergence error.
+    :raises ValueError: If input density matrix is not a square matrix.
+    :raises ValueError: If the product of dimensions provided/assumed does not match the dimension of density matrix.
+    :raises ValueError: If the density matrix provided is singular (or is not of full rank).
+    :raises RuntimeError: If the Sinkhorn algorithm does not converge before :code:`max_iterations` iterations.
+    :returns:
+        A tuple of 2 items :code:`(sigma, local_ops)` where,
+
+        - :code:`sigma` is the locally normalized form of  :code:`rho`.
```
purva-thakre, line 95:
```suggestion
        - :math:`sigma` is the locally normalized form of  :math:`rho`.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+        - :code:`local_ops` is the list of invertible local operators which can obtain :code:`sigma` from :code:`rho`.
```
purva-thakre, line 96:
```suggestion
        - :code:`local_ops` is the list of invertible local operators which can obtain :math:`sigma` from :math:`rho`.
```

toqito/matrices/operator_sinkhorn.py
```diff
@@ -0,0 +1,202 @@
...
+    # Stabilize the output for numerical stability.
+    sigma = (rho2 + rho2.conj().T) / 2
+    # Handle cases where trace(sigma) can be near zero:
+    # This can happen
+    # 1. if input trace is near zero (singular matrix) - Algorithm will not converge.
+    # 2. if there was some numerical instability and result drifted off - Warn users about
+    # about a potentially wrong answer.
+    sigma = np.trace(rho) * sigma / np.trace(sigma)
+    if np.trace(sigma) < 10 * np.finfo(float).eps:
+        warnings.warn("Final trace is near zero, but initial trace was not. Result may be unreliable.", RuntimeWarning)
```
purva-thakre, line 200: I don't think this `RuntimeWarning` is mentioned in the docstring,

toqito/matrices/tests/test_operator_sinkhorn.py
```diff
@@ -0,0 +1,148 @@
...
+
+def test_operator_sinkhorn_trace_property():
+    """Test the trace property of operator Sinkhorn when a valid inputs are passed."""
+    rho = random_density_matrix(9)
+
+    # The dimension `3` divides evenly into `9`, so no error should occur.
+    sigma, local_ops = operator_sinkhorn(rho, dim=[3])
+
+    # Check that sigma is a valid density matrix with trace equal to 1.
+    np.testing.assert_almost_equal(np.trace(sigma), 1)
```
purva-thakre, line 34: I prefer to not use the np testing module. Could you please use pytest assert for the unit tests in this file?

purva-thakre (approved):

toqito/matrices/operator_sinkhorn.py
NV-Karthik: @purva-thakre. Done for these two. I used the `is_square` function too, do you want me to add it here?

toqito/matrices/operator_sinkhorn.py
NV-Karthik: @purva-thakre Yes, the code doesn't check for `rho` being really multipartite because the algorithm still converges without errors in such cases. Neither does the `QETLAB` function check for this. I repeatedly mentioned multipartite in the docs, to highlight/warn the user that the function expects a multipartite density matrix.

In cases where we pass a 2×2 density matrix which is not bipartite, it will be a Garbage in - Garbage out scenario. The function will not raise any errors. (This is exactly what's done in some tests where random density matrices of matching dimensions are passed).

But there are checks for dimensions (i.e, product of dims should match the size of matrix etc) because that is essential for the algorithm to proceed.

Moreover afaik implementing checks for state being multipartite is complicated enough to require a seperate function. (and I could not find any implementation already. Please let me know if I'm wrong).

toqito/matrices/operator_sinkhorn.py
purva-thakre: I'll leave it up to you.

purva-thakre (approved):

PR #1122, opened 2025-04-02
Update contributing.rst file

### LINKED ISSUE #1119: Update contributing guide to align with new Poetry version
We should update the contributing guide [here](https://toqito.readthedocs.io/en/latest/contributing.html#contrib-guide-reference-label) as some of the approach will be slightly different with newer versions of Poetry.

- We should not tell the user to `pip install poetry`. The new way is to use `pipx`, but we shouldn't write this explicitly on our docs either. Instead, we should note that they need `poetry` and guide them to the [poetry installation page](https://python-poetry.org/docs/#installation) for further instructions on how to configure it on their machine.

- We should get rid of the `(local_venv)` part of the shell snippets, as those are misleading.

- We should update the shell command of how to run tests from 

```sh
(local_venv)~/toqito$ pytest toqito/
```

to

```sh
poetry run pytest
```

- The same is true for this line

```sh
(local_venv)~/toqito/docs$ make clean HTML
```

to

```
toqito/docs$ poetry make clean html
```

Open as well to any other improvements, clarifications, etc. Any external contributor that wants to work on this would be particularly valuable, as I'm likely not the best person to view the project as a newcomer. Anything that is unclear, undocumented, etc. from the perspective of someone coming to this project for this first time would be valuable to include.

Abhishek2634:
## Description

It fixes #1119 

## Checklist

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -24,24 +24,22 @@ We welcome contributions from external contributors, and this document describes
    versions <https://github.com/pyenv/pyenv-virtualenv>`_ or :code:`conda` to create virtual environments with `different Python
    versions <https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#managing-environments>`_.
 
-5. Install :code:`poetry` using the following command. :code:`poetry` is a better dependency resolver than :code:`pip`.
+5. Install :code:`poetry`. Instead of installing Poetry with :code:`pip`, follow the official Poetry installation guide using pipx:
+    Poetry Installation Guide <https://python-poetry.org/docs/#installation>_
 
-.. code-block:: bash
-
-    (local_venv) pip install poetry --upgrade pip
 
 6. Now, navigate to your local clone of the :code:`toqito` repository as shown below.
 
 .. code-block:: bash
 
-    (local_venv) cd toqito
+    cd toqito
 
 7. Use :code:`poetry` as shown below in the :code:`toqito` folder. This should install an editable version of :code:`toqito`
    alongside other development dependencies.
 
 .. code-block:: bash
 
-    (local_venv)~/toqito$ poetry install
+    poetry install
 
 You are now free to make the desired changes in your fork of :code:`toqito`. 
 
@@ -81,7 +79,7 @@ shown below.
 
 .. code-block:: bash
 
-    (local_venv)~/toqito$ pytest toqito/
+    poetry run pytest
 
 The :code:`pytest` module is used for testing and :code:`pytest-cov` can be used to generate
 coverage reports locally. In order to run and :code:`pytest`, you will need to ensure it is installed on your machine
@@ -181,7 +179,7 @@ install :code:`toqito`.
 
 .. code-block:: bash
 
-    (local_venv)~/toqito/docs$ make clean html
+    toqito/docs$ poetry make clean html
 
 If you would prefer to decrease the amount of time taken by :code:`sphinx` to build the documentation locally, use :code:`make html`
 instead.
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -40,7 +40,7 @@ We welcome contributions from external contributors, and this document describes
 
 .. code-block:: bash
 
-    ~/toqito$ poetry install
+    toqito/ $ poetry install
 
 You are now free to make the desired changes in your fork of :code:`toqito`. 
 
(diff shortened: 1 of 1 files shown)


Abhishek2634:
@vprusso  is this fine or need any change ?

docs/contributing.rst
```diff
@@ -24,24 +24,22 @@ We welcome contributions from external contributors, and this document describes
    versions <https://github.com/pyenv/pyenv-virtualenv>`_ or :code:`conda` to create virtual environments with `different Python
    versions <https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#managing-environments>`_.
 
-5. Install :code:`poetry` using the following command. :code:`poetry` is a better dependency resolver than :code:`pip`.
+5. Install :code:`poetry`. Instead of installing Poetry with :code:`pip`, follow the official Poetry installation guide using pipx:
```
vprusso, line 27: How about:

You will require [poetry](https://python-poetry.org/) to manage the dependencies of :code:`toqito`. Please consult the [installation instructions](https://python-poetry.org/docs/#installation) for further details. 

We don't necessarily want to explicitly mention `pip`, `pipx`, etc., as this is all subject to change from the side of poetry.

docs/contributing.rst
```diff
@@ -24,24 +24,22 @@ We welcome contributions from external contributors, and this document describes
...
-.. code-block:: bash
-
-    (local_venv) pip install poetry --upgrade pip
 
 6. Now, navigate to your local clone of the :code:`toqito` repository as shown below.
 
 .. code-block:: bash
 
-    (local_venv) cd toqito
+    cd toqito
```
vprusso, line 35:
```suggestion
    $ cd toqito/
```

docs/contributing.rst
```diff
@@ -24,24 +24,22 @@ We welcome contributions from external contributors, and this document describes
...
 
 7. Use :code:`poetry` as shown below in the :code:`toqito` folder. This should install an editable version of :code:`toqito`
    alongside other development dependencies.
 
 .. code-block:: bash
 
-    (local_venv)~/toqito$ poetry install
+    poetry install
```
vprusso, line 42:
```suggestion
    ~/toqito$ poetry install
```

docs/contributing.rst
```diff
@@ -81,7 +79,7 @@ shown below.
 
 .. code-block:: bash
 
-    (local_venv)~/toqito$ pytest toqito/
+    poetry run pytest
```
vprusso, line 82:
```suggestion
    ~/toqito$ poetry run pytest
```

docs/contributing.rst
```diff
@@ -181,7 +179,7 @@ install :code:`toqito`.
 
 .. code-block:: bash
 
-    (local_venv)~/toqito/docs$ make clean html
+    toqito/docs$ poetry make clean html
```
vprusso, line 182:
```suggestion
    toqito/docs$ poetry run make clean html
```

vprusso (commented):
A few minor changes, but looks overall pretty good! Nice work, @Abhishek2634 !

Abhishek2634:
@vprusso, I am done with all purposed changes.

docs/contributing.rst
```diff
@@ -81,7 +80,7 @@ shown below.
 
 .. code-block:: bash
 
-    (local_venv)~/toqito$ pytest toqito/
+    ~/toqito$ poetry run pytest
```
vprusso, line 83:
```suggestion
    toqito/ $ poetry run pytest
```

docs/contributing.rst
```diff
@@ -24,24 +24,23 @@ We welcome contributions from external contributors, and this document describes
...
-    (local_venv) cd toqito
+    $ cd toqito/
 
 7. Use :code:`poetry` as shown below in the :code:`toqito` folder. This should install an editable version of :code:`toqito`
    alongside other development dependencies.
 
 .. code-block:: bash
 
-    (local_venv)~/toqito$ poetry install
+    ~/toqito$ poetry install
```
vprusso, line 43:
```suggestion
    toqito/ $ poetry install
```

vprusso (approved):
Sorry to be pedantic. I made a few more. Modulo those though, it looks good! I'll approve and let @purva-thakre also take a pass on this as well prior to merging.

docs/contributing.rst
purva-thakre: Why do you need to specify `/`? 

I have often used `cd toqito` without any issues.

vprusso: Yes, you can do this, however the full path (if you say, use "tab" to complete the name) will add the `/` at the end. It also just makes it more explicit that it's a directory as well. So, yes, while it will work, I think the slash makes it more explicit.

PR #1131, opened 2025-04-04
fix: Changed dependencies from >= to ^= (#1001)

ryanbijoy:
## Description

Ref - #1001 

## Changes

Changed to ^= in:
  -  [x] [tool.poetry.dependencies]
  - [x] [tool.poetry.group.dev.dependencies]
  - [x] [tool.poetry.group.lint.dependencies]
  - [x] [tool.poetry.group.docs.dependencies]

## Checklist


ryanbijoy:
@vprusso Can you check and run this.

ryanbijoy:
@vprusso `^=` is a incorrect syntax, I think. I think it should be `^` or `>=`

vprusso:
@ryanbijoy just a heads-up that you'll likely need to run `poetry lock` to update the lock file for the build failures.

ryanbijoy:
> @ryanbijoy just a heads-up that you'll likely need to run `poetry lock` to update the lock file for the build failures.

Yaa just was doing that and submitting. Also i mean it Downgrading pre-commit and setuptools so its fine right?

Screenshot 2025-04-05 at 12 39 47 AM

vprusso:
> > @ryanbijoy just a heads-up that you'll likely need to run `poetry lock` to update the lock file for the build failures.
> 
> Yaa just was doing that and submitting. Also i mean it Downgrading pre-commit and setuptools so its fine right?
> 
> Screenshot 2025-04-05 at 12 39 47 AM

Hmm, not sure I understand. Why would `pre-commit` and `setuptools` be downgraded? @ryanbijoy

ryanbijoy:
> > > @ryanbijoy just a heads-up that you'll likely need to run `poetry lock` to update the lock file for the build failures.
> > 
> > 
> > Yaa just was doing that and submitting. Also i mean it Downgrading pre-commit and setuptools so its fine right?
> > Screenshot 2025-04-05 at 12 39 47 AM
> 
> Hmm, not sure I understand. Why would `pre-commit` and `setuptools` be downgraded? @ryanbijoy

Actually i just saw that ^3.0.0 allows versions from 3.0.0 up to but not including 4.0.0. similarly Allow all versions from 65.5.1 up to (but not including) 66.0.0

Need to make some changes...

vprusso:
> Also `setuptools` got updated from `version = "77.0.3"` to `version = "78.1.0"`. Hope thats fine?

Yes, I think updating packages is actually a **good** thing, so yes, that seems fine to me.

ryanbijoy:
> > Also `setuptools` got updated from `version = "77.0.3"` to `version = "78.1.0"`. Hope thats fine?
> 
> Yes, I think updating packages is actually a **good** thing, so yes, that seems fine to me.

In that case do you want me to 'poetry update' which will update all the dependencies? or its fine as it is.

vprusso:
Yes, I think we are going to want to have the updated packages as defined in the toml file

ryanbijoy:
Can you re run the checks, as quantiki is black online

vprusso:
Looks like it passed, but I'll still hold off on merging until @purva-thakre can take a look as I know she has had some input on resolving dependency issues like this in the past. Thank you, @ryanbijoy !

ryanbijoy:
@vprusso Np problem, Happy to do it :)

pyproject.toml
```diff
@@ -32,39 +32,39 @@ repository = "https://github.com/vprusso/toqito"
...
-sphinx-autoapi = "3.6.0"
-sphinx-copybutton = "0.5.2"
+sphinx = "^8.1.3"
+furo = "^2024.8.6"
+sphinxcontrib-bibtex = "^2.6.3"
+sphinx-autoapi = "^3.6.0"
+sphinx-copybutton = "^0.5.2"
 
 [build-system]
 requires = ["poetry-core>=2.0.0,<3.0.0"]
```
purva-thakre, line 70: I am not too sure of this dependency syntax. We are using `poetry>=2.0` 

https://python-poetry.org/docs/dependency-specification/#projectdependencies-and-toolpoetrydependencies

`^` is used for prior to `2.0`.

vprusso: The `^` symbol is still supported in Poetry 2.0+ according to the same link you provided.

purva-thakre: Maybe I am not looking at the right section?

image

I was thinking of switching to `project.dependencies` as we are using `> 2.0`.

vprusso: Yes, but this isn't saying anything about `^` vs `>=`. It's more about `dependencies` vs. `[tool.poetry.dependencies]`. If you ctrl-f for other instance of `^`, you will see that the `^` delimitter is still supported and used in Poetry > 2.0

Screenshot 2025-04-05 at 4 20 09 PM

purva-thakre: The problem is if we decide to switch to `project.dependencies`, we can't use caret syntax. 

image

vprusso: @purva-thakre  Oh, this is a good call, and somehow I completely missed that. 

Okay, we definitely don't want to go adding things here that would not be compatible with Poetry 2.0+. 

@ryanbijoy what do you think about using the more modern `project.dependencies` way of specifying packages as described on the official Poetry docs? I think this would be the better alternative so we stay compatible with new versions of Poetry. We also I think want to move to `uv` at some point, so doing this will also allow us to likely make that shift in an easier way.

vprusso: Following up on this thread, @ryanbijoy does using the more modern `project.dependencies` seem like a sensible approach for you to implement? LMK if that seems reasonable to you!

purva-thakre:
@ryanbijoy Are you planning to work on resolving the suggestions in this PR?

Issue #1142, opened 2025-04-06
Bug: `is_stochastic` fails to return `False` on certain non-doubly stochastic matrices


AbdullahKazi500:
## Before reporting this bug:
- [x] Verify there is not an existing issue for it already
- [x] You are using the latest version of `toqito` with a python version `>=3.10`

## Describe the bug
There is a critical logic error in the is_stochastic function from the toqito.matrix_props module. The function incorrectly returns after checking only the first axis, regardless of the stochastic matrix type being checked. This causes the function to fail when validating doubly stochastic matrices, as it doesn't properly check both axes.
This block will only iterate once and return True or False after checking just the first axis in axis_num. That breaks the logic for checking "doubly" stochastic matrices, which are expected to satisfy both:

Sum of columns (axis=0) == 1

Sum of rows (axis=1) == 1
For a "doubly" stochastic matrix, axis_num will be [0, 1], but due to the return inside the loop, it will exit after the first check. This means:

-  If the matrix passes the first axis, it returns True without checking the second.

- If it fails the first axis, it returns False without checking the second.

So it's not doing what "doubly" should be doing: verifying both conditions hold.
The return False statement is indented at the same level as the if statement inside the loop, which means it will execute after the first iteration regardless of the outcome of the if condition.
For "doubly" stochastic matrices, axis_num is [0, 1], meaning both row sums and column sums need to be checked, the function will return after checking only the first axis (either rows or columns), without checking the second axis.
If the first axis check passes, it returns True immediately without verifying the second axis. If the first check fails, it returns False immediately without giving the second axis a chance.
 the floating-point comparison using ==1.0 is prone to precision errors. Using np.isclose() would be more robust.

## To Reproduce
```
import numpy as np
from toqito.matrices import pauli
from toqito.matrix_props import is_stochastic

# Create a matrix that is both left and right stochastic
test_matrix = np.array([[0.5, 0.5], [0.5, 0.5]])

# This should return True but may return False due to the bug
result = is_stochastic(test_matrix, "doubly")
print(result)  # May incorrectly print False

# The issue is in this code segment:
# for i in axis_num:
#     if np.all(np.sum(mat, i)==1.0):
#         return True
#     return False  # This line is improperly indented and causes early return
```
## Expected behavior
[...]

PR #1144, opened 2025-04-06
Updated is_stochastic.py

AbdullahKazi500:
Fixed

## Description
Fixed the indentation in the code as per the discussion in #1142  

## Changes
Changed the code to fix indentation 

  -  [x]  The return False statement is at the same indentation level as the if statement inside the loop, causing the function to exit after checking only the first axis.

## Checklist


vprusso (commented):
But it looks like this `return` statement is outside the scope of the function. I'm not sure this is the correct fix.

vprusso:
But the return is outside the scope of the function. The change you provided doesn't impose the fix you described and in fact will fail to run as well @AbdullahKazi500

PR #1154, opened 2025-04-08
fix: enforce strict positivity in is_totally_positive and update tests

### LINKED ISSUE #1098: `is_totally_positive` incorrectly identifies matrices
The function `is_totally_positive` currently considers the following matrix:

```py
[[1, 2], 
 [2, 4]]
```

as totally positive. That is,

```py
>>> from toqito.matrix_props import is_totally_positive
>>> is_totally_positive([[1, 2], [2, 4]])
True
```

However, this matrix is *not* totally positive. This can be checked my considering all 1×1 minors (individual entries):

A[1,1] = 1 > 0 ✓
A[1,2] = 2 > 0 ✓
A[2,1] = 2 > 0 ✓
A[2,2] = 4 > 0 ✓

The 2×2 minor (determinant of the full matrix):

det(A) = (1 × 4) - (2 × 2) = 4 - 4 = 0

Since the determinant of the full matrix equals zero, this matrix is **not** totally positive. A totally positive matrix requires all minors to be strictly positive, but here we have a zero determinant.

mdciofalo:
## Description
Modifies `is_totally_positive.py` and its corresponding test file to ensure strict positivity of all minors.  
Closes #1098.

## Changes
  - Adjusts test case for identity matrix (identity matrix has minors equal to 0, so it's not totally positive)
  - Adds new test case: 2x2 matrix with zero determinant
  - Updates logic in `is_totally_positive.py` (line 78) to enforce strict positivity 

## Checklist

### DIFF
--- a/toqito/matrix_props/is_totally_positive.py
+++ b/toqito/matrix_props/is_totally_positive.py
@@ -75,6 +75,6 @@ def is_totally_positive(mat: np.ndarray, tol: float = 1e-6, sub_sizes: list | No
             for kr in sub_ind_r:
                 for kc in sub_ind_c:
                     d = np.linalg.det(mat[np.ix_(kr, kc)])
-                    if d < -tol or abs(np.imag(d)) > tol:
+                    if d < tol or abs(np.imag(d)) > tol:
                         return False
     return True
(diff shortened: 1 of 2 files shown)


mdciofalo:
I noticed a GPG signature warning from Codecov, but all tests and coverage ran successfully otherwise. Since this seems to be unrelated to the PR, I believe this is ready for review :)

PR #1151, opened 2025-04-07
Fixed fourier.py

AbdullahKazi500:
## Description
This PR improves the docstring formatting of the `fourier` function for better readability and accuracy in rendered documentation. It ensures the mathematical expressions for the Fourier matrix are consistent and correctly typeset, without modifying the core logic of the function.
Fixes #1147 

## Changes
- [x] Corrected the matrix formula rendering in the LaTeX-style math block for `W_N`.
  - [x] Updated the example Fourier matrix `W_3` to reflect consistent notation and fix indexing inaccuracies.

## Checklist
Before marking your PR ready for review, make sure you checked the following locally. If this is your first PR, you might be notified of some workflow failures after a maintainer has approved the workflow jobs to be run on your PR. 

Additional information is available in the [documentation](https://toqito.readthedocs.io/en/latest/contributing.html#testing).

  -  [x] Use `ruff` for errors related to code style and formatting.
  -  [x] Verify all previous and newly added unit tests pass in `pytest`.
  -  [x] Check the documentation build does not lead to any failures. `Sphinx` build can be checked locally for any failures related to your PR
  -  [x] Use `linkcheck` to check for broken links in the documentation
  -  [x] Use `doctest` to verify the examples in the function docstrings work as expected.

PR #1155, opened 2025-04-08
fix: fixing ruff lint style in fourier

vprusso:
Looks like a few minor lint errors snuck in from https://github.com/vprusso/toqito/pull/1151


vprusso:
> @vprusso what exactly causes this linting error in the code ,I have seen a lot of memes on linter even with people cleaning the code thoroughly this lint error pops up

Linters are tools that automatically run on the code to ensure that certain stylistic guidelines are followed. If any of them are violated, the pipeline fails. The `ruff` tool is configured to work with `toqito`, so you can run something like:

```
poetry run ruff check . --fix   
```

For it to automatically catch and fix any linter issues.

Issue #1166, opened 2025-04-10
Warnings: Resolve complex casting warnings for tests in the `rand/` module


vprusso:
Running `poetry run pytest toqito/rand/tests` yields the following set of warnings which should be addressed:

```
toqito/rand/tests/test_random_density_matrix.py: 2 warnings
toqito/rand/tests/test_random_ginibre.py: 2 warnings
toqito/rand/tests/test_random_orthonormal_basis.py: 1 warning
toqito/rand/tests/test_random_povm.py: 1 warning
toqito/rand/tests/test_random_state_vector.py: 3 warnings
toqito/rand/tests/test_random_states.py: 4 warnings
toqito/rand/tests/test_random_unitary.py: 1 warning
  /Users/vrusso/Projects/research/toqito/.venv/lib/python3.12/site-packages/numpy/ma/testutils.py:80: ComplexWarning: Casting complex values to real discards the imaginary part
    ).astype(np.float64)
```

PR #1167, opened 2025-04-10
fix: update np testing for rand functions.

vprusso:
Closes: #1166 

The `assert_array_almost_equal` function is internally written in such a way that it converts its inputs to a real type before the comparison. 

When the output density matrix is complex (even if some entries have zero imaginary parts) and it is compared to an expected complex (or sometimes a real) array, the internal conversion may discard the imaginary part. This triggers the warning about “casting complex values to real.”


purva-thakre:
Will review this by Friday evening.

purva-thakre (approved):

PR #1185, opened 2025-04-16
feat: enhance measurement function

### LINKED ISSUE #1184: Enhance: Measurement function should allow the retrieval of the post-measurement state
Currently, the [`measure`](https://github.com/vprusso/toqito/blob/master/toqito/measurement_ops/measure.py) function takes in a measurement and a state and provides the probability of obtaining a measurement outcome applied to state. We should enhance this function to:

- Take as input either a POVM or a set of Kraus operators
- Provide the probability and also the post-measurement state (if the user wishes to obtain this).

Below is a candidate implementation:

```py
import numpy as np


def measure(state, measurement, tol=1e-10, update=False):
    r"""Applies a measurement to a quantum state.

    The measurement can be provided as a single operator (POVM element or Kraus operator) or as a
    list of operators (assumed to be Kraus operators) describing a complete quantum measurement.

    When a single operator is provided:
      - Returns the measurement outcome probability if `update` is False.
      - Returns a tuple (probability, post_state) if `update` is True.

    When a list of operators is provided, the function verifies that they satisfy the completeness relation
    
    .. math::
       \sum_i K_i^\dagger K_i = I,
    
    when `update` is True. Then, for each operator :math:`K_i`, the outcome probability is computed as
    
    .. math::
       p_i = \mathrm{Tr}\Bigl(K_i^\dagger K_i\, \rho\Bigr),
    
    and, if :math:`p_i > tol`, the post‐measurement state is updated via
    
    .. math::
       \rho_i = \frac{K_i\, \rho\, K_i^\dagger}{p_i}.
    
    If :math:`p_i \le tol`, the corresponding post‐measurement state is a zero matrix.

    Examples
    =========
    **Single operator (POVM element):**

    .. jupyter-execute::
        import numpy as np
        rho = np.array([[0.5, 0.5], [0.5, 0.5]])
        proj_0 = np.array([[1, 0], [0, 0]])

        # Without update; simply returns the probability.
        print(measure(rho, proj_0))

        # With state update; returns (probability, post_state).
        p, post_state = measure(rho, proj_0, update=True)
        print(p)

    **Multiple operators (Kraus operators):**

    .. jupyter-execute::
        rho = np.array([[0.5, 0.5], [0.5, 0.5]])
        K0 = np.array([[1, 0], [0, 0]])
        K1 = np.array([[0, 0], [0, 1]])
    
        # Returns list of probabilities.
        print(measure(rho, [K0, K1]))

        # Returns list of (probability, post_state) tuples.
        print(measure(rho, [K0, K1], update=True))

    :param state: Quantum state as a density matrix (an np.ndarray with shape (d, d)).
    :param measurement: Either a single measurement operator (an np.ndarray) or a list/tuple of operators.
                        When providing a list, they are assumed to be Kraus operators satisfying the completeness
                        relation.
    :param tol: Tolerance for numerical precision (default is 1e-10).
    :param update: If True, also return the post-measurement state(s); otherwise, only the probability or
                   probabilities are returned.
    :raises ValueError: If a list of operators does not satisfy the completeness relation.
    :return: If a single operator is provided, returns a float (probability) or a tuple (probability, post_state)
             if `update` is True. If a list is provided, returns a list of probabilities or a list of tuples if
             `update` is True.
    """
    if isinstance(measurement, (list, tuple)):
        if update:
            d = state.shape[0]
            completeness = sum(op.T.conj() @ op for op in measurement)
            if not np.allclose(completeness, np.eye(d), atol=tol):
                raise ValueError("Kraus operators do not satisfy completeness relation: ∑ Kᵢ†Kᵢ ≠ I.")
        outcomes = []
        for op in measurement:
            result = op @ state @ op.conj().T
            prob = np.trace(result).real
            if prob > tol:
                post_state = result / prob
            else:
                post_state = np.zeros_like(state)
            outcomes.append((prob, post_state) if update else prob)
        return outcomes
    else:
        result = measurement @ state @ measurement.conj().T
        prob = np.trace(result).real
        if prob > tol:
            post_state = result / prob
        else:
            post_state = np.zeros_like(state)
        return (prob, post_state) if update else prob
```

Any current usages of `measure` would need to be updated accordingly. Additionally, the unit test suite should be updated to cover these additional branches of functionality.

vprusso:
Closes #1184

### DIFF
--- a/toqito/measurement_ops/measure.py
+++ b/toqito/measurement_ops/measure.py
@@ -19,11 +19,10 @@ def measure(
       - Returns the measurement outcome probability if ``state_update`` is False.
       - Returns a tuple (probability, post_state) if ``state_update`` is True.
 
-    When a list of operators is provided, the function verifies that they satisfy the completeness relation in  Eq.
-    :ref:`completeness_rel_equation` when ``state_update`` is True.
+    When a list of operators is provided, the function verifies that they satisfy the completeness relation when
+    ``state_update`` is True.
 
     .. math::
-      :label: completeness_rel_equation
        \sum_i K_i^\dagger K_i = \mathbb{I},
 
     when `state_update` is True. Then, for each operator :math:`K_i`, the outcome probability is computed as
--- a/toqito/measurement_ops/measure.py
+++ b/toqito/measurement_ops/measure.py
@@ -1,6 +1,7 @@
 """Apply measurement to a quantum state."""
 
 import numpy as np
+from toqito.matrix_props import is_density
 
 
 def measure(
@@ -84,7 +85,10 @@ def measure(
              ``state_update`` is True.
 
     """
-    # Single-operator case
+    if not is_density(state):
+        raise ValueError("Input must be a valid density matrix.")
+
+    # Single-operator case.
     if not isinstance(measurement, (list, tuple)):
         result = measurement @ state @ measurement.conj().T
         prob = np.trace(result).real
@@ -94,7 +98,7 @@ def measure(
             post_state = np.zeros_like(state)
         return (prob, post_state) if state_update else prob
 
-    # List-of-operators case
+    # List-of-operators case.
     outcomes: list[float | tuple[float, np.ndarray]] = []
     probs: list[float] = []
 
@@ -110,7 +114,7 @@ def measure(
 
         outcomes.append((prob, post_state) if state_update else prob)
 
-    # Only enforce completeness if we're doing the update AND every outcome was nonzero
+    # Only enforce completeness if we're doing the update AND every outcome was nonzero.
     if state_update and all(p > tol for p in probs):
         d = state.shape[0]
         completeness = sum(op.T.conj() @ op for op in measurement)
--- a/toqito/measurement_ops/measure.py
+++ b/toqito/measurement_ops/measure.py
@@ -70,7 +70,7 @@ def measure(
         # Returns list of (probability, post_state) tuples.
         print(measure(rho, [K0, K1], update=True))
 
-    :param state: Quantum state as a density matrix (an np.ndarray with shape (d, d)).
+    :param state: Quantum state as a density matrix shape (d, d) where d is......
     :param measurement: Either a single measurement operator (an np.ndarray) or a list/tuple of operators.
                         When providing a list, they are assumed to be Kraus operators satisfying the completeness
                         relation.
--- a/toqito/measurement_ops/measure.py
+++ b/toqito/measurement_ops/measure.py
@@ -18,7 +18,7 @@ def measure(
       - Returns the measurement outcome probability if ``update`` is False.
       - Returns a tuple (probability, post_state) if ``update`` is True.
 
-    When a list of operators is provided, the function verifies that they satisfy the completeness relation
+    When a list of operators is provided, the function verifies that they satisfy the completeness relation in  Eq. :ref:`completeness_rel_equation` when ``update`` is True. 
 
     .. math::
       :label: completeness_rel_equation
--- a/toqito/measurement_ops/measure.py
+++ b/toqito/measurement_ops/measure.py
@@ -15,7 +15,7 @@ def measure(
     list of operators (assumed to be Kraus operators) describing a complete quantum measurement.
 
     When a single operator is provided:
-      - Returns the measurement outcome probability if `update` is False.
+      - Returns the measurement outcome probability if ``update`` is False.
       - Returns a tuple (probability, post_state) if `update` is True.
 
     When a list of operators is provided, the function verifies that they satisfy the completeness relation
--- a/toqito/measurement_ops/measure.py
+++ b/toqito/measurement_ops/measure.py
@@ -16,7 +16,7 @@ def measure(
 
     When a single operator is provided:
       - Returns the measurement outcome probability if ``update`` is False.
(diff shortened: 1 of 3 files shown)


toqito/measurement_ops/measure.py
```diff
@@ -1,76 +1,118 @@
...
+        print(measure(rho, [K0, K1]))
 
+        # Returns list of (probability, post_state) tuples.
+        print(measure(rho, [K0, K1], update=True))
 
-    :param measurement: The measurement to apply.
-    :param state: The state to apply the measurement to.
-    :return: Returns the probability of obtaining a given outcome after applying
-             the variable :code:`measurement` to the variable :code:`state`.
+    :param state: Quantum state as a density matrix (an np.ndarray with shape (d, d)).
```
purva-thakre, line 72:
```suggestion
    :param state: Quantum state as a density matrix shape (d, d) where d is......
```
No need to add type here but you should describe what d is.

toqito/measurement_ops/measure.py
```diff
@@ -1,76 +1,118 @@
...
+    :param measurement: Either a single measurement operator (an np.ndarray) or a list/tuple of operators.
+                        When providing a list, they are assumed to be Kraus operators satisfying the completeness
+                        relation.
+    :param tol: Tolerance for numerical precision (default is 1e-10).
+    :param update: If True, also return the post-measurement state(s); otherwise, only the probability or
+                   probabilities are returned.
+    :raises ValueError: If a list of operators does not satisfy the completeness relation.
+    :return: If a single operator is provided, returns a float (probability) or a tuple (probability, post_state)
+             if `update` is True. If a list is provided, returns a list of probabilities or a list of tuples if
```
purva-thakre, line 81:
```suggestion
             if ``update`` is True. If a list is provided, returns a list of probabilities or a list of tuples if
```

toqito/measurement_ops/measure.py
```diff
@@ -1,76 +1,118 @@
...
+             `update` is True.
```
purva-thakre, line 82:
```suggestion
             ``update`` is True.
```

toqito/measurement_ops/measure.py
```diff
@@ -1,76 +1,118 @@
...
+            post_state = result / prob
+        else:
+            post_state = np.zeros_like(state)
+
+        outcomes.append((prob, post_state) if update else prob)
+
+    # Only enforce completeness if we're doing the update AND every outcome was nonzero
+    if update and all(p > tol for p in probs):
+        d = state.shape[0]
+        completeness = sum(op.T.conj() @ op for op in measurement)
```
purva-thakre, line 114: I think the completeness relation should be verified before L99. 

Maybe I am missing some condition as to why you check this only if L112 is satisfied?

purva-thakre (changes requested):

toqito/measurement_ops/measure.py
```diff
@@ -1,76 +1,118 @@
...
 import numpy as np
 
 
-def measure(measurement: np.ndarray, state: np.ndarray) -> float:
-    r"""Determine probability of obtaining a measurement outcome applied to state.
+def measure(
+    state: np.ndarray,
+    measurement: np.ndarray | list[np.ndarray] | tuple[np.ndarray, ...],
+    tol: float = 1e-10,
+    update: bool = False,
```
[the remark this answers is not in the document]
vprusso: Yes, I agree. Let's go with `state_update`. Done in ada74f5ecf652876e5950fb757cef1da84c65288

toqito/measurement_ops/measure.py
```diff
@@ -1,76 +1,118 @@
...
+    :param tol: Tolerance for numerical precision (default is 1e-10).
+    :param update: If True, also return the post-measurement state(s); otherwise, only the probability or
+                   probabilities are returned.
+    :raises ValueError: If a list of operators does not satisfy the completeness relation.
+    :return: If a single operator is provided, returns a float (probability) or a tuple (probability, post_state)
+             if `update` is True. If a list is provided, returns a list of probabilities or a list of tuples if
+             `update` is True.
 
     """
-    return float(np.trace(measurement.conj().T @ state))
```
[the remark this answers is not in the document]
vprusso: Good point, added this check! dee8e6c47a6929e19af9d2a58f26a9bd64396f3c

toqito/measurement_ops/measure.py
vprusso: It's a performance compromise. This defers the completeness check until after measurement application, and only checks it if all outcome probabilities are nonzero.

toqito/measurement_ops/measure.py
```diff
@@ -1,76 +1,123 @@
...
-    that satisfies
+    When a list of operators is provided, the function verifies that they satisfy the completeness relation when
+    ``state_update`` is True.
 
     .. math::
-        \sum_{a \in \Sigma} \mu(a) = \mathbb{I}_{\mathcal{X}}.
+       \sum_i K_i^\dagger K_i = \mathbb{I},
 
-    Further information can be found here :cite:`WikiQuantMeas`.
+    when `state_update` is True. Then, for each operator :math:`K_i`, the outcome probability is computed as
```
purva-thakre, line 28:
```suggestion
    when ``state_update`` is True. Then, for each operator :math:`K_i`, the outcome probability is computed as
```

purva-thakre (approved):

PR #1186, opened 2025-04-18
Added .mypy file

RohitP2005:
## Description
This PR is a initial step to work on adding mypy as stated in the issue: #299

## Changes
Have added .ini.mypy file and added mypy in the worfllows but have commented out them 

## Checklist

### DIFF
--- a/.github/workflows/build-test-actions.yml
+++ b/.github/workflows/build-test-actions.yml
@@ -39,6 +39,8 @@ jobs:
           python -m pip install --upgrade pip
           pip install poetry
           poetry install --without docs --no-interaction
+     #- name: Run mypy
+      #  run: poetry run mypy toqito/ --explicit-package-bases
 
 
   toqito-pytest:
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -4,3 +4,8 @@ ignore_missing_imports = True
 [mypy-toqito.*]
 ignore_missing_imports = True
 
+[mypy]
+explicit_package_bases = True
+python_version = 3.10
+ignore_missing_imports = True
+
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,0 +1,6 @@
+[mypy-scipy.*]
+ignore_missing_imports = True
+
+[mypy-toqito.*]
+ignore_missing_imports = True
+
--- a/toqito/channel_metrics/completely_bounded_spectral_norm.py
+++ b/toqito/channel_metrics/completely_bounded_spectral_norm.py
@@ -2,8 +2,8 @@
 
 import numpy as np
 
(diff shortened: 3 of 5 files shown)


RohitP2005:
@purva-thakre from #299 , i have handled the `duplicate module` and created `.mypy.ini` to ignore the scipy error.
There still a lot of work to be done in the code base before implementing mypy in the CI.

It would be helpful if u guide me on what to do next !

.github/workflows/build-test-actions.yml
```diff
@@ -39,6 +39,8 @@ jobs:
           python -m pip install --upgrade pip
           pip install poetry
           poetry install --without docs --no-interaction
+     #- name: Run mypy
```
vprusso, line 42: Is there a reason this is commented out?

RohitP2005: As I mentioned earlier , the workflow can be uncommented, once the existing errors can be resolved

purva-thakre:
> It would be helpful if u guide me on what to do next !

Not sure what you want to say here? Mypy failres should provide enough information on what you should do next...

RohitP2005:
> > It would be helpful if u guide me on what to do next !
> 
> Not sure what you want me to say here? Mypy failures should provide enough information on what you should do next...

Should I fix the errors in Mypy failures, If yes it requires a lot  of rework to be done in the code base which cannot be tracked in a single PR

vprusso:
@RohitP2005 

I would say that part of this issue would be indeed hooking up the automatic mypy checking but also fixing any of the mypy failures. Perhaps one could go about this in the reverse way. That is, run mypy, determine the items that need to be fixed, fix those in separate MRs, then once merged, hook in the auto mypy in the pipeline so we don't incur mypy failures in the interim. 

Would that seem reasonable to you, @RohitP2005 / @purva-thakre ?

vprusso:
> `Found 477 errors in 115 files (checked 397 source files)` this was my output when i ran `mypy toqito/ --explicit-package-bases` locally

Just reiterating my approach, as it does indeed seem like there are quite a few issues to be resolved. Likely, many of these are small changes, but there might be a few that are tricky to completely resolve.

RohitP2005:
> > `Found 477 errors in 115 files (checked 397 source files)` this was my output when i ran `mypy toqito/ --explicit-package-bases` locally
> 
> Just reiterating my approach, as it does indeed seem like there are quite a few issues to be resolved. Likely, many of these are small changes, but there might be a few that are tricky to completely resolve.

Thanks, @purva-thakre, for the help!
I understand it better now.
I'll start looking into fixing the mypy errors file by file.

Issue #1083, opened 2025-03-26
Add Github Actions to check and update poetry.lock file


NV-Karthik:
Lack of github action to update outdated `poetry.lock` files can cause automated checks to fail.

`poetry.lock` file becomes outdated when the dependencies or their versions in pyproject.toml file change, and if it is not updated accordingly.

### Fix

Adding a Github action (job step) to

1. Check if the `poetry.lock` file is outdated using the bash command `poetry check --lock`
2. And if the file is outdated, run the bash command `poetry lock --no-interaction` before running the `Install toqito + dependencies` job step
3. If the file is up-to-date, continue to `Install toqito + dependencies` as it was before

### More Info

- Committing `poetry.lock` file to version control

While it is a workaround for now. Official documentation discourages doing this for library developers.

> Library developers have more to consider. Your users are application developers, and your library will run in a Python environment you don’t control.

> The application ignores your library’s lock file. It can use whatever dependency version meets the constraints in your pyproject.toml. The application will probably use the latest compatible dependency version. If your library’s poetry.lock falls behind some new dependency version that breaks things for your users, you’re likely to be the last to find out about it.

> If you do not want to give up the reproducibility and performance benefits, consider a regular refresh of poetry.lock to stay up-to-date and reduce the risk of sudden breakage for users.

https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control

### Other Possible Fixes

- Adding a github action to delete the `poetry.lock` file before every installation

Doing this will resolve dependencies on every install using `pyproject.toml` only. And creates another lock file on every install. More info here  (Thanks @purva-thakre)

https://python-poetry.org/docs/basic-usage/#installing-without-poetrylock

RohitP2005:
hey @purva-thakre @NV-Karthik  is this being worked on 
I would like to address this if this not

purva-thakre:
If you look at the top right section of the screen, you can see it is assigned to someone. When an issue is assigned, the expectation is is they are working on it.

RohitP2005:
I am aware of that .
I am up for working on it if its being unattended 😅

purva-thakre:
I'd suggest you focus on one issue at a time.

RohitP2005:
Yeah sure , I adhere to it .

purva-thakre:
Closed by #1198

purva-thakre:
[...]

PR #1190, opened 2025-04-26
Added job to check and update lock file if it is outdated

NV-Karthik:
## Changes

  -  [ ] Change 1

## Checklist
Before marking your PR ready for review, make sure you checked the following locally. If this is your first PR, you might be notified of some workflow failures after a maintainer has approved the workflow jobs to be run on your PR. 

Additional information is available in the [documentation](https://toqito.readthedocs.io/en/latest/contributing.html#testing).

  -  [ ] Use `ruff` for errors related to code style and formatting.
  -  [ ] Verify all previous and newly added unit tests pass in `pytest`.
  -  [ ] Check the documentation build does not lead to any failures. `Sphinx` build can be checked locally for any failures related to your PR
  -  [ ] Use `linkcheck` to check for broken links in the documentation
  -  [ ] Use `doctest` to verify the examples in the function docstrings work as expected.

PR #1198, opened 2025-05-05
fix: Github action to check and auto-commit poetry.lock (new PR)

NV-Karthik:
## Description
Creates a job in the build-test-actions.yml to check for outdated poetry.lock files and Autocommits the updated lock file through gh-actions-bot, Solves #1083.

## Changes
  -  [x] Added a job to check if the lock file is outdated and update it automatically (and rerun the tests). Proceed with other checks if the file is not outdated.
  - [x] Modified the code-style, docs, and pytest jobs to wait for the poetry.lock file to be updated. (they will fail if this is not done)
  - [x] Added a concurrency clause stating that, if the lock-file-test fails, cancel the other tests (explanation below)
  - [x] This job requires a Github Personal Access Token (PAT) with write access to the repository. (named temporarily as `GH_BOT_TOKEN`). This is required to be added from the github repo settings to allow the bot to push commits to the PR branch. (TBD) 

### 1. Why can't we remove the `poetry.lock` file on every PR?

As far as I read, deleting and replacing `poetry.lock` file on every merge request will update the file as per every contributor's configuration without any major changes to the dependency list, Making it less consistent and thus reproducible. This is probably why the official documentation suggests package developers to "refresh the lock file whenever there is a dependency update" (https://python-poetry.org/docs/basic-usage/#as-a-library-developer). And that is what I tried to follow here.

### 2. Is a new job necessary just to check poetry.lock file?

Every test in the workflow installs toqito through poetry. If we try to embed these lock-file-checks into any one job, the other jobs would have to wait for one check to complete (which wastes time if there is no lock-file-update). And moreover the other jobs will not identify the updated files. Adding a seperate job will speed up the auto-commit process, and ensure tests run parallelly. 

This job is set to run on just one python version because it ensures the lock file is committed only once. (and test results mostly should not change depending on python version).

### 3. Why force the other jobs wait for poetry lock check to complete?

If this is not done, all jobs would parallelly install toqito and fail due to the incorrect lock file, wasting resources. Moreover on average the `check-lock` routine takes about 15-20 seconds to complete which is non-disruptive than other alternatives.  

### 4. Why do we need concurrency in workflows now?

Whenever there is an autocommit from the `gh-actions-bot`, the older test (which are bound to fail) keep running unnecessarily. wasting time and resources. With concurrency we can cancel other jobs if the `check-lock` job fails (as we know that the other jobs will fail for sure).

### 5. Do we really require a Github Personal Access Token?

Using Github PAT is required to authorize the actions-bot to push commits to the branch. `{{ secrects.GITHUB_TOKEN }}` can also be used but this cannot trigger automatic tests. So afaik it is necessary to go forward with PAT.

## Checklist
  -  [x] check for `ruff`, `pytest`, `Sphinx`, `linkcheck` and `doctest` errors.
  -  [x] Tested that the workflow works on a PR in my private toqito fork (https://github.com/NV-Karthik/toqito/pull/2). I deliberately added matplotlib as a dependency (and removed it) to test the gh-actions-bot auto-commiting the poetry.lock file. Please feel free to look at the tests and commit messages (from the bot) at the final stage of the PR.


vprusso (approved):
Think this looks good to me, but I'll go ahead and have @purva-thakre give it a look as well. Thank you, @NV-Karthik !

NV-Karthik:
Hi @vprusso, in the older PR we were discussing on my decision to run the lock-file test only for python 3.10. (https://github.com/vprusso/toqito/pull/1190#discussion_r2062616181)

Shall I go ahead and mark it as resolved before closing that PR? or do you want me to change the version to something else? Thanks

vprusso:
> Hi @vprusso, in the older PR we were discussing on my decision to run the lock-file test only for python 3.10. ([#1190 (comment)](https://github.com/vprusso/toqito/pull/1190#discussion_r2062616181))
> 
> Shall I go ahead and mark it as resolved before closing that PR? or do you want me to change the version to something else? Thanks

That sounds like a plan to me if that works for you, @NV-Karthik !

purva-thakre (approved):

purva-thakre:
I think something from this PR is not working correctly. Merging the PR did not trigger the github workflows. 

image

See the other latest commit where the github workflow was triggered.

PR #1235, opened 2025-05-31
Fix: circular import with kraus to choi

### LINKED ISSUE #621: TODO in `is_separable` check
https://github.com/vprusso/toqito/blob/0b62ebfdce13b739cdecccd48b3a27d1cbd81a30/toqito/state_props/is_separable.py#L122-L133

Related to #518

### LINKED ISSUE #1081: Circular import with kraus_to_choi in channels directory
## Description

The order of the listing of imports in the `__init__.py` files can cause circular imports to occur. These can happen if the ordering of the listing has a function that also uses a subsequent implementation of a function that is listed below. 

Ideally, we don't want to construct such a fragile tower of dependencies in this way, as even if this current arrangement of items "works", it's just a matter of time before a new feature or refactor occurs to cause an issue (refer to the example below for one such instance, but this is also not an isolated incident). 

We should consider some of other type of strategy here where exposing these functions for external usage won't cause any such circular import error to crop up. 

## Example
It appears that the common usage of the `kraus_to_choi` function within the `channels/` directory is leading to some strange errors which are rooted in the fact that circular imports are occurring. For instance, this can be seen in the following linked comments:
https://github.com/vprusso/toqito/pull/1077#issuecomment-2751037697
https://github.com/vprusso/toqito/pull/1072#issuecomment-2750345195

In both of these, there is no explicit mention of a circular import being the root cause, but upon further analysis, this appears to be the problem. Indeed, moving the function `kraus_to_choi` from the global import scope to the function-level import scope resolved the issue in the aforementioned issues. 

However, this solution is not ideal. The imports should, as a general rule, be at the top of the file. We likely need to find a better solution keeping this convention and yet avoiding circular imports.

poig:
## Description

close #1081 by refactoring core utility functions to establish a clearer, acyclic dependency hierarchy within `toqito`. This fixes initialization order problems that previously led to `ImportError` or `TypeError: 'module' object is not callable` when importing or using functions like `kraus_to_choi` via the `toqito.channels` namespace, especially during full test suite runs.

The primary insight is that foundational utilities like `partial_trace` and `partial_transpose` should not reside in a higher-level package like `toqito.channels` if lower-level packages (e.g., `toqito.channel_ops`) depend on them.

## Changes

*   **Resolves Circular Dependencies/Initialization Conflicts:** The primary cause of the errors in #1081 is addressed by ensuring a unidirectional flow of dependencies for core utilities.
*   **Increased Robustness:** The import system is now less fragile and less dependent on the specific order of imports within `__init__.py` files.
*   **Clearer Architecture:** Establishes a more layered structure:
    *   `toqito.matrix_ops`: Foundational utilities.
    *   `toqito.channel_ops`, `toqito.state_props`, etc.: Can depend on `matrix_ops`.
    *   `toqito.channels` (specific channel implementations): Can depend on `matrix_ops` and `channel_ops`.
    *   `toqito.channels.__init__.py`: Primarily aggregates local channel modules and provides a convenient user-facing namespace by re-exporting selected functions from `channel_ops` and `matrix_ops`.

*   **Guidance for Future Development:**
    1.  **Import from Canonical Locations Internally:** Within `toqito`'s own modules, prefer importing functionalities directly from the subpackage where they are defined (e.g., import `kraus_to_choi` from `toqito.channel_ops`, import `partial_trace` from `toqito.matrix_ops`).
    2.  **Use `__init__.py` for Public API Aggregation:** The `__init__.py` of a user-facing package like `toqito.channels` can re-export functions from other locations to create a convenient API for *users*. However, internal modules should favor direct imports for stability.
    3.  **Place New Utilities Foundationally:** If a new utility function is broadly applicable (like `partial_trace`), place it in a foundational package (`matrix_ops`, `state_ops`, `helpers`, etc.) rather than a higher-level one like `channels` if other subpackages will need it. This helps prevent future circular dependency issues.

This refactoring aims to make `toqito` more robust and easier to maintain. The "band-aid" fix of using function-level imports for `kraus_to_choi` should no longer be required.

now if we do ``from toqito.channels import kraus_to_choi, partial_trace, amplitude_damping`` There won't be any circular import error pop up anymore.

"__get_itter__" could also be useful is some scenario

a easy test to see if the file can import it is by:
```
print(f"Imported partial_trace: {partial_trace}")
print(f"Is it callable? {callable(partial_trace)}")
```
```
pycycle --here
```

## Checklist


toqito/channels/__init__.py
```diff
@@ -1,13 +1,39 @@
...
+    "depolarizing",
+    "pauli_channel",
+    "phase_damping",
+    "realignment",
+    "reduction", # Add if reduction is defined/imported above
+    "kraus_to_choi",
+    "choi_to_kraus", # Add if imported above
+    "partial_trace",    # If you decided to re-export for backward compatibility
+    "partial_transpose",# If you decided to re-export for backward compatibility
+]
```
purva-thakre, line 39: Can you describe your motivation for this addition? I am not too familiar with this.

poig: this is use for regular user, when where we write code with library, we aren't looking into the sources code most of the time, this will help notify the user what tool i can import from here

purva-thakre: But the public API already exists...

https://toqito.readthedocs.io/en/latest/autoapi/channels/index.html

poig: yes, it also can be use for ``from module import *``, otherwise it will import all file without underscore, might potential cause circular import

poig: I will delete it if its not been use at the end, but for now I will keep it there

purva-thakre: I am currently trying to understand the changes in this PR. But I do not understand how `pauli_channel` is being imported from `matrix_ops`. 

https://github.com/vprusso/toqito/actions/runs/15376598963/job/43262692906?pr=1235#step:7:22

toqito/matrix_ops/partial_trace.py
```diff
@@ -7,6 +7,8 @@
 from toqito.helper import expr_as_np_array, np_array_as_expr
 from toqito.perms import permute_systems
 
+# from toqito.channels import kraus_to_choi
+
```
purva-thakre, line 11: why is this commented out

toqito/channels/__init__.py
poig: that's the trade off there if we want ``from toqito.channels import kraus_to_choi, partial_trace``, vec are a follow up error after I resolve the first one, because its use in another file, we can't import something while they are seperated in matrix_ops and perms

poig: because partial_transpose need permute_systems from perms, which need vec from matrix_ops, if I don't move it to perms, it will be another circular import issues

toqito/matrix_ops/partial_trace.py
poig: its for Modules like is_ppt.py (in state_props), channel_fidelity.py (in channel_metrics), and any modules in channel_ops can now import partial_trace directly from toqito.matrix_ops, allows toqito.channels.__init__.py to more safely import from toqito.channel_ops (like kraus_to_choi) because the dependency chain of kraus_to_choi will no longer attempt to call back into a partially initialized toqito.channels to get partial_trace.

toqito/channels/__init__.py
vprusso: Hi @poig thanks for your efforts on this! This is definitely one that would be good to have properly resolved. Just to capture some of the high-level points from this thread: 

1. You're probably correct that things like `partial_transpose` and `partial_trace` could come from `matrix_ops` more appropriately, and keeping them also from the `channels/` module (as you have done) makes sense for backwards compatability. 

2. I think the broader question here is how to not make the order of imports in an `__init__.py` file matter. Using the `__all__` mechanism could certainly be one way of doing that, but I think before we go placing `partial_transpose.py` and `partial_trace.py`, etc. into another location, we should properly resolve the circular import issue. Once that is good done though, I agree with what you propose as a subsequent item, but perhaps we should try and keep things as clear and concise as we can for now. 

Any thoughts, and have you found that using `__all__` prevents any circular import issues in your investigations here?

purva-thakre: > have you found that using __all__ prevents any circular import issues in your investigations here

@vprusso From what I understand, `__all__` comes into play when the wildcard import `from toqito import *` is used. We do not use an import like this anywhere. Which is why I was a bit skeptical about the additions. 
 
https://stackoverflow.com/a/44842

poig:
> BTW I have been thinking `importlib` could be a good fallback option. https://docs.python.org/3/library/importlib.html#importlib.import_module
> 
> This defers the module initialization process until the function is executed.
> 
> Edit: [This blogpost](https://docs.kanaries.net/topics/Python/python-circular-import#using-the-importlibimport_module-function) states that indeed this could be useful for circular imports.

It doesn't inherently solve circular imports if the underlying module dependencies are still circular at the top level. If module A's top-level code imports module B, and module B's top-level code imports module A, using importlib.import_module("B") inside A will still hit the same problem. It just delay import, similar to ``__getattr__``, while importlib.import_module could be used for lazy imports to break cycles, the module-level __getattr__ is often a more direct solution if the goal is to lazily provide names from an __init__.py that are sourced from other modules. However, both are ways to defer the import, not necessarily to fix the underlying architectural reason for the cycle.

purva-thakre:
There is a build error. https://github.com/vprusso/toqito/actions/runs/15434998384/job/43475316567?pr=1235#step:7:30

toqito/channel_metrics/__init__.py
```diff
@@ -5,3 +5,9 @@
 from toqito.channel_metrics.fidelity_of_separability import fidelity_of_separability
 from toqito.channel_metrics.completely_bounded_trace_norm import completely_bounded_trace_norm
 from toqito.channel_metrics.completely_bounded_spectral_norm import completely_bounded_spectral_norm
+
+
+__all__ = ["channel_fidelity", "completely_bounded_spectral_norm",
+    "completely_bounded_trace_norm", "diamond_distance",
+    "fidelity_of_separability",
+]
```
purva-thakre, line 13: Maybe I am missing something here https://github.com/vprusso/toqito/pull/1235#discussion_r2125571480, but I still do not see a reason for `from toqito import *`.  Even the [PEP 8 style guide](https://peps.python.org/pep-0008/#imports) states that wildcard imports like this should be avoided. 

@vprusso Thoughts?

purva-thakre (approved):
Once you have removed the `__all__` additions to the init files, this PR LGTM. I'll wait for @vprusso to approve before merging this PR.

Thanks for the great work in https://github.com/vprusso/toqito/pull/1235#discussion_r2125496962 and https://github.com/vprusso/toqito/pull/1235#discussion_r2125571480 helping us debug the issue! If we run into similar errors in the future, we now have a better idea of where the partially initialized modules might be. If there's anything to add about the debugging process, please add a comment to help future us. 

Great work @poig!

toqito/channel_metrics/__init__.py
poig: for me as a programmer, I would love to have that, I can undo everything if you guys don't care a user explores toqito, say toqito.channels, having __all__ in channels/__init__.py clearly tells them: "These are the functions/classes you are meant to use directly from this namespace." Without it, they might see (via dir() or IDE introspection) many imported modules or helper functions that are actually implementation details of channels/__init__.py or its submodules. __all__ filters this down to the intended interface.

Many IDEs and interactive shells (like IPython/Jupyter) use __all__ (if present) to provide more accurate and less cluttered autocompletion suggestions. If you type from toqito.channels import and press Tab, you're more likely to see only the public API if __all__ is well-defined.

It doesn't fix the import errors, but it makes the (now correctly working) package easier and safer to use and understand. Adding __all__ is like putting clear labels on the doors ("Living Room", "Kitchen", "Private - Staff Only"), putting up a clear floor plan for visitors, and ensuring the light switches in the public areas are easy to find and don't accidentally control something in the basement. It's about usability, safety, and clarity for those interacting with the house.

but since you guys didn't care much now, I will delete it for now as I already solve the main challenge here already

PR #1250, opened 2025-06-06
let isort change init import order & add back operator sinkhorn

### LINKED ISSUE #848: Feature: OperatorSinkhorn function

purva-thakre:
## Description

Closes #848 

## Changes

  -  [ ] Change 1

## Checklist

### DIFF
--- a/docs/refs.bib
+++ b/docs/refs.bib
@@ -443,6 +443,21 @@ primaryClass = {quant-ph}
   month={Dec} 
 }
 
+@article{Gurvits_2004_Classical,
+  title={Classical complexity and quantum entanglement},
+  journal={Journal of Computer and System Sciences},
+  volume={69},
+  number={3},
+  pages={448-484},
+  year={2004},
+  note={Special Issue on STOC 2003},
+  issn={0022-0000},
+  doi={https://doi.org/10.1016/j.jcss.2004.06.003},
+  url={https://www.sciencedirect.com/science/article/pii/S0022000004000893},
+  author={Leonid Gurvits},
+  keywords={Entanglement, Complexity, Determinant}
+}
+
 @article{Gühne_2009_Horodecki,
   title={Entanglement detection},
   journal={Physics Reports},
@@ -1002,6 +1017,19 @@ abstract = {One of the most cited books in physics of all time, Quantum Computat
   url={https://sites.google.com/site/jamiesikora/teaching/sdps-in-quantum-theory}
 }
 
+@article{Sinkhorn_1964_Relationship,
+  author={Richard Sinkhorn},
+  title={A relationship between arbitrary positive matrices and doubly stochastic matrices},
+  volume={35},
+  journal={The Annals of Mathematical Statistics},
+  number={2},
+  publisher={Institute of Mathematical Statistics},
+  pages={876 -- 879},
+  year={1964},
+  doi={10.1214/aoms/1177703591},
+  URL={https://doi.org/10.1214/aoms/1177703591}
+}
+
 @misc{SO_43884189,
   author={Stack Overflow Post},
   title={Check if a large matrix is diagonal matrix in python},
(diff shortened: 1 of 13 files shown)


purva-thakre:
RTD build is failing due to a circular import error. 🤯 

recent work related to circular import is in https://github.com/vprusso/toqito/pull/1235#discussion_r2125496962

vprusso:
> RTD build is failing due to a circular import error. 🤯
> 
> recent work related to circular import is in [#1235 (comment)](https://github.com/vprusso/toqito/pull/1235#discussion_r2125496962)

Good intel gathering here, @purva-thakre , I think we would have eventually hit this case one way or another, so it's better to get out in front of it and try and really determine what the root problem here is likely to be. 

Any thoughts on why your PR for this may not have been applicable in this case, @poig ?

purva-thakre:
I think we need a `utils` folder where we put all the functions that lead to circular import errors.

poig:
> > RTD build is failing due to a circular import error. 🤯
> > recent work related to circular import is in [#1235 (comment)](https://github.com/vprusso/toqito/pull/1235#discussion_r2125496962)
> 
> Good intel gathering here, @purva-thakre , I think we would have eventually hit this case one way or another, so it's better to get out in front of it and try and really determine what the root problem here is likely to be.
> 
> Any thoughts on why your PR for this may not have been applicable in this case, @poig ?

```
[Your Code]
    |
    v
toqito.channel_ops.__init__.py
    |
    +-----> tries to import from toqito.channel_ops.partial_channel.py (BEGINS LOADING)
                  |
                  v
              (This file needs something from toqito.states)
                  |
                  v
              toqito.states ... -> toqito.state_props.is_separable.py
                  |
                  v
              (This file needs to import from toqito.channel_ops.partial_channel.py)
                  |
                  +-----> LOOP! Tries to import from a module that is not done loading.
```

so we basically got few option here
1. moving file since channel_ops -> states -> state_props -> channel_ops, we can 
    - Move the file `toqito/channel_ops/apply_channel.py` 
    - to the more appropriate location: `toqito/matrix_ops/apply_channel.py`.
Tips: always look at the last file that cause the issue and see what its trying to import from the initial location, then just move it to a better location(just move it to matrix_ops where it doesn't need channels_ops), would mostly solve the issue, its kind of like sorting file.

2. use lazy import, but performance will be hurt
```
# toqito/matrix_props/is_orthonormal.py

import importlib.util
import sys

# 1. Find the "spec" for the module we want to lazy load.
#    The spec contains metadata about how to load it.
spec = importlib.util.find_spec("toqito.state_props")

# 2. Create a "LazyLoader" instance.
#    This object will act as our module proxy.
loader = importlib.util.LazyLoader(spec.loader)

# 3. Create a module object from the spec and attach the lazy loader.
state_props = importlib.util.module_from_spec(spec)
state_props._is_loading = True # A flag to prevent re-entrancy issues
spec.loader.exec_module(state_props)

# 4. Put the lazy-loaded module into sys.modules. This makes it
#    available for other parts of the code to use.
sys.modules["toqito.state_props"] = state_props
```

this is architecture design issue and to really fix them we should 
- Step 1: Identify the most fundamental functions.
- Step 2: Relocate the fundamental functions to a low-level module.

vprusso:
> maybe a week after the hack I will be free, I will see what I can contribute.

We'd be much obliged if that would be possible! As @purva-thakre has mentioned, this one has been quite the thorn in our side. Your merged PR still is an improvement, but it would be great to actually nip this thing in the bud once and for all and get it resolved. Thank you, @poig !

purva-thakre:
Closing for now.

PR #1236, opened 2025-06-01
Features: Enhance the is_separable function

poig:
## Description
close #621 It integrates several new separability/entanglement checks, by incorporating additional separability checks, including the full operational criterion for low-rank PPT states, and by improving documentation and code structure. The changes aim to provide more robust entanglement detection and separability verification.

A key improvement is the reordering of these checks to correctly identify complex cases, such as certain bound entangled states (e.g., the "Tiles" state which was previously misclassified).

## Changes

- [x] Refined Order of Checks: The sequence of separability tests has been logically reordered for efficiency and correctness, ensuring specific necessary and sufficient conditions (e.g., for 3x3 rank-4 PPT states) are evaluated appropriately relative to more general criteria.

- [x] Enhanced Input Validation: Strengthened the initial validation of the input state and dimension parameters for greater robustness.

- [x] Expanded the main docstring for is_separable to provide a detailed overview of the criteria checked.

- [x] Incorporated :cite: directives for numerous criteria, linking them to their respective sources in the bibliography (assuming refs.bib is updated accordingly).

- [x] Boolean Return Types: Ensured all return paths explicitly yield Python bool types (True/False) to prevent issues with numpy.bool_ in assertions.

- [x] Maintained Existing Criteria: Preserved and integrated user-provided logic for specific checks, including the Plucker coordinate-based test for 3x3 rank-4 states and other QETLAB-inspired criteria like Ha-Kye and Breuer-Hall maps.

- [x] added citation ``ref.bib``

## Checklist
Before marking your PR ready for review, make sure you checked the following locally. If this is your first PR, you might be notified of some workflow failures after a maintainer has approved the workflow jobs to be run on your PR. 

Additional information is available in the [documentation](https://toqito.readthedocs.io/en/latest/contributing.html#testing).

  -  [x] Use `ruff` for errors related to code style and formatting.
  -  [x] Verify all previous and newly added unit tests pass in `pytest`.
  -  [x] Check the documentation build does not lead to any failures. `Sphinx` build can be checked locally for any failures related to your PR

PR #1232, opened 2025-05-28
feat: add is_abs_ppt function to check if a density matrix is absolut…

luffy-orf:
### Overview

This PR implements the `is_abs_ppt` function in `toqito.state_props`, providing a way to determine if a bipartite density matrix is **absolutely PPT** (i.e., PPT from spectrum), as requested in #208 and inspired by QETLAB's [IsAbsPPT](https://qetlab.com/IsAbsPPT) and [AbsPPTConstraints](https://qetlab.com/AbsPPTConstraints).

### Key Changes

- **New function:** `is_abs_ppt` for checking if a density matrix is absolutely PPT.
    - Supports 2x2, 2x3, and 3x3 bipartite systems (explicit constraint matrices).
    - Returns `True` if absolutely PPT, `False` if not, and `-1` for unsupported higher dimensions.
    - Validates input for squareness, Hermiticity, trace, and positivity.
- **Helper:** Internal function for generating constraint matrices as per Hildebrand (2007).
- **Tests:** Added tests for maximally mixed states, random states, known non-absolutely PPT states, invalid input, and large dimensions.
- **API:** Exposed `is_abs_ppt` in the public API.

Closes #208

Issue #1197, opened 2025-05-04
Bibliography references link to tutorials over function


vprusso:
There seems to be an odd behavior where if a given reference link is used both in the function docstring and also in a tutorial in the docs, clicking on the reference link in the function docstring takes the user to the reference in the tutorial (instead of the local one in the function).

One example of this is in [`state_props.is_antidistinguishable`](https://toqito.readthedocs.io/en/latest/autoapi/state_props/is_antidistinguishable/index.html)

However, this is just one example, and this behavior works the same way for any reference that has this type of "double citation".

vprusso:
Having separate bib files might be the way to go, but for what it's worth, I did put up an issue on their board. Thanks , @purva-thakre !

https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/370

PR #1256, opened 2025-06-10
Add `is_absolutely_ppt` and `abs_ppt_constraints`

### LINKED ISSUE #208: Feature: Function to determine if density matrix is absolutely PPT
Similar to the `IsAbsPPT` function from QETLAB ([link](https://github.com/nathanieljohnston/QETLAB/blob/master/IsAbsPPT.m)).

This would also require implementing the `AbsPPTConstraints` or equivalent in toqito as well.

Newtech66:
## Description
Resolves #208.

## Changes
  -  [x] Add `is_absolutely_ppt`
  -  [x] Add `abs_ppt_constraints`
  -  [x] Docstrings
  -  [x] References
  -  [x] Tests
  -  [x] (additional) fix a reference (`Hildebrand_2005_PPT` was an incorrect reference; replaced by `Hildebrand_2007_AbsPPT`)

## Notes:
  -  `abs_ppt_constraints` does not include the "criss-cross" subordering check: it is faster to not do the check, and it is okay to have redundant constraint matrices. The check is important if we want to create constraints for `cvxpy` however: the fewer (redundant) constraint matrices there are, the easier it is for `cvxpy` to compile + solve a problem.

## Checklist


purva-thakre:
@Newtech66 Read the feedback from #1232 as we do not want to make the same exact changes as the MATLAB function.

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,87 @@
...
+        return 1
+    # 4. Check Theorem 7.2 of arXiv:1406.1277
+    if sum(eigs[:p-1]) <= eigs[-1] + sum(eigs[-p:]):
+        return 1
+
+    # Main check
+    # 33592 is the upper bound on the number of constraint matrices that is required
+    # for p = 6 without any additional checks in abs_ppt_constraints
+    # Reference:
+    # http://njohnston.ca/2014/02/counting-the-possible-orderings-of-pairwise-multiplication/
```
purva-thakre, line 79: I know this a WIP PR but please take a look at the contributing guide on how to add references. There is a `refs.bib` file where we store our references.

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,87 @@
...
+        ==========
+        .. bibliography::
+            :filter: docname in docnames
+
+        :param mat: A square matrix.
+        :param dim: A 1-by-2 vector containing the dimensions of the
+                    subsystems on which :code:`mat` acts.
+        :return: :code:`1` if :code:`mat` is absolutely PPT,
+        :return: :code:`0` if :code:`mat is not absolutely PPT,
+        :return: :code:`-1` if the function could not decide.
```
purva-thakre, line 34: According to @vprusso, this is MATLAB convention, which is what I was referring to in an earlier comment. See https://github.com/vprusso/toqito/pull/1232#discussion_r2112387569

toqito/state_props/is_abs_ppt.py
Newtech66: Added references.

Newtech66:
Added tests. Review requested. @vprusso @purva-thakre

Newtech66:
I know, but as I said, the problem is that I can't find a way to construct a non-trivial absolutely PPT state that isn't caught by one of the quick checks. I'll fix the ruff related errors though.

purva-thakre:
> I know, but as I said, the problem is that I can't find a way to construct a non-trivial absolutely PPT state that isn't caught by one of the quick checks. I'll fix the ruff related errors though. If you could find any reference which constructs a non-trivial absolutely PPT or absolutely separable state, I could add it. It must not belong to the separable ball.

It might be possible to use mock testing. https://stackoverflow.com/questions/2665812/what-is-mocking

Newtech66:
I reordered the checks so that all the preliminary checks are tested.

toqito/state_props/tests/test_is_abs_ppt.py
```diff
@@ -0,0 +1,54 @@
...
+    """Test that matrix in separable ball returns True."""
+    mat = np.diag([0.7, 0.7, 0.2, 0.2])
+    mat /= np.trace(mat)
+    assert is_abs_ppt(mat)
+
+
+def test_not_absolutely_ppt():
+    """Test that a random PSD matrix is not PPT. Passes with high probability."""
+    mat = random_psd_operator(40)
+    assert not is_abs_ppt(mat)
```
purva-thakre, line 39: You can use `parametrize` for these tests too.

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,92 @@
...
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
+    :param eigs: A list of eigenvalues.
+    :param p: The dimension of the smaller subsystem in the bipartite system.
+    :param lim_cons: The maximum number of constraint matrices to compute. By default, this is
+                     equal to :math:`33592` which is an upper bound on the optimal number of
+                     constraint matrices which must be computed for :math:`p \leq 6`
+                     :cite:`Johnston_2014_Orderings`.
```
purva-thakre, line 37: This reference is blank in the docs build. https://toqito--1256.org.readthedocs.build/en/1256/autoapi/state_props/abs_ppt_constraints/index.html

image

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,92 @@
...
+    :return: A list of `lim_cons` constraint matrices which must be positive
```
purva-thakre, line 38:
```suggestion
    :return: A list of :code:`lim_cons` constraint matrices, which must be positive
```

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,108 @@
...
+    :param atol: The absolute tolerance parameter (default 1e-08).
+    :return: :code:`True` if :code:`mat` is absolutely PPT, :code:`False` if :code:`mat is not
+             absolutely PPT, and :code:`None` if the function could not decide.
+    :notes: If :code:`min(dim)` :math:`\leq 6`, this function checks all constraints
+            and therefore returns :code:`True` or :code:`False` in all cases. However, if
+            :code:`min(dim)` :math:`\geq 7`, only the first :math:`33592` constraints are
+            checked, since there are over :math:`23178480` constraints in this case
+            :cite:`Johnston_2014_Orderings`. Therefore the function returns either
+            :code:`False` if at least one constraint was not satisfied, or :code:`None`
+            if all checked constraints were satisfied.
```
purva-thakre, line 59: You can add a note with this https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html#notes

toqito/state_props/tests/test_abs_ppt_constraints.py
```diff
@@ -0,0 +1,54 @@
+"""Test abs_ppt_constraints."""
+
+import numpy as np
+import pytest
+
+from toqito.state_props import abs_ppt_constraints
+
```
purva-thakre, line 7: if the unit tests follow the same pattern, it might be easier to use parametrize.

purva-thakre (changes requested):
Took a quick look through the docstrings and unit tests. I will go over the code later tonight and tomorrow morning.

toqito/state_props/abs_ppt_constraints.py
Newtech66: image
This one isn't working either

toqito/state_props/tests/test_is_abs_ppt.py
Newtech66: Done.

purva-thakre:
> I can't seem to get it to work. I tried `monkeypatch.setattr("toqito.state_props.in_separable_ball", lambda mat: False)`, but it's not working.

You can simply define a mock object that satisfies these conditions. Then, pytest will consider the line as covered by unit tests if we get the expected result. 

image

just an FYI, we have until Monday to finalize this PR.

toqito/state_props/abs_ppt_constraints.py
purva-thakre: I wonder if this is related to #1197

toqito/state_props/tests/test_abs_ppt_constraints.py
Newtech66: Done.

toqito/state_props/abs_ppt_constraints.py
Newtech66: Well it doesn't seem to be an issue with my PR: if you check the read-the-docs build, the references are broken on ALL pages. For example, [see this page](https://toqito--1256.org.readthedocs.build/en/1256/autoapi/channel_ops/choi_to_kraus/index.html) which is unrelated to the PR.

Newtech66: I take that back, I just built it locally and it might be a problem with my PR that's somehow killing a lot of the references...

I'm getting a ton of "reference not found" warnings, I'll figure out what's going on.

Newtech66: @purva-thakre I think I tracked it down: The issue definitely stems from PR #1236. If you check the read-the-docs build for the [same page](https://toqito--1236.org.readthedocs.build/en/1236/autoapi/channel_ops/choi_to_kraus/index.html), the references are broken. poig mentions this in [this comment](https://github.com/vprusso/toqito/pull/1236#issuecomment-2953500174) as well. So I suppose it's not an issue from my PR, but I'll try to track down and solve this bug in this PR if I can.

Newtech66: Seems like what happened is the same paper is available under two different names. This is [reference 1](https://iopscience.iop.org/article/10.1088/0305-4470/39/45/020) and this is [reference 2](https://arxiv.org/abs/quant-ph/0607035). They're the same paper because the abstract is identical. I will remove the arxiv reference because it was published earlier (and the later version would be better?).

Newtech66:
@purva-thakre I didn't have to mock: I was able to use AMPLpy + Gurobi to explicitly construct an absolutely PPT $4 \times 4$ matrix.

Here's the code if you're interested:

```python
ampl = AMPL()
ampl.eval(r"""
param n, default 2;
param n2, default n * n;
var a{1..n2} >= 0;

maximize ob:
  0;

subject to trace:
  sum{j in 1..n2} a[j] = 1;
subject to order{j in 1..(n2 - 1)}:
  a[j] >= a[j + 1];
subject to circle:
  sum{j in 1..n} (a[j] + a[n2 - j]) >= 2 * a[n2];
subject to fro:
  sum{j in 1..n2} (2 * a[j] - 2.9 * a[j] * a[j]) <= 1;
subject to absppt:
  (a[1] - a[3]) * (a[1] - a[3]) <= 4 * a[2] * a[4];
""")

ampl.solve(solver="gurobi")

print('Solve status:', ampl.solve_result)
a = ampl.get_variable("a").get_values().to_pandas()['a.val'].to_numpy()
print('Eigenvalues:', a)
````

File: toqito/state_props/abs_ppt_constraints.py [hunk 8]
```diff
@@ -0,0 +1,91 @@
+"""Compute the constraints on a spectrum for it to be absolutely PPT."""
+
+import numpy as np
+
+
+def abs_ppt_constraints(eigs: np.ndarray, p: int, lim_cons: int = 33592) -> list[np.ndarray]:
```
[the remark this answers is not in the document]
Newtech66: Done.

docs/refs.bib
```diff
@@ -700,6 +704,20 @@ @misc{Hall_2006_Indecomposable
...
+journal = {Linear Algebra and its Applications},
+volume = {469},
+pages = {276-304},
+year = {2015},
+issn = {0024-3795},
+doi = {https://doi.org/10.1016/j.laa.2014.11.031},
+url = {https://www.sciencedirect.com/science/article/pii/S0024379514007666},
+author = {Maria Anastasia Jivulescu and Nicolae Lupa and Ion Nechita and David Reeb},
+keywords = {Quantum entanglement, Reduction criterion, Entanglement from eigenvalues},
+abstract = {We study the problem of whether all bipartite quantum states having a prescribed spectrum remain positive under the reduction map applied to one subsystem. We provide necessary and sufficient conditions, in the form of a family of linear inequalities, which the spectrum has to verify. Our conditions become explicit when one of the two subsystems is a qubit, as well as for further sets of states. Finally, we introduce a family of simple entanglement criteria for spectra, closely related to the reduction and positive partial transpose criteria, which also provide new insight into the set of spectra that guarantee separability or positivity of the partial transpose.}
```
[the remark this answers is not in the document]
Newtech66: I've opened a PR to address this.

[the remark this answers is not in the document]
vprusso: Sounds good, thank you, @Newtech66 !

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,91 @@
...
+    def _create_constraint(eigs: np.ndarray, X: np.ndarray, p: int) -> np.ndarray:
+        r"""Return constraint matrix from order matrix."""
+        L = np.zeros((p, p))
+        # Set upper triangle + diagonal
+        upper_inds = np.triu_indices(p)
+        L[upper_inds] = eigs[-X[upper_inds]]
+        strictly_upper_inds = np.triu_indices(p, 1)
+        # Set lower triangle
+        L.T[strictly_upper_inds] = -eigs[np.unique(X[strictly_upper_inds], return_inverse=True)[1]]
+        return L + L.T
```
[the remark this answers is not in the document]
Newtech66: So, if I remove `np.sort(eigs)[::-1]`, I can pass a `cvxpy` Variable of the right size to `abs_ppt_constraints` and it generates the right constraint matrices, except each entry is a `cvxpy` Expression, so it can be used with CVXPY as it is done in QETLAB. The issue I have with this is that it means that, if the user passes in a normal array, it is up to the user to ensure that the array they pass is sorted. I think that's error-prone for the user, so I'd want to keep the sort at the beginning, maybe with a type check with `isinstance`. What do you think?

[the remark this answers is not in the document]
Newtech66: Also, like, I don't know how I should explain to the user how to use this function with CVXPY without it looking awkward, maybe I should put it in a notes block.

docs/refs.bib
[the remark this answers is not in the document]
Newtech66: Done.

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,92 @@
...
+    :param max_constraints: The maximum number of constraint matrices to compute. By default, this is
+                            equal to :math:`33592` which is an upper bound on the optimal number of
+                            constraint matrices which must be computed for :math:`p \leq 6`
+                            :cite:`Johnston_2014_Orderings`.
+    :return: A list of :code:`max_constraints` constraint matrices which must be positive
+             semidefinite for an absolutely PPT spectrum.
+
+    """
+    eigs = np.sort(eigs)[::-1]
+    p_p = p * (p + 1) // 2
```
purva-thakre, line 43: this is supposed to be $p_{+}$, right?

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,110 @@
...
+        # Floating-point arithmetic is risky
+        dim = 1
+        for j in range(1, nm + 1):
+            if j**2 > nm:
+                break
+            if nm % j == 0:
+                dim = j
+
+    if nm % dim != 0:
+        raise ValueError("dim must divide the dimensions of the matrix")
```
purva-thakre, line 81:
```suggestion
    if nm % dim != 0:
        raise ValueError("Calculated matrix dimension and provided dimension are unequal.")
```

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,110 @@
...
+    eigs = np.sort(np.linalg.eigvalsh(mat))[::-1]
+    # 2. Is PSD
+    if eigs[-1] < -abs(atol):
+        return False
+    # 3. Check Theorem 7.2 of :cite:`Jivulescu_2015_Reduction`
+    if sum(eigs[: p - 1]) <= eigs[-1] + sum(eigs[-p:]):
+        return True
+    # 4. Check if mat is in separable ball
+    if in_separable_ball(mat):
+        return True
```
purva-thakre, line 102: You should describe these checks in the function docstring.

toqito/state_props/tests/test_abs_ppt_constraints.py
```diff
@@ -0,0 +1,59 @@
...
+    """Test constraint matrices."""
+    cons = abs_ppt_constraints(eigs, p)
+    expected_cons = analytical_constraints(eigs)
+    any_match = False
+    for cons_ordered in permutations(cons):
+        all_match = True
+        for computed_cons, analytical_cons in zip(cons_ordered, expected_cons):
+            all_match &= computed_cons == pytest.approx(analytical_cons)
+        any_match |= all_match
+    assert any_match
```
purva-thakre, line 45: making a note to review this unit test later.

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,110 @@
...
+    ==========
+    A random density matrix will likely not be absolutely PPT:
+
+    .. jupyter-execute::
+
+        import numpy as np
+        from toqito.rand import random_density_matrix
+        from toqito.state_props import is_abs_ppt, is_separable
+        rho = random_density_matrix(9) # assumed to act on a 3 x 3 bipartite system
+        print("rho is absolutely PPT:", is_abs_ppt(rho, 3))
```
vprusso, line 34:
```suggestion
        print(f"ρ is absolutely PPT: {is_abs_ppt(rho, 3)}")
```

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,110 @@
...
+
+    The maximally-mixed state is an example of an absolutely PPT state:
+
+    .. jupyter-execute::
+
+        import numpy as np
+        from toqito.states import max_mixed
+        from toqito.state_props import is_abs_ppt, is_separable
+        rho = max_mixed(9) # assumed to act on a 3 x 3 bipartite system
+        print("rho is absolutely PPT:", is_abs_ppt(rho, 3))
```
vprusso, line 44:
```suggestion
        print(f"ρ is absolutely PPT: {is_abs_ppt(rho, 3)}")
```

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Examples
+    ==========
+    A random density matrix will likely not be absolutely PPT:
+
+    .. jupyter-execute::
+
+        import numpy as np
+        from toqito.rand import random_density_matrix
+        from toqito.state_props import is_abs_ppt, is_separable
```
vprusso, line 32: why is `is_separable` being imported here?

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,110 @@
...
+        rho = random_density_matrix(9) # assumed to act on a 3 x 3 bipartite system
+        print("rho is absolutely PPT:", is_abs_ppt(rho, 3))
+
+    The maximally-mixed state is an example of an absolutely PPT state:
+
+    .. jupyter-execute::
+
+        import numpy as np
+        from toqito.states import max_mixed
+        from toqito.state_props import is_abs_ppt, is_separable
```
vprusso, line 42: similar question here about `is_separable`.

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,110 @@
...
+    n, m = dim, nm // dim
+    p = min(n, m)
+
+    # Quick checks
+    # 1. Is Hermitian
+    if not is_hermitian(mat, rtol, atol):
+        return False
+    # Compute eigenvalues (in descending order)
+    # eigsvalsh normally returns eigenvalues in ascending order
+    # But it is risky to assume this will remain the default behaviour in the future
```
vprusso, line 92: [praise]: I like this line of thinking. Nice work.

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,92 @@
...
+    X = (p_p + 1) * np.ones((p, p), dtype=np.int32)
```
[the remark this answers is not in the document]
Newtech66: Yeah, I just sort of copied over the exact same variable names from the QETLAB function. I'll change this and others like it.

toqito/state_props/is_abs_ppt.py
Newtech66: Oh, whoops, that was from a time when I thought being absolutely PPT is a necessary condition for being separable. Later I realised that that's not true: PPT is necessary for separability, and absolutely PPT is necessary for absolute separability. I removed the lines talking about separability, but forgot to remove this import. I'll remove it.

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,92 @@
...
+        L = np.zeros((p, p))
+        # Set upper triangle + diagonal
+        upper_inds = np.triu_indices(p)
+        L[upper_inds] = eigs[-X[upper_inds]]
+        strictly_upper_inds = np.triu_indices(p, 1)
+        # Set lower triangle
+        L.T[strictly_upper_inds] = -eigs[np.unique(X[strictly_upper_inds], return_inverse=True)[1]]
+        return L + L.T
+
+    _fill_matrix(0, 2, 3)
```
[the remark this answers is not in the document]
Newtech66: The algorithm processes the upper triangle of `X` in row major order. We set the first two elements of the first row, so we start from the third element of the first row (that's `(0, 2)` in zero-based indexing). And `3` because we already used `1` and `2`. I'll add more comments to explain how the algorithm works I suppose. And use more descriptive variable names.

docs/refs.bib
```diff
@@ -737,6 +753,13 @@ @article{Johnston_2013_Spectrum
   url = {https://link.aps.org/doi/10.1103/PhysRevA.88.062330}
 }
 
+@misc{Johnston_2014_Orderings,
+  title = {Counting the Possible Orderings of Pairwise Multiplication},
+  author = {Johnston, Nathaniel},
+  year = {2014},
+  url = {http://njohnston.ca/2014/02/counting-the-possible-orderings-of-pairwise-multiplication/},
+}
+
```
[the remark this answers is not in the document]
Newtech66: Would be nice to run a program on `refs.bib` to sort all entries in alphabetical order by their key instead of resolving these things manually...

toqito/state_props/is_abs_ppt.py
Newtech66: No, `dim` can be any positive number that divides the dimension of the matrix.

toqito/state_props/abs_ppt_constraints.py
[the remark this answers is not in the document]
Newtech66: There's no cvxpy code block, the usage of cvxpy is completely external. This function would just constructs a matrix out of the input, and it's possible write the function such that the input can both be a numpy array or a cvxpy Variable.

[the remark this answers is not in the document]
Newtech66: I added `cvxpy` support.

toqito/state_props/is_abs_ppt.py
```diff
@@ -0,0 +1,121 @@
...
+        return True
+    # 4. Check if mat is in separable ball
+    if in_separable_ball(mat):
+        return True
+
+    # Main check
+    constraints = abs_ppt_constraints(eigs, p)
+    for constraint in constraints:
+        if not is_positive_semidefinite(constraint, rtol, atol):
+            return False
```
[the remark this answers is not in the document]
Newtech66: We don't allow passing a cvxpy Variable to `is_abs_ppt`. Also, we don't need to solve an SDP in this case. It's just that the conditions are framed as a set of PSD conditions.

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,185 @@
...
+                      number of constraint matrices. (default: :code:`False`)
+    :return: A list of :code:`max_constraints` constraint matrices which must be positive
+             semidefinite for an absolutely PPT spectrum.
+
+    """
+    if isinstance(eigs, np.ndarray):
+        # Sort eigenvalues in non-increasing order
+        eigs = np.sort(eigs)[::-1]
+    elif isinstance(eigs, cvxpy.Variable):
+        # Don't do anything
```
vprusso, line 76:
```suggestion
```

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,185 @@
...
+    :param p: The dimension of the smaller subsystem in the bipartite system.
+    :param max_constraints: The maximum number of constraint matrices to compute. (default: 33592)
+    :param use_check: Use the "criss-cross" ordering check described in :cite:`Johnston_2014_Orderings` to reduce the
+                      number of constraint matrices. (default: :code:`False`)
+    :return: A list of :code:`max_constraints` constraint matrices which must be positive
+             semidefinite for an absolutely PPT spectrum.
+
+    """
+    if isinstance(eigs, np.ndarray):
+        # Sort eigenvalues in non-increasing order
```
vprusso, line 73:
```suggestion
```

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,185 @@
...
+
+    References
+    ==========
+    .. bibliography::
+        :filter: docname in docnames
+
+    :raises TypeError: If :code:`eigs` is not a :code:`numpy` ndarray or a :code:`cvxpy` Variable.
+    :param eigs: A list of eigenvalues.
+    :param p: The dimension of the smaller subsystem in the bipartite system.
+    :param max_constraints: The maximum number of constraint matrices to compute. (default: 33592)
```
vprusso, line 65:
```suggestion
    :param max_constraints: The maximum number of constraint matrices to compute. (default: 33,592)
```

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,185 @@
...
+
+    """
+    if isinstance(eigs, np.ndarray):
+        # Sort eigenvalues in non-increasing order
+        eigs = np.sort(eigs)[::-1]
+    elif isinstance(eigs, cvxpy.Variable):
+        # Don't do anything
+        pass
+    else:
+        # Raise an error for unsupported type
```
vprusso, line 79:
```suggestion
```

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,185 @@
...
+        return []
+    if p == 2:
+        return [np.array([[2 * eigs[-1], eigs[-2] - eigs[0]], [eigs[-2] - eigs[0], 2 * eigs[-3]]])]
+
+    p_plus = p * (p + 1) // 2
+    order_matrix = np.zeros((p, p), dtype=np.int32)
+    available = np.ones(p_plus, dtype=bool)
+    constraints = []
+
+    # The first two elements of the first row and the last two elements of the last column are fixed
```
vprusso, line 93:
```suggestion
    # The first two elements of the first row and the last two elements of the last column are fixed.
```

vprusso (approved):
A few nit-picky comments, but LGTM overall, @Newtech66 !

Seems like there are some test failures, so modulo those and the minor comments, I think it's good to go! Nice work, @Newtech66 !

docs/refs.bib
```diff
@@ -688,6 +692,25 @@ @misc{Ha_2011_Positive_map
...
+pages = {276-304},
+year = {2015},
+issn = {0024-3795},
+doi = {https://doi.org/10.1016/j.laa.2014.11.031},
+url = {https://www.sciencedirect.com/science/article/pii/S0024379514007666},
+author = {Maria Anastasia Jivulescu and Nicolae Lupa and Ion Nechita and David Reeb},
+}
+
+@misc{Johnston_2014_Orderings,
+  title = {Counting the Possible Orderings of Pairwise Multiplication},
```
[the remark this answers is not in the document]
Newtech66: That's not the exact capitalization the source uses.
image

toqito/state_props/abs_ppt_constraints.py
```diff
@@ -0,0 +1,185 @@
...
+        :cite:`Johnston_2014_Orderings`.
+            * With :code:`use_checks=False`, the number of matrices returned starting from :math:`p=1` is
+              :math:`[0, 1, 2, 12, 286, 33592, 23178480, \ldots]`.
+            * With :code:`use_checks=True`, the number of matrices returned starting from :math:`p=1` is
+              :math:`[0, 1, 2, 10, 114, 2612, 108664, \ldots]`.
+        However, the optimal number of matrices starting from :math:`p=1` is given by :math:`[0, 1, 2, 10, 114,
+        2608, 107498]`.
+
+    .. note::
+        This function accepts a :code:`cvxpy` Variable as input for :code:`eigs`. The function will return the assembled
```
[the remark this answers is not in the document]
Newtech66: Ok that's actually a bad idea, it would really mess up the solver I wrote above if I allowed creation of constraint matrices through `abs_ppt_constraints`. Also that's not the approach QETLAB takes. QETLAB instead allows usage of `cvx` through `IsAbsPPT`. I'll take the same pattern here. Allow passing a cvxpy Variable to `is_abs_ppt` and then we can actually make the constraints we want.

toqito/state_props/tests/test_abs_ppt_constraints.py
Newtech66: Reviewed?

Newtech66:
I just realised that both `cvxpy` and `numpy` support same kind of indexing that MATLAB does, and in fact `cvxpy` attempts to copy `cvx` exactly. So...what I was doing previously was rather inefficient, and how it is does in MATLAB is a much better option. I changed `_create_constraint` to more close resemble [`eigen_from_order`](https://github.com/nathanieljohnston/QETLAB/blob/f589c35a29e03624ecfd972fd209014799d24862/AbsPPTConstraints.m#L179) from `AbsPPTConstraints.m`.

toqito/state_props/abs_ppt_constraints.py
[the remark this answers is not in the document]
vprusso: Thanks for the input and iteration on this, @Newtech66 -- I appreciate your attention to detail and showcasing your thought process! 

I think what you said here makes sense to me, and I believe what you propose to be a sensible course of action. Retaining the API of `IsAbsPPT` sounds like a plan to me!

vprusso:
> @vprusso I wrote the solver. Check it out:

Very cool, @Newtech66 ! This is getting quite close once the merge conflicts have been resolved!

Newtech66:
Added `cvxpy` support for `is_abs_ppt`. Also updated the solver above.

docs/refs.bib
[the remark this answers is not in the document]
vprusso: Ah, fair enough. Let's keep it true to the source then.

vprusso (approved):
This LGTM! Unless @purva-thakre has anything to add, I'd be happy to put this into the project! Thank you again for your contribution and sticking with things after the official hackathon end. Thank you!

purva-thakre (approved):

PR #1281, opened 2025-07-05
Fix NPA hierarchy constraint generation and support for extended levels like "1+ab"

Wilson123d:
## Description
This PR improves the `npa_constraints` function to correctly support symbolic hierarchy levels like `"1+ab"` and fixes issues with inconsistent or incorrect upper bounds observed in earlier implementations.

The changes ensure that the symbolic operator words are properly reduced, the moment matrix is correctly constructed, and dependent outcome constraints are applied to fully determine the assemblage.

## Changes

Added support for mixed NPA hierarchy levels such as `"1+ab"`

Enforced dependent outcome reconstruction constraints for Alice and Bob

Ensured block reuse in the moment matrix when operator products reduce to the same word

Verified normalization, positivity, and no-signaling constraints

## Results
Tested on the 4-MUB extended nonlocal game. Results match literature expectations:

Classical value: 0.654508

Quantum lower bound: 0.660986

NPA upper bound (`k=1`): 0.760573

NPA upper bound (`k='1+ab'`): 0.670397

No-signaling value: 0.788675

## Checklist
Fixed incorrect NPA bounds for symbolic levels

All constraints pass unit tests

Verified results on 4-MUB game

### DIFF
--- a/toqito/nonlocal_games/tests/test_extended_nonlocal_game.py
+++ b/toqito/nonlocal_games/tests/test_extended_nonlocal_game.py
@@ -484,7 +484,7 @@ class TestExtendedNonlocalGame(unittest.TestCase):
                 ket = B[x][a]
                 pred_mat[:, :, a, a, x, x] = np.outer(ket, ket.conj())
         return pi, pred_mat
-    
+
     def test_four_mub_unentangled_value(self):
         """Unentangled (classical) value of the 4-MUB game."""
         pi, pred_mat = self.four_mub_game()


toqito/nonlocal_games/tests/Game4MUB.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
```
vprusso, line 1: Could we add these tests to the existing `test_extended_nonlocal_games.py` file? 
https://github.com/vprusso/toqito/blob/master/toqito/nonlocal_games/tests/test_extended_nonlocal_game.py

toqito/nonlocal_games/extended_nonlocal_game.py
```diff
@@ -533,33 +533,26 @@ def commuting_measurement_value_upper_bound(self, k: int | str = 1) -> float:
...
+            for y in range(B_in):
+                for a in range(A_out):
+                    for b in range(B_out):
+                        P_ref = self.pred_mat[:, :, a, b, x, y]
+                        blk = K[(x, y)][a*dR:(a+1)*dR, b*dR:(b+1)*dR]
+                        total_win += self.prob_mat[x, y] * cvxpy.trace(P_ref.conj().T @ blk)
+        cons = npa_constraints(K, k, referee_dim=dR)
+        prob = cvxpy.Problem(cvxpy.Maximize(cvxpy.real(total_win)), cons)
+        cs_val = prob.solve(
+            solver=cvxpy.SCS,
```
vprusso, line 552: Should we instead provide these as defaults to the function arguments? It would likely be confusing to have the user provide an argument through the function only to have it overwritten in this call. Perhaps we can replace the default arguments with these values?

toqito/nonlocal_games/extended_nonlocal_game.py
```diff
@@ -533,33 +533,26 @@ def commuting_measurement_value_upper_bound(self, k: int | str = 1) -> float:
...
+        for x in range(A_in):
+            for y in range(B_in):
+                K[(x, y)] = cvxpy.Variable((A_out * dR, B_out * dR), hermitian=True, name=f"K({x},{y})")
+        total_win = cvxpy.Constant(0)
+        for x in range(A_in):
+            for y in range(B_in):
+                for a in range(A_out):
+                    for b in range(B_out):
+                        P_ref = self.pred_mat[:, :, a, b, x, y]
+                        blk = K[(x, y)][a*dR:(a+1)*dR, b*dR:(b+1)*dR]
```
vprusso, line 547: [nit]:
```suggestion
                        blk = K[(x, y)][a * dR:(a+1) * dR, b * dR:(b+1) * dR]
```

toqito/state_opt/npa_hierarchy.py
```diff
@@ -2,436 +2,131 @@
...
-
-
-    :param assemblage: The commuting measurement assemblage operator.
-    :param k: The level of the NPA hierarchy to use (default=1).
-    :param referee_dim: The dimension of the referee's quantum system (default=1).
-    :return: A list of cvxpy constraints.
-
-    """
-    a_out, a_in, b_out, b_in = _get_nonlocal_game_params(assemblage, referee_dim)
+    return assemblage[0,0].shape[0]//dR, a_in+1, assemblage[0,0].shape[1]//dR, b_in+1
```
vprusso, line 72:
```suggestion
    return assemblage[0,0].shape[0] // dR, a_in + 1, assemblage[0,0].shape[1] // dR, b_in + 1
```

vprusso (commented):
Thanks for the MR, @Wilson123d! I've added a couple of comments. 

I might also ping @poig on this, as I know he worked quite a bit on adapting some of this code. It does seem like your fix resolves the issue we were previously seeing, but if he is able to take a peek as well, his feedback would be most welcome. 

Looks like there are also some test failures that would need to be resolved as well, @Wilson123d:
https://github.com/vprusso/toqito/actions/runs/16085495652/job/45403432806?pr=1281

poig:
Thanks for the ping, @vprusso.
Great work on this PR, @Wilson123d! This is a fantastic and much-needed fix. My thoughts align closely with what's already been said.

The new implementation in ``toqito/state_opt/npa_hierarchy.py`` is much more concise. It would be great to add a few comments to explain the core logic, which will help with long-term maintenance. I also noticed the function name confusion (``_get_params`` vs ``_get_nonlocal_game_params``) that needs to be resolved.

For toqito/nonlocal_games/extended_nonlocal_game.py and the new tests, I agree completely with @vprusso's comments. Once those changes are made, this will be in excellent shape.

vprusso:
Looks like there are still some test and lint failures on the pipeline, @Wilson123d https://github.com/vprusso/toqito/actions/runs/16108511632/job/45473485959?pr=1281

Also, you should be able to run these locally if you'd prefer to do that as that's likely easier to do!

PR #1282, opened 2025-07-05
Add example for Antidistinguishability of Circulant States and the Eigenvalue Criterion

CreativeBinBag:
This PR adds a new example to the documentation demonstrating the antidistinguishability of circulant states.

**What it does:**
The script verifies a key result from Johnston et al. (`arXiv:2311.17047`, Thm 5.1) which provides an exact "if and only if" condition for when a set of circulant states is antidistinguishable.

**How it does it:**
1.  Creates a random `5x5` circulant Gram matrix using `toqito.rand.random_circulant_gram_matrix`.
2.  Computes the eigenvalues of the matrix and applies the theorem's inequality (`√λ₀ ≤ Σ√λⱼ`) to predict whether the states are antidistinguishable.
3.  Solves the fundamental state exclusion semidefinite program (SDP) using `toqito.state_opt.state_exclusion` to find the "ground truth" answer.
4.  Confirms that the analytical prediction from the theorem matches the numerical result from the SDP, thus validating the theorem.

### DIFF
--- a/examples/involvingstates/example_circulant_states.py
+++ b/examples/involvingstates/example_circulant_states.py
@@ -1,0 +1,121 @@
+"""
+Antidistinguishability of Circulant States and the Eigenvalue Criterion
+=======================================================================
+
+In this tutorial, we investigate the antidistinguishability of a special
+class of quantum states known as circulant states. We will numerically verify a
+powerful necessary and sufficient condition based on the eigenvalues of the
+states' Gram matrix, as presented in the paper by Johnston et al.
+:footcite:`Johnston_2025_Tight`.
+
+This tutorial builds upon the concepts introduced in the
+:ref:`sphx_glr_auto_examples_involvingstates_example_state_exclusion.py` tutorial.
+"""
+
+# %%
+# Eigenvalue Criterion for Circulant States
+# ---------------------------------------------
+#
+# A set of :math:`n` pure states is called *circulant* if its Gram matrix is
+# circulant. A matrix is circulant if each of its rows is a cyclic shift of the
+# row above it. Such sets of states have a high degree of symmetry and appear
+# in various quantum information contexts.
+#
+# A key result from :footcite:`Johnston_2025_Tight` (Theorem 5.1) provides a
+# simple and exact criterion for determining if a circulant set is
+# antidistinguishable, based solely on the eigenvalues of its Gram matrix.
+#
+# The theorem states that a set of :math:`n` states with a circulant Gram
+# matrix :math:`G` is **antidistinguishable if and only if** its eigenvalues
+# :math:`\lambda_0 \ge \lambda_1 \ge \cdots \ge \lambda_{n-1}` satisfy the
+# following inequality:
+#
+# .. math::
+#    \sqrt{\lambda_0} \le \sum_{j=1}^{n-1} \sqrt{\lambda_j}
+#
+# This gives us a direct analytical test that is much more efficient than
+# solving a full semidefinite program (SDP). We can use :code:`|toqito⟩` to
+# verify this equivalence.
+
+# %%
+# Numerical Verification
+# ^^^^^^^^^^^^^^^^^^^^^^
+#
+# Our plan to verify this theorem is as follows:
+#
+# 1.  Generate a random circulant Gram matrix :math:`G` using
+#     :func:`~toqito.rand.random_circulant_gram_matrix`.
+# 2.  Compute its eigenvalues and perform the **analytical check** using the
+#     inequality from the theorem.
+# 3.  Generate the corresponding set of state vectors from :math:`G` using
+#     :func:`~toqito.matrix_ops.vectors_from_gram_matrix`.
+# 4.  Perform a **numerical check** by computing the minimum error probability
+#     of state exclusion using :func:`~toqito.state_opt.state_exclusion`. The
+#     set is antidistinguishable if this probability is zero.
+# 5.  Confirm that the analytical and numerical checks yield the same conclusion.
+
+import numpy as np
+
+from toqito.matrix_ops import vectors_from_gram_matrix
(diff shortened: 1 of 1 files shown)


examples/involvingstates/example_circulant_states.py
```diff
@@ -0,0 +1,121 @@
...
+print(f"  sqrt(λ₀) = {lhs:.4f}")
+print(f"  Σ sqrt(λⱼ) for j>0 = {rhs:.4f}")
+print(f"  Is sqrt(λ₀) <= Σ sqrt(λⱼ)? {analytical_is_ad}")
+print(f"  Conclusion: The set SHOULD BE antidistinguishable: {analytical_is_ad}")
+
+# 3. Generate states from the Gram matrix for the numerical check.
+states = vectors_from_gram_matrix(gram_matrix)
+
+# 4. Perform the numerical check by solving the state exclusion SDP.
+opt_val, _ = state_exclusion(states)
```
vprusso, line 96: You could also use the `toqito.state_props.is_antidistinguishable` module instead of using the state exclusion one which is a bit more direct. If you do this, you would then need to change the import and also the enumerated plan above, but it would likely be somewhat more direct overall. 

Then again, if this builds off of the state exclusion tutorial, you could keep it as is. Either way is okay, but just pointing that out as a possibility.

examples/involvingstates/example_circulant_states.py
```diff
@@ -0,0 +1,121 @@
...
+# The set is antidistinguishable if and only if the optimal error probability is zero.
+numerical_is_ad = np.isclose(opt_val, 0)
+
+print("\nNUMERICAL CHECK (via state exclusion SDP):")
+print(f"  Optimal error probability from SDP: {opt_val:.2e}")
+print(f"  Conclusion: The set IS antidistinguishable: {numerical_is_ad}")
+
+# 5. Verify that both methods agree.
+print("\n------------------------------------------------------")
+print(f"Do the analytical and numerical results agree? {analytical_is_ad == numerical_is_ad}")
```
vprusso, line 106: The `==` equality check is probably okay, but perhaps we can an `np.isclose` if this gives any grief for straight-up equality checking.

vprusso (approved):
A few minor comments, but overall, LGTM! I know @purva-thakre is currently out, but she can likely take a look sometime next week.

examples/involvingstates/example_circulant_states.py
```diff
@@ -0,0 +1,121 @@
+"""
+Antidistinguishability of Circulant States and the Eigenvalue Criterion
+=======================================================================
+
+In this tutorial, we investigate the antidistinguishability of a special
+class of quantum states known as circulant states. We will numerically verify a
+powerful necessary and sufficient condition based on the eigenvalues of the
```
purva-thakre, line 7:
```suggestion
necessary, and sufficient condition based on the eigenvalues of the
```

examples/involvingstates/example_circulant_states.py
```diff
@@ -0,0 +1,121 @@
...
+# %%
+# Eigenvalue Criterion for Circulant States
+# ---------------------------------------------
+#
+# A set of :math:`n` pure states is called *circulant* if its Gram matrix is
+# circulant. A matrix is circulant if each of its rows is a cyclic shift of the
+# row above it. Such sets of states have a high degree of symmetry and appear
+# in various quantum information contexts.
+#
+# A key result from :footcite:`Johnston_2025_Tight` (Theorem 5.1) provides a
```
purva-thakre, line 24:
```suggestion
# A key result from (Theorem 5.1) :footcite:`Johnston_2025_Tight` provides a
```

examples/involvingstates/example_circulant_states.py
```diff
@@ -0,0 +1,121 @@
...
+# verify this equivalence.
+
+# %%
+# Numerical Verification
+# ^^^^^^^^^^^^^^^^^^^^^^
+#
+# Our plan to verify this theorem is as follows:
+#
+# 1.  Generate a random circulant Gram matrix :math:`G` using
+#     :func:`~toqito.rand.random_circulant_gram_matrix`.
```
purva-thakre, line 47:
```suggestion
#     :func:`.random_circulant_gram_matrix`.
```

examples/involvingstates/example_circulant_states.py
```diff
@@ -0,0 +1,121 @@
...
+# 2.  Compute its eigenvalues and perform the **analytical check** using the
+#     inequality from the theorem.
+# 3.  Generate the corresponding set of state vectors from :math:`G` using
+#     :func:`~toqito.matrix_ops.vectors_from_gram_matrix`.
```
purva-thakre, line 51:
```suggestion
#     :func:`.vectors_from_gram_matrix`.
```

examples/involvingstates/example_circulant_states.py
```diff
@@ -0,0 +1,121 @@
...
+# 4.  Perform a **numerical check** by computing the minimum error probability
+#     of state exclusion using :func:`~toqito.state_opt.state_exclusion`. The
```
purva-thakre, line 53:
```suggestion
#     of state exclusion using :func:`.state_exclusion`. The
```

examples/involvingstates/example_circulant_states.py
CreativeBinBag: @vprusso Great suggestion, thanks! I've updated the PR to use is_antidistinguishable as you recommended. It's much cleaner and definitely better showcases the library's API now

examples/involvingstates/example_circulant_states.py
CreativeBinBag: since analytical_is_ad and numerical_is_ad are both booleans now, the direct == check seems safe

PR #1304, opened 2025-08-13
perf: optimising bell state generation

RajdeepAher:
- [x] removed dependency of `toqito.states.bell` from `toqito.states.basis` 
 - [x] able to reduce total time required for execution for `1000` iterations of all the dim from  `~ 1.4 x 10^8 ns` to `~ 0.9 x 10^6 ns`


vprusso (approved):
🚀

PR #1307, opened 2025-08-17
perf: optimising depolarizing channel

RajdeepAher:
We do not need `toqito.states.max_entangled` to calculate `|psi><psi|` instead we can set it up manually ini `O(dim^2)` which is slightly less expensive

### DIFF
--- a/toqito/channels/depolarizing.py
+++ b/toqito/channels/depolarizing.py
@@ -84,11 +84,12 @@ def depolarizing(dim: int, param_p: float = 0) -> np.ndarray:
     :param param_p: Depolarizing probability \(p \) \in [0,1] that mixes the input state
                     with the maximally mixed state. Default 0.
     :return: The Choi matrix of the completely depolarizing channel.
+    :raises ValueError: If `param_p` is outside the interval [0,1].
 
     """
     # Compute the Choi matrix of the depolarizing channel.
     if param_p > 1 or param_p < 0:
-        raise ValueError("the depolarizing probability must be between 0 and 1.")
+        raise ValueError("The depolarizing probability must be between 0 and 1.")
 
     result = np.zeros((dim**2, dim**2), dtype=np.float64)
     np.fill_diagonal(result, (1 - param_p) / dim)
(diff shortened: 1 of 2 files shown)


toqito/channels/depolarizing.py
```diff
@@ -88,7 +86,11 @@ def depolarizing(dim: int, param_p: float = 0) -> np.ndarray:
 
     """
     # Compute the Choi matrix of the depolarizing channel.
+    result = np.zeros((dim**2, dim**2), dtype=np.float64)
```
purva-thakre, line 89: I cannot directly suggest a change but can you please elaborate on what `param_p` is in the docstring? I am talking about this line: `:param param_p: Default 0.`

toqito/channels/depolarizing.py
```diff
@@ -88,7 +86,11 @@ def depolarizing(dim: int, param_p: float = 0) -> np.ndarray:
...
+    np.fill_diagonal(result, (1 - param_p) / dim)
+
```
purva-thakre, line 91: add a check to raise an error if `param_p` is greater than 1

toqito/channels/tests/test_depolarizing.py
```diff
@@ -28,3 +31,10 @@ def test_depolarizing_partially_depolarizing():
 
     bool_mat = np.isclose(expected_res, res)
     np.testing.assert_equal(np.all(bool_mat), True)
+
+
+@pytest.mark.parametrize("param_p", [-0.1, 1.5])
+def test_depolarizing_invalid_param_p(param_p):
+    """Test invalid input to param_p."""
+    with pytest.raises(ValueError, match=re.escape("the depolarizing probability must be between 0 and 1.")):
```
purva-thakre, line 39:
```suggestion
    with pytest.raises(ValueError, match=re.escape("The depolarizing probability must be between 0 and 1.")):
```

vprusso (approved):

purva-thakre (approved):

PR #1311, opened 2025-08-21
improving `partial_trace`

RajdeepAher:
# Changes:

- Cleaned up how `dim` is handled. Instead of guessing with rounding, the code now checks if the matrix size divides evenly and throws clear errors when it doesn’t. https://github.com/vprusso/toqito/pull/1311/commits/4b36254391f0781b48380e0a191f89224c16cc41

- Simplified how `prod_dim_sys` is calculated. Instead of handling single vs multiple elements separately, the new code uses a unified `np.prod` approach for lists/arrays and direct indexing for ints. https://github.com/vprusso/toqito/pull/1311/commits/03288e444f6c4627c664bca869422cdbd11bbb16

- Added `dim` handling logic to `is_trace_preserving.py` and also added `try and expect` block to `is_quantum_channel.py` to deal with the failure of first example [here](https://toqito.readthedocs.io/en/latest/autoapi/channel_props/is_quantum_channel/index.html#module-channel_props.is_quantum_channel) which used to throw `ValueError` as mentioned https://github.com/vprusso/toqito/pull/1311#issuecomment-3211459788

- We did not need the construction of whole `sub_sys_vec` in the code using `sub_sys_vec = prod_dim * np.ones(int(sub_prod)) / sub_prod` as later in the code, only `sub_sys_vec[0]` is used, so we prefer keeping it as a scalar since beginning : https://github.com/vprusso/toqito/pull/1311/commits/9cdb462a513b491e1dee0cd0fae7467ecef73c91

- Replaced manual `range` construction with a clearer numpy expression for diagonal indices: https://github.com/vprusso/toqito/pull/1311/commits/1606b083855a6fb240c085fcfe893be30691f509

- Switched from manual set operations to numpy’s `setdiff1d` and `concatenate`. https://github.com/vprusso/toqito/pull/1311/commits/127fae1bad12bae6bc7e3b6b08accb7d0f5c1010

### DIFF
--- a/toqito/matrix_ops/tests/test_partial_trace.py
+++ b/toqito/matrix_ops/tests/test_partial_trace.py
@@ -72,6 +72,8 @@ test_input_mat7 = np.arange(1, 4097).reshape(64, 64)
         (test_input_mat, np.array([[12, 14], [20, 22]]), 0, None),
         # specify non-zero sys value and default dim value
         (test_input_mat, np.array([[7, 11], [23, 27]]), [1], None),
+        # specify non-zero sys value and single list dim value.
+        (test_input_mat, np.array([[7, 11], [23, 27]]), [1], [2]),
         # specify dim value as int and default sys value
         (test_input_mat, np.array([[34]]), None, 1),
         # specify non-zero sys value and dim value


RajdeepAher:
@vprusso I had a question regarding the current flow of this function. When `input_mat` is a `2x2` matrix and `dim=None`, the logic sets `dim = [1, 2]` by default. While this technically works, it feels a bit unnatural. My thought is that when `dim=None`, it might make more sense to default to `dim = [sqrt(input_mat.shape[0]), sqrt(input_mat.shape[0])]`. I might be missing something here, so I’d be happy to hear your thoughts or suggestions.

vprusso:
> @vprusso I had a question regarding the current flow of this function. When `input_mat` is a `2x2` matrix and `dim=None`, the logic sets `dim = [1, 2]` by default. While this technically works, it feels a bit unnatural. My thought is that when `dim=None`, it might make more sense to default to `dim = [sqrt(input_mat.shape[0]), sqrt(input_mat.shape[0])]`. I might be missing something here, so I’d be happy to hear your thoughts or suggestions.

That is a bit awkward, yes, I agree. I think:

1. Making sure that this change doesn't break something elsewhere in the project and
2. Adding a line or two in the doc-string to the `partial_trace` function would be good to be more explicit about behavior in this case
3. Adding a test to cover this case
4. Checking to see if this is also true for the `partial_transpose` function.

LMK if you think those checks make sense, but I agree with the overall spirit of what you are conveying here!

RajdeepAher:
This is a list of functions that use `partial_trace` internally. We can setup an issue with an objective of checking whether the implementations hold (logically and code wise too) for the following function:

```text
File toqito/channels/tests/test_pauli_channel.py imports partial_trace
File toqito/channel_metrics/channel_fidelity.py imports partial_trace
File toqito/channel_props/is_trace_preserving.py imports partial_trace
File toqito/matrix_props/sk_norm.py imports partial_trace
File toqito/matrix_props/tests/test_majorizes.py imports partial_trace
File toqito/nonlocal_games/quantum_hedging.py imports partial_trace
File toqito/state_opt/optimal_clone.py imports partial_trace
File toqito/state_opt/symmetric_extension_hierarchy.py imports partial_trace
File toqito/state_props/entanglement_of_formation.py imports partial_trace
File toqito/state_props/has_symmetric_extension.py imports partial_trace
File toqito/state_props/is_separable.py imports partial_trace
File toqito/state_props/tests/test_is_separable.py imports partial_trace
```

Issue #1328, opened 2025-09-12
`max_entangled` should check if the input `dim` is valid


purva-thakre:
https://github.com/vprusso/toqito/blob/30a5d3b85c712f1c1bec1c05b213a684e1852020/toqito/states/max_entangled.py#L10

In `max_entangled`, `dim` is supposed to be the dimension of the entangled state. This function does not check if this value is greater than 0. We want to add a check for this, as well as add the appropriate unit test.

NobleCoder69:
I'd like to work on this issue. Please assign it to me.

PoojasPatel013:
Hi there! I've suggested a fix for this.

I opened PR #1365 , which adds the required validation and tests. Please let me know if any changes are needed

PR #1365, opened 2025-11-25
 `Fix: Add validation for 'dim' parameter in max_entangled` #1328

PoojasPatel013:
This PR addresses issue #1328 by adding input validation to the `max_entangled` function. It ensures that the dimension parameter (`dim`) is strictly a positive integer, preventing runtime errors or unexpected behavior with invalid inputs.

**Changes Made**
* **`toqito/states/max_entangled.py`**: Added a conditional check at the beginning of the function to verify that `dim` is a positive integer. It now raises a `ValueError` if `dim <= 0` or if `dim` is not an integer (e.g., floats).
* **`toqito/tests/test_states/test_max_entangled.py`**: Added a new test function `test_max_entangled_invalid_dim` using `pytest.raises` and parametrization to verify that the function correctly rejects negative integers, zero, and floats.

**Verification**
* Ran `pytest toqito/tests/test_states/test_max_entangled.py` locally, and all tests passed (including the new test case).

### DIFF
--- a/toqito/states/max_entangled.py
+++ b/toqito/states/max_entangled.py
@@ -56,7 +56,8 @@ def max_entangled(dim: int, is_sparse: bool = False, is_normalized: bool = True)
     :return: The maximally entangled state of dimension :code:`dim`.
 
     """
-    if not isinstance(dim, int) or dim <= 0:
+    # Allow both standard int and numpy integer types
+    if not isinstance(dim, (int, np.integer)) or dim <= 0:
         raise ValueError("Dimension must be a positive integer.")
 
     norm_factor = 1 / np.sqrt(dim) if is_normalized else 1.0


vprusso (approved):
Minor comment, but LGTM! Thanks for the contribution, @PoojasPatel013 !

vprusso:
> That was my initial assumption as well! However, when I ran the tests with only `int`, the build failed in several internal functions (like `kraus_to_choi`).
> 
> The issue is that `isinstance(np.int64(5), int)` returns `False` in Python. Since many parts of toqito calculate dimensions using NumPy shapes (which return `np.int64`), we need to explicitly allow `np.integer` to support those internal calls without breaking them.

Ah, interesting :) 

Okay, let's keep that in there then, and thank you for checking!

PR #1368, opened 2025-12-04
validate dim in realignment to raise errors for invalid values (#1329)

### LINKED ISSUE #1329: `realignment` should check & raise an error for invalid `dim`
https://github.com/vprusso/toqito/blob/30a5d3b85c712f1c1bec1c05b213a684e1852020/toqito/channels/realignment.py#L9

If `dim` is not greater than 0 or any of the values in `list[int]` are not greater than 0, raise an error as well as add the appropriate unit tests.

Ameeghai:
This PR adds validation for the `dim` parameter in `toqito.channels.realignment`.

- Raises ValueError when dim <= 0.
- Raises TypeError when dim is not an int or list/tuple of ints.
- Keeps existing behaviour when dim is None.

It also adds unit tests that check invalid integer values, negative dimensions in lists, non-int elements, and incorrect types.

 Closes #1329.

### DIFF
--- a/toqito/channels/tests/test_realignment_invalid_dim.py
+++ b/toqito/channels/tests/test_realignment_invalid_dim.py
@@ -1,0 +1,23 @@
+import pytest
+import numpy as np
+from toqito.channels.realignment import realignment
+
+def test_realignment_invalid_dim_zero():
+    mat = np.eye(4)
+    with pytest.raises(ValueError):
+        realignment(mat, dim=0)
+
+def test_realignment_invalid_dim_negative_in_list():
+    mat = np.eye(4)
+    with pytest.raises(ValueError):
+        realignment(mat, dim=[1, -2])
+
+def test_realignment_invalid_dim_non_int_in_list():
+    mat = np.eye(4)
+    with pytest.raises(TypeError):
+        realignment(mat, dim=[1, 2.5])
+
+def test_realignment_invalid_dim_wrong_type():
+    mat = np.eye(4)
+    with pytest.raises(TypeError):
+        realignment(mat, dim="a")


toqito/channels/realignment.py
```diff
@@ -1,10 +1,29 @@
...
 
 from toqito.matrix_ops import partial_transpose
 from toqito.perms import swap
 
+def _validate_dim_param(dim):
+    """Validate dim parameter: must be int > 0 or iterable of ints > 0."""
+    # Accept None (preserve current behaviour)
+    if dim is None:
+        return
+    # Check iterable of ints (exclude str/bytes)
```
vprusso: It seems weird to explicitly exclude these types. Couldn't we instead have some type of "else" catch all if all of these `if` statements fail?

toqito/channels/realignment.py
```diff
@@ -1,10 +1,29 @@
...
 
 import numpy as np
+from collections.abc import Iterable
 
 from toqito.matrix_ops import partial_transpose
 from toqito.perms import swap
 
+def _validate_dim_param(dim):
+    """Validate dim parameter: must be int > 0 or iterable of ints > 0."""
+    # Accept None (preserve current behaviour)
```
vprusso: What is the "current behavior"?

toqito/channels/tests/test_realignment_invalid_dim.py
```diff
@@ -0,0 +1,23 @@
+import pytest
```
vprusso: We don't have to have a test file that deviates from the naming convention of the file it is testing. This breaks project consistency.

vprusso (commented):
Hi @Ameeghai 

Thanks for the contribution. It seems like a lot of this was likely generated from some LLM tool. Given the amount of comments vs. the amount of code, it seems like there are quite a few structural issues here with the way in which this is implemented.

PR #1369, opened 2025-12-07
add non-negative check and unit test

### LINKED ISSUE #1326: Check system dimension is non-negative in `kraus_to_choi`
https://github.com/vprusso/toqito/blob/30a5d3b85c712f1c1bec1c05b213a684e1852020/toqito/channel_ops/kraus_to_choi.py#L10

The `kraus_to_choi` does not check if the input system dimension is non-negative. We want to add a check for this as well as the appropriate unit test.

tibisabau:
## Description
This PR adds input validation to the kraus_to_choi function to ensure that the provided system dimension (sys) is non-negative. This update introduces an explicit check and raises a clear exception when an invalid value is supplied. A corresponding unit test has also been added to confirm that the validation behaves as expected. Closes #1326.

**The code is generated by GitHub Copilot. (Claude Sonnet 4.5)**

## Changes

  -  [x] Add validation in kraus_to_choi to ensure sys is non-negative.
  -  [x] Add a unit test verifying that a negative sys value raises the appropriate error.

### DIFF
--- a/toqito/channel_ops/tests/test_kraus_to_choi.py
+++ b/toqito/channel_ops/tests/test_kraus_to_choi.py
@@ -108,42 +108,34 @@ expected_non_square = np.array(
 
 
 @pytest.mark.parametrize(
-    "test_input, expected",
+    "test_input, expected, sys",
     [
         # Choi matrix of the transpose map is the swap operator.
-        (kraus_ops_transpose, expected_choi_res_transpose),
+        (kraus_ops_transpose, expected_choi_res_transpose, None),
         # As Kraus operators are non-unique, these also should yield the swap operator
-        (kraus_ops_swap_operator_non_unique, expected_choi_res_swap_operator_non_unique),
+        (kraus_ops_swap_operator_non_unique, expected_choi_res_swap_operator_non_unique, None),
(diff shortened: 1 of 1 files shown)


toqito/channel_ops/kraus_to_choi.py
```diff
@@ -61,6 +61,9 @@ def kraus_to_choi(kraus_ops: list[list[np.ndarray]], sys: int = 2) -> np.ndarray
     :return: The corresponding Choi matrix of the provided Kraus operators.
 
     """
+    if sys < 0:
```
vprusso: But this only checks the system and not the dimension, but the message is for boht system and dimension.

toqito/channel_ops/tests/test_kraus_to_choi.py
```diff
@@ -128,3 +128,9 @@ def test_kraus_to_choi(test_input, expected):
     """Test function works as expected for valid inputs."""
     calculated = kraus_to_choi(test_input)
     assert np.isclose(calculated, expected).all()
+
+
```
vprusso: We should add some tests where the `sys` argument supplied is valid (i.e. non-negative).

Issue #1330, opened 2025-09-12
Functions in `toqito/channels` are inconsistent


purva-thakre:
Some functions in `toqito/channels` apply a quantum channel to the input, while some return the quantum channel itself. We want to add an additional option to each function that allows users to choose whether to apply the function or return the quantum channel itself. 

```py
def some_channel_function(......., apply_channel: bool = False, .........):

```
With an addition like the above, if `apply_channel` is `True`, then the function applies the quantum channel to the user's input. Otherwise, it returns the quantum channel only. 

To close this issue, make sure to add the appropriate unit tests too and update the respective documentation.

Rahi-padwal:
Hi @purva-thakre 
Is this issue still open, or is someone already working on it? If it’s available, I’d be happy to help.

Rahi-padwal:
@vprusso I've opened a Pull Request that adds 'apply_channel' flag to 'amplitude_damping.py' and have also updated the docstring.
Could you please review this approach before I apply the same pattern to the remaining channel functions?

PR link : https://github.com/vprusso/toqito/pull/1376

vprusso:
Re-scoping this issue based on discussion in #1376 and offline thinking.

**The inconsistency is real** — three patterns currently exist in `toqito/channels/`:
- `input_mat | None` fast path (`amplitude_damping`, `bitflip`, `phase_damping`).
- Both `input_mat` and `return_kraus_ops` flags (`pauli_channel`).
- Return the channel only (`choi`, `dephasing`, `depolarizing`, `reduction`).

But adding `apply_channel: bool` to every function doesn't fix the underlying issue — it just renames it. Each function would still return different types depending on a flag, the flag would be coupled to `input_mat`, and it duplicates the existing `toqito.channel_ops.apply_channel` utility.

**New plan:** each channel function in `toqito/channels/` does exactly one thing — returns its canonical channel representation. Applications route through `apply_channel(channel_fn(...), state)`. Kraus extraction from a Choi-returning function goes through `choi_to_kraus`.

PR #1539 implements phase 1: `input_mat` (and `return_kraus_ops` for `pauli_channel`) are kept for backward compatibility but emit `DeprecationWarning` directing users to the new pattern. After one or two minor versions the deprecated kwargs will be removed.

Unifying Kraus-vs-Choi canonical representations across the module is a separate design decision that can be tackled later if needed; practically, `apply_channel` already accepts both so the pain is minimal.

[...]

PR #1376, opened 2025-12-31
Add apply_channel option to amplitude_damping

Rahi-padwal:
## Description
This Pull Request adds 'apply_channel' flag to 'amplitude_damping'.
-if 'apply_channel = True' function applies the quantum channel.
- Docstring updated accordingly.

Related to #1330

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,5 @@ build/
 coverage.xml
 jupyter_execute/
 venv/*
+.venv/
 docs/auto_examples/
--- a/toqito/channels/amplitude_damping.py
+++ b/toqito/channels/amplitude_damping.py
@@ -7,6 +7,7 @@ def amplitude_damping(
     input_mat: np.ndarray | None = None,
     gamma: float = 0,
     prob: float = 1,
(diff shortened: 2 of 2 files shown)


vprusso (commented):
Hi @Rahi-padwal . First off, thank you for the MR! I think the Co-Pilot bot had some good comments, but I'm not sure that any of them can be "easily" dealt with. My sense is that perhaps the initial issue is one that we don't actually want and that putting in the `apply_channel` convention as a convenience actually causes more confusion than not. Maybe it's best we close out the issue and subsequently the MR as well?

vprusso:
@Rahi-padwal thanks for starting on this! The issue (#1330) has been re-scoped away from per-function `apply_channel` flags toward deprecating the `input_mat` convenience path and routing applications through the existing `toqito.channel_ops.apply_channel` utility. PR #1539 implements that direction. Sorry for the mid-flight redirect — see [this comment](https://github.com/vprusso/toqito/issues/1330#issuecomment-4297730008) for the reasoning.

PR #1390, opened 2026-01-23
parameterize random_psd_operator with wishart option

### LINKED ISSUE #1275: Eigenvalue spread of `random_psd_operator` is very skewed
## The issue
The eigenvalue spread of a PSD matrix generated by `random_psd_operator` has a strange distribution that I didn't expect:

```python
from toqito.rand import random_psd_operator
import matplotlib.pyplot as plt
import seaborn as sns

eigs_rpsd = []
for _ in range(10):
    eigs_rpsd.extend(eigvalsh(random_psd_operator(2 ** 6, is_real=True)).tolist())
plt.title('random_psd_operator(N)')
ax = sns.histplot(data=eigs_rpsd, kde=True)
ax.lines[0].set_color('crimson')
plt.xlabel('Eigenvalues')
plt.show()
```
Image

## Expected behavior
I'm not sure what I should expect really: when we say "random PSD operator", we have to define what random really means. Perhaps drawing from a probability measure over PSD operators, like `scipy.stats.wishart`? In my case, I thought the eigenvalues would be uniformly distributed.

Wishart has the following distribution for reference:
```python
from scipy.stats import wishart
import matplotlib.pyplot as plt
import seaborn as sns

eigs_wishart = []
for _ in range(10):
    eigs_wishart.extend(eigvalsh(wishart.rvs(2 ** 8, np.eye(2 ** 8) / 2 ** 8)).tolist())
plt.title('Wishart(N, I / N)')
ax = sns.histplot(data=eigs_wishart, kde=True)
ax.lines[0].set_color('crimson')
plt.xlabel('Eigenvalues')
plt.show()
```
Image

aman-coder03:
## Description

closes #1275

this PR parameterizes the distribution used in `random_psd_operator` by introducing a new `distribution` argument. This allows users to select alternative sampling strategies when generating random positive semidefinite (PSD) operators
the default behavior remains unchanged to preserve backward compatibility

## Changes

- [x] Added `distribution: str = "uniform"` parameter to `random_psd_operator`
- [x] Preserved existing behavior as the default (`"uniform"`) for full backward compatibility
- [x] Added `"wishart"` option using Gaussian sampling (`X @ X.conj().T`) to generate PSD operators
- [x] Added validation for invalid distribution values
- [x] Added unit tests for Wishart sampling and invalid distribution handling

## Checklist

- [x] Use `ruff` for errors related to code style and formatting
- [x] Verify all previous and newly added unit tests pass in `pytest` (targeted tests for modified module pass locally)
- [x] Check the documentation build does not lead to any failures


toqito/rand/random_psd_operator.py
vprusso:
```suggestion
    from toqito.matrix_props import is_positive_semidefinite
    
    is_positive_semidefinite(real_psd_mat)
```

toqito/rand/random_psd_operator.py
```diff
@@ -77,21 +80,55 @@ def random_psd_operator(
...
+
+    if distribution == "uniform":
+        rand_mat = gen.random((dim, dim))
+
+        if not is_real:
+            rand_mat = rand_mat + 1j * gen.random((dim, dim))
+
+        rand_mat = (rand_mat.conj().T + rand_mat) / 2
+        eigenvals, eigenvecs = np.linalg.eigh(rand_mat)
+        Q, _ = np.linalg.qr(eigenvecs)
```
vprusso: Upper case letters are not appropriate for variable names according to PEP-8

vprusso (changes requested):
Thanks for the contribution! The core implementation looks solid and well-tested. A few things to address before we can merge:

## Blocking

**1. Docstring format regression**

The PR uses Sphinx `:code:` and `:math:` directives, but the codebase has migrated to mkdocs/markdown. Compare with `random_density_matrix.py` which uses backticks for inline code (`` `dim` ``) and `\(\)` for inline math. The `:code:` and `:math:` syntax will render as raw text in the docs.

Fix: replace all `:code:\`"..."\`` with `` `"..."` `` and `:math:\text{dim}=2` with `\(\text{dim}=2\)`.

**2. Docstring examples: redundant imports and missing blank lines**

The existing style (see `random_density_matrix.py`) puts blank lines between import, call, and print within each code block, and avoids re-importing modules already available in the same `session`. The Wishart example re-imports `random_psd_operator` despite being in `session="psd_operator"` where it's already imported.

## Should Fix

**3. Empty `References` section**

```python
    References:
        .. footbibliography::
```

This is a Sphinx directive that does nothing in mkdocs. Either add an actual citation to the Wishart distribution using the project's `[@citation_key]` / `refs.bib` system, or remove the section entirely.

**4. `num_degrees < dim` produces singular matrices**

When `num_degrees < dim`, the Wishart matrix is guaranteed to be rank-deficient. While technically valid PSD, this is likely unintentional for most users. At minimum, document this behavior in the docstring. A stronger option would be to emit a warning.

**5. Wishart seeded tests should have hardcoded expected values**

The uniform seeded tests assert against hardcoded expected matrices (good -- catches regressions). The Wishart seeded tests only check reproducibility (`matrix == matrix2`) but not against a fixed expected value. A regression that consistently produces wrong output would pass undetected. Please add hardcoded expected matrices for the Wishart seed cases too.

## Nits

**6.** The non-PSD scale matrix test case is a ~140-char single line. Break it across lines for readability.

vprusso (commented):
Thanks for the work on this @aman-coder03. A few issues that need to be addressed before this can be merged:

### Duplicate code fence markers

Lines 54, 62, and 74 each have a duplicate opening fence:
```
        ```python exec="1" source="above" session="psd_operator"
        ```python exec="1" source="above" result="text" session="psd_operator"
```

This will break the docs build. Only one opening fence per code block — remove the duplicate lines (the ones without `result="text"`).

### Args section is duplicated

The `Args` section appears twice in the docstring — once near the top (lines 27-34 in the original) and once in the new code (after line 81). The second one overrides the first, but the first one is now orphaned. Remove the first `Args` block and keep only the comprehensive one at the bottom.

### `dim` validation is a behavior change

Adding `if not isinstance(dim, int) or dim < 1: raise ValueError` changes the existing API contract. Previously `dim=0` or float dims would just produce empty/weird matrices. While the validation is reasonable, it should be called out as a breaking change. It also doesn't match other `rand/` functions which don't validate `dim` this way (e.g., `random_state_vector`, `random_density_matrix`). I'd remove this validation for consistency — let numpy handle bad dims naturally.

### Wishart complex scaling comment

The `/ np.sqrt(2)` scaling comment says it ensures the expected scale matrix is `scale` rather than `2 * scale`. This is correct for the standard complex Wishart definition, but worth adding a brief reference (e.g., "see Goodman (1963) complex Wishart distribution").

### Test `rtol` change

The `rtol=1e-6` change in `assert_allclose` for the seeded tests is fine, but should be explained — why was the default tolerance insufficient?

### Minor
- The `Raises` section in the docstring should list `ValueError` with descriptions of all the conditions that can trigger it (invalid dim, invalid distribution, bad scale, bad num_degrees).

toqito/rand/random_psd_operator.py
```diff
@@ -35,57 +59,98 @@ def random_psd_operator(
...
+        rand_mat = (rand_mat.conj().T + rand_mat) / 2
+        eigenvals, eigenvecs = np.linalg.eigh(rand_mat)
+        q_mat, _ = np.linalg.qr(eigenvecs)
+        return q_mat @ np.diag(np.abs(eigenvals)) @ q_mat.conj().T
+
+    if distribution == "wishart":
+        if scale is None:
+            scale = np.eye(dim)
+        else:
+            if scale.shape != (dim, dim):
```
vprusso, line 127: A caller passing a plain list (e.g., `scale=[[1,0],[0,1]]`) will crash here with `AttributeError: 'list' object has no attribute 'shape'` instead of the clean `ValueError` you're aiming for — and the type annotation `scale: np.ndarray | None` doesn't actually prevent that at runtime.

Suggest coercing first:

```python
else:
    scale = np.asarray(scale)
    if scale.shape != (dim, dim):
        raise ValueError(f"scale must be a {dim}x{dim} matrix, got {scale.shape}.")
    if not is_positive_semidefinite(scale):
        raise ValueError("scale must be a positive semidefinite matrix.")
```

toqito/rand/random_psd_operator.py
```diff
@@ -35,57 +59,98 @@ def random_psd_operator(
...
+                raise ValueError(f"scale must be a {dim}x{dim} matrix, got {scale.shape}.")
+            if not is_positive_semidefinite(scale):
+                raise ValueError("scale must be a positive semidefinite matrix.")
+
+        if num_degrees is None:
+            num_degrees = dim
+        if num_degrees < 1:
```
vprusso: The error message says "must be a positive integer" but you only check the positivity, not the integer-ness. `num_degrees=2.5` slips through here and then crashes inside `np.random.Generator.multivariate_normal` with a less helpful error.

```python
if not isinstance(num_degrees, (int, np.integer)) or num_degrees < 1:
    raise ValueError("num_degrees must be a positive integer.")
```

toqito/rand/random_psd_operator.py
```diff
@@ -35,57 +59,98 @@ def random_psd_operator(
...
+            x_mat = gen.multivariate_normal(np.zeros(dim), scale, size=num_degrees).T
+        else:
+            # Each component is drawn independently with covariance `scale`.
+            # Dividing by sqrt(2) ensures the resulting complex Wishart matrix
+            # x_mat @ x_mat.conj().T has the expected scale matrix `scale`
+            # rather than 2 * scale; see Goodman (1963), "Statistical analysis
+            # based on a certain multivariate complex Gaussian distribution."
+            x_mat = (
+                gen.multivariate_normal(np.zeros(dim), scale, size=num_degrees).T
+                + 1j * gen.multivariate_normal(np.zeros(dim), scale, size=num_degrees).T
```
vprusso, line 154: Minor: this calls `multivariate_normal` twice per invocation. You can generate the complex Wishart sample more directly by drawing a single white complex Gaussian and left-multiplying by the Cholesky factor of `scale`:

```python
L = np.linalg.cholesky(scale)
z = (gen.standard_normal((dim, num_degrees))
     + 1j * gen.standard_normal((dim, num_degrees))) / np.sqrt(2)
x_mat = L @ z
```

Same distribution (`E[x_mat x_mat^H] = num_degrees * scale`), one RNG path, and it keeps the real and complex branches structurally similar. Not blocking — current code is correct.

toqito/rand/random_psd_operator.py
```diff
@@ -35,57 +59,98 @@ def random_psd_operator(
...
+            ) / np.sqrt(2)
+        return x_mat @ x_mat.conj().T
+
+    raise ValueError("Invalid distribution. Supported options are 'uniform' and 'wishart'.")
```
vprusso, line 158: Consider typing `distribution` as `Literal["uniform", "wishart"]` in the signature so type checkers and IDE autocomplete can surface the valid options. The runtime `ValueError` here is still the right safety net for untyped callers.

toqito/rand/tests/test_random_psd_operator.py
```diff
@@ -73,9 +90,98 @@ def test_random_psd_operator(dim, is_real):
...
+def test_random_psd_operator_with_seed(dim, is_real, seed, distribution, expected_mat):
     """Test that random_psd_operator function returns the expected output when seeded."""
-    matrix = random_psd_operator(dim, is_real, seed)
-    assert_allclose(matrix, expected_mat)
+    # rtol=1e-6 used due to floating-point accumulation in Wishart matrix construction.
+    matrix = random_psd_operator(dim, is_real, seed, distribution=distribution)
+    assert_allclose(matrix, expected_mat, rtol=1e-6)
+
+
+def test_random_psd_operator_invalid_distribution():
```
vprusso, line 142: `pytest.raises(ValueError)` without `match=` is too permissive — any `ValueError` raised anywhere in the function will satisfy this assertion, including ones raised for unrelated reasons if the code is later refactored. Please add `match="Supported options"` (or similar) so the test actually pins the intended error path.

```python
with pytest.raises(ValueError, match="Supported options"):
    random_psd_operator(4, distribution="invalid")
```

toqito/rand/tests/test_random_psd_operator.py
```diff
@@ -73,9 +90,98 @@ def test_random_psd_operator(dim, is_real):
...
     ],
 )
-def test_random_psd_operator_with_seed(dim, is_real, seed, expected_mat):
+def test_random_psd_operator_with_seed(dim, is_real, seed, distribution, expected_mat):
     """Test that random_psd_operator function returns the expected output when seeded."""
-    matrix = random_psd_operator(dim, is_real, seed)
-    assert_allclose(matrix, expected_mat)
+    # rtol=1e-6 used due to floating-point accumulation in Wishart matrix construction.
+    matrix = random_psd_operator(dim, is_real, seed, distribution=distribution)
+    assert_allclose(matrix, expected_mat, rtol=1e-6)
```
vprusso, line 139: Nit: the inline comment attributes `rtol=1e-6` to "floating-point accumulation in Wishart matrix construction," but if the same seed + same NumPy version produce the same output on your machine to machine precision, the real reason you need a looser tolerance is more likely cross-platform BLAS/LAPACK nondeterminism. Worth updating the comment so future maintainers understand why the tolerance is loose.

vprusso (commented):
Thanks for picking up #1275 — the overall shape of this is good: backward compatibility is preserved (existing seeded `uniform` expected matrices are unchanged, positional call `random_psd_operator(dim, is_real, seed)` still works), the docstring got a nice cleanup, and the Wishart normalization convention with the `/sqrt(2)` factor and Goodman (1963) citation is correct. A handful of things to address before merging.

**Should fix**

1. **`scale` must survive list/tuple input or the type annotation should forbid it.** `scale: np.ndarray | None` is the declared type, but most users will pass a plain list. The current code does `scale.shape != (dim, dim)` directly, which crashes with `AttributeError` instead of a clean `ValueError`. Add `scale = np.asarray(scale)` as the first line of the `else` branch (or coerce on entry).

2. **`num_degrees` is not checked for being an integer.** A caller passing `num_degrees=2.5` passes the `< 1` check and then crashes deep inside `np.random.Generator.multivariate_normal` with an unhelpful error. Add `isinstance(num_degrees, (int, np.integer))` to the validation.

3. **`test_random_psd_operator_invalid_distribution` doesn't use `match=`.** Right now any `ValueError` anywhere in the function satisfies this test — e.g., if a future refactor makes it raise for a different reason, the test still passes. Please match on `"Supported options"` or similar so the test actually pins the error path.

**Style / clarity**

4. **Complex Wishart draws `multivariate_normal` twice.** This works but is wasteful and a little awkward. A cleaner alternative is to draw once at double width and split, or (standard) Cholesky-factor the scale matrix and multiply a white complex Gaussian:

   ```python
   L = np.linalg.cholesky(scale)  # fall back to eigendecomp if scale is singular
   z = (gen.standard_normal((dim, num_degrees))
        + 1j * gen.standard_normal((dim, num_degrees))) / np.sqrt(2)
   x_mat = L @ z
   ```

   Same distribution, half the MVN calls, and it composes naturally if `scale` is later allowed to be rank-deficient. Not blocking — current code is correct.

5. **`distribution` could be `Literal["uniform", "wishart"]`** for better type-checker hints and autocomplete, rather than bare `str`.

6. **`rtol=1e-6` comment** says "due to floating-point accumulation in Wishart matrix construction" — if the same seed + same code should produce bit-identical output on the same machine, the likely real reason you needed a looser tolerance is cross-platform BLAS nondeterminism. Consider updating the comment to say that, otherwise it reads like the math is intrinsically lossy.

7. **`scale is not None or num_degrees is not None` on the uniform path emits a warning.** Reasonable, but arguably this should be a `ValueError` — silently ignoring parameters the user explicitly passed is a footgun, and since there are only two distributions right now the user's intent is unambiguous. Judgment call; happy either way but worth a thought.

**Out of scope (just flagging)**

8. **The pre-existing uniform path does `q_mat, _ = np.linalg.qr(eigenvecs)`** on already-unitary eigenvectors from `np.linalg.eigh`, which is redundant — the QR of a unitary matrix is itself up to a diagonal sign. This isn't introduced by your PR, so don't fix it here, but worth a follow-up issue to simplify the uniform branch to `eigenvecs @ diag(|eigvals|) @ eigenvecs.conj().T`.

aman-coder03:
you can have a look now @vprusso

PR #1391, opened 2026-01-23
ensure `partial_trace` raises `ValueError` for invalid parameters

### LINKED ISSUE #1388: ensure error is raised for invalid parameters in `partial_trace`
currently, `partial_trace` does not explicitly validate:
that `input_mat` is square.
that `np.prod(dim)` equals the matrix dimension.
that `sys` indices are within valid subsystem bounds.

in some cases, invalid inputs lead to `NumPy` `IndexError` or unintended behavior rather than a clear `ValueError`.
i propose adding explicit validation checks along with corresponding unit tests for invalid inputs.

aman-coder03:
## Description

closes #1388

adds explicit validation checks to `partial_trace` to ensure that invalid parameters raise clear and consistent `ValueError`.
specifically, this PR ensures:
`input_mat` must be square.
`np.prod(dim)` must match the matrix dimension.
`sys` indices must be within valid subsystem bounds.

previously, some invalid inputs could result in NumPy `IndexError` or unintended behavior instead of a clear `ValueError`.
corresponding unit tests have been added to maintain coverage for the new validation logic.

## Changes

- [x] added explicit validation for non-square matrices
- [x] added validation for `dim` product mismatch
- [x] added validation for out-of-bounds `sys` indices
- [x] added unit tests covering new validation branches
- [x] maintained full coverage for modified logic

## Checklist

- [x] Use ruff for errors related to code style and formatting.
- [x] Verify all previous and newly added unit tests pass in pytest.
- [x] Check the documentation build does not lead to any failures. Sphinx build can be checked locally for any failures related to your PR


vprusso (approved):
Clean PR -- the validation improvements are exactly the kind of defensive checks this function needed. A few minor notes:

## Minor

**1. Negative `sys` indices are now rejected**

Previously, `sys=-1` would silently work (Python negative indexing → last subsystem). The PR now rejects it. This is the right call for this domain (negative subsystem indices are almost certainly a bug), but worth noting as a behavior change.

**2. Error message for 1D input**

If someone passes a 1D array, the check `input_mat.ndim != 2` triggers with "input_mat must be a square matrix." Technically correct but "must be a 2D square matrix" would be slightly more helpful. Very minor.

**3. Pre-existing test file issue (not your PR)**

Line 14 of the existing test file has a double comma in the parametrize string:
```python
"input_mat, , sys_arg, dim_arg, msg",
```
Not caused by this PR, just FYI.

Looks good to merge once CI passes.

vprusso (commented):
This looks good overall — the validation logic is solid and the tests are comprehensive. A few minor items:

### 1. Changed behavior for non-square matrix error message
The test on line 32 changed the expected error from `"Cannot infer subsystem dimensions directly"` to `"input_mat must be a 2D square matrix"`. This is a behavior change — previously a 3x4 matrix would fall through to the dim-inference logic and fail there. The new explicit check is better, but just noting it's a subtle change in error message for existing users.

### 2. The `dim` type broadening is good
Accepting `tuple` and `np.ndarray` in addition to `list` (line 140) is a nice improvement — users often pass numpy arrays or tuples for dimensions.

### 3. The `prod_dim != n` check
Good addition at line 149. Previously this would silently produce wrong results for mismatched dims.

### 4. Negative index handling
The `sys < 0` check is correct — numpy would wrap negative indices silently, which could produce unexpected results. Good catch.

### Approve with minor suggestion
Consider adding the `dim` type check (`else: raise ValueError`) to the docstring's `Raises` section so users know what input types are accepted. Otherwise this is ready to merge.

aman-coder03:
@vprusso i have addressed all the feedback and the PR is ready for a final review and merge at your convenience. thank you!

toqito/matrix_ops/partial_trace.py
```diff
@@ -26,6 +26,104 @@ def partial_trace(
...
+    .. math::
+        X_{pt, 2} = \begin{pmatrix}
+                    7 & 11 \\
+                    23 & 27
+                 \end{pmatrix}.
+
+    By default, the partial trace function in :code:`|toqito⟩` takes the trace of the second
+    subsystem.
+
+    .. jupyter-execute::
```
vprusso: This entire RST block (roughly lines 32–126) is Sphinx syntax — `.. jupyter-execute::`, `.. math::`, `:code:`, `.. footbibliography::`, `:raises:`, `:param:`, `:return:`. toqito's docs are rendered by mkdocs + mkdocstrings + markdown-exec, not Sphinx, so none of this will render, and it duplicates the original docstring that still follows below starting at the `Args:` section (line 130). `help(partial_trace)` will also print the description twice.

Please remove this RST block entirely — the docstring rewrite is out of scope for #1388 (which is just about validation). If you want to update the docstring later, that should be a separate PR.

toqito/matrix_ops/partial_trace.py
```diff
@@ -26,6 +26,104 @@ def partial_trace(
...
+    :raises ValueError: If the product of :code:`dim` does not match the matrix dimension.
+    :raises ValueError: If any index in :code:`sys` is negative or out of bounds.
+    :raises ValueError: If :code:`sys` is not an :code:`int` or a list of ints.
+    :param input_mat: A square matrix.
+    :param sys: Scalar or vector specifying the size of the subsystems.
+    :param dim: Dimension of the subsystems. If :code:`None`, all dimensions are assumed to be
+                equal. Accepted types are :code:`int`, :code:`list`, :code:`tuple`, or
+                :code:`numpy.ndarray`.
+    :return: The partial trace of matrix :code:`input_mat`.
     subsystems are given by the vector `dim` and the subsystems to take the trace on are
```
vprusso, line 29: Orphan sentence left over from the duplication — this is a continuation of the opening paragraph at line 28–30 that got stranded between the new `:return:` line and the original `Args:` section. Will be cleaned up automatically once the RST block above is removed.

toqito/matrix_ops/partial_trace.py
```diff
@@ -123,35 +221,62 @@ def partial_trace(
...
         dim = np.array([d, d])
     elif isinstance(dim, int):
         if n % dim != 0:
-            raise ValueError("Invalid: If `dim` is a scalar, it must evenly divide matrix dimension.")
+            raise ValueError(
+                "Invalid: If `dim` is a scalar, it must evenly divide matrix dimension."
+            )
         dim = np.array([dim, n // dim])
-    elif isinstance(dim, list):
+    elif isinstance(dim, (list, tuple, np.ndarray)):
```
vprusso, line 146: Good extension — accepting `tuple` and `np.ndarray` is more consistent with the type annotation. Two small things:

1. The function annotation at line 14 declares `dim: int | list[int] | np.ndarray | None` — `tuple` isn't listed there. Please add it (or drop `tuple` from the runtime check).
2. For a numpy input, you now always call `np.array(dim)` even when it's already an `ndarray`. `np.asarray(dim)` avoids the unnecessary copy.

toqito/matrix_ops/partial_trace.py
```diff
@@ -123,35 +221,62 @@ def partial_trace(
...
+        raise ValueError("Product of `dim` must match the dimension of input_mat.")
+
+    # Validate sys indices and compute subsystem product.
     if isinstance(sys, int):
+        if sys < 0 or sys >= num_sys:
+            raise ValueError("Subsystem indices in `sys` are out of bounds.")
         prod_dim_sys = dim[sys]
         sys = np.array([sys])
+
     elif isinstance(sys, (list, np.ndarray)):
```
vprusso: `dim` was just extended to accept `tuple`, but `sys` still only accepts `list`/`ndarray`. A user passing `partial_trace(mat, (0, 1), [2, 2])` will hit the `else` branch and get a `ValueError` — inconsistent with `dim`. Either accept `tuple` here too, or (preferably) standardize on one rule for both parameters and update the annotations to match.

toqito/matrix_ops/partial_trace.py
```diff
@@ -178,4 +303,4 @@ def partial_trace(
     pt_mat = permuted_reshaped_mat[:, :, diag_idx]
     pt_mat = np.sum(pt_mat, axis=2)
 
-    return pt_mat
+    return pt_mat
```
vprusso: The file no longer ends with a newline (`\ No newline at end of file` in the diff). ruff's `W292` will flag this — please re-add the trailing newline.

vprusso (changes requested):
Thanks for tackling #1388 — the validation logic itself is a clear improvement (the early square check in particular replaces a misleading "Cannot infer subsystem dimensions" error with an accurate one). A few things need to be addressed before this can merge.

**Blocking**

1. **The docstring was duplicated into two formats.** The PR inserts a new Sphinx-RST block (`Examples`, `.. jupyter-execute::`, `.. math::`, `:code:`, `:raises:`, `:param:`, `.. footbibliography::`) from lines 32–126 while leaving the original mkdocs-style docstring (`Args:` / `Returns:` / `Examples:` with ` ```python exec="1" ``` ` fences) intact below it. toqito's docs are built with **mkdocs + mkdocstrings + markdown-exec**, not Sphinx — none of the RST directives will render, and `help(partial_trace)` now prints the description twice. The docstring rewrite is also outside the scope of #1388. Please remove the added RST block entirely and keep only the validation changes in this PR. If you want to overhaul the docstring format, that should be a separate PR/discussion.

2. **Orphan text at lines 127–128** (`subsystems are given by the vector \`dim\`…`) is a leftover fragment from the duplication and will appear as garbled floating text in `help()`. Will go away when the RST block is removed.

3. **Trailing newline was removed from `partial_trace.py`** (`\ No newline at end of file` at the end of the diff). This will trip ruff's W292. Please add it back.

**Non-blocking**

4. **Tuple inconsistency between `dim` and `sys`.** `dim` now accepts `tuple`, but `sys` still only accepts `int | list | np.ndarray`. Either accept `tuple` for both, or neither — and update the type annotations (`sys: int | list[int] | None`, `dim: int | list[int] | np.ndarray | None`) to match whichever you choose.

5. **Scope creep in the test file.** `test_partial_trace.py` also fixes an unrelated pre-existing typo in the parametrize id string (`"input_mat, , sys_arg, dim_arg, msg"` → `"input_mat, sys_arg, dim_arg, msg"`). That's a fine fix, but it should be called out in the PR description so reviewers aren't surprised.

6. **`test_dim_list_branches` only asserts `.shape`.** Both parametrized cases (`[2]` and `[2, 2]`) will produce a 2×2 result regardless of whether the partial trace is computed correctly. Consider comparing against a known expected matrix so the test actually guards correctness of those branches.

7. **Consider validating the CVXPY `Variable` path.** The new square check runs after the `isinstance(input_mat, Variable)` recursion, so it's exercised for Variable inputs too — but there's no test for that path. A one-line test passing a non-square `Variable` and asserting `ValueError` would close the loop.

8. **Nit (l. 271):** `any(s < 0 or s >= num_sys for s in sys)` — after converting to `np.asarray` you can do this vectorized (`np.any((arr < 0) | (arr >= num_sys))`), but the current form is fine.

aman-coder03:
@vprusso , addressed all feedback!!

PR #1400, opened 2026-01-28
Fix issue #1159: Add missing 0.5 factor and normalize identity in dia…

### LINKED ISSUE #1159: Ensure usage of `np.identity` is normalized in docstrings
Some examples in the function docstrings use `np.identity` in place of a quantum channel. We want to make sure these are properly normalized wherever needed. 

This was noticed in #1106 

For example, `choi_identity` in [this code block](https://toqito.readthedocs.io/en/latest/autoapi/channel_metrics/diamond_distance/index.html#channel_metrics.diamond_distance.diamond_distance) should be changed to what's shown below to ensure the diamond distance is not greater than 1. 

```py
choi_identity = 0.25* np.identity(2**2)
```

abhayrajjais01:
…mond_distance

- Added missing 0.5 multiplier in diamond_distance implementation to match docstring
- Normalized identity Choi matrices in docstring examples (0.5 * np.identity(2**2))
- Added comprehensive unit tests to verify diamond distance bounds and properties

Fixes #1159

This PR addresses two critical bugs in the `diamond_distance` function that were causing diamond norm calculations to exceed the theoretical bound of 1:

1. **Missing 0.5 factor in implementation**: The docstring states the function returns "half of the completely bounded trace norm" but the implementation was missing the `0.5` multiplier
2. **Unscaled identity terms in examples**: Docstring examples used `np.identity(2**2)` without proper normalization for the identity channel's Choi matrix

### Unit Tests
All new tests verify:
-  Diamond distance of identical channels = 0
-  Diamond distance between CPTP channels ≤ 1
-  Identity Choi matrix is Hermitian, positive semidefinite, and properly normalized

### Manual Verification
Verified all changes by:
- Inspecting modified files
- Confirming mathematical correctness
- Checking consistency with quantum information theory

## Breaking Changes

 The 0.5 factor fix will change all `diamond_distance` return values by a factor of 2. Existing code that calls this function will get different (but now correct) results.

### DIFF
--- a/toqito/channel_metrics/diamond_distance.py
+++ b/toqito/channel_metrics/diamond_distance.py
@@ -8,7 +8,7 @@ def diamond_distance(choi_1: np.ndarray, choi_2: np.ndarray) -> float:
 
     This function is a wrapper around
     :py:func:`~toqito.channel_metrics.completely_bounded_trace_norm.completely_bounded_trace_norm`, in that it returns
-    half of the completely bounded trace norm of the difference of its arguments.
+    the completely bounded trace norm of the difference of its arguments.
 
     .. note::
         This calculation becomes very slow for 4 or more qubits.


abhayrajjais01:
Okay! I've updated the PR to just fix the docstring instead.
Thanks for the guidance!

Issue #1241, opened 2025-06-02
Add `binder` option to the examples gallery


purva-thakre:
`sphinx-gallery` extension has [an experimental feature](https://sphinx-gallery.github.io/stable/configuration.html#generate-binder-links-for-gallery-notebooks-experimental) to add a `binder` button to all `ipnyb` files. This `binder button` opens an interactive jupyter notebook in a person's browser. 

An example is [available](https://geopandas.org/en/stable/gallery/cartopy_convert.html) in the `geopandas` docs where an interactive jupyter notebook is opened after I click on the `launch:binder` button. 

![Image](https://github.com/user-attachments/assets/425c5bd6-cee4-4972-9e09-b3e6ef86ccdf)

![Image](https://github.com/user-attachments/assets/8aae2c09-7c24-46cb-9b6a-21ed4ec26f50)

Once this feature is no longer experimental in `sphinx-gallery`, we want to add a `binder` button to the jupyter notebooks in our examples library.

PR #1421, opened 2026-02-18
Add binder configuration for examples gallery

tmdeveloper007:
This PR solves Issue #1241 

The Binder configuration will add a "Launch Binder" button to each example in the Sphinx-Gallery, allowing users to run the examples interactively in JupyterLab on mybinder.org. The button will use the dependencies specified in the newly created requirements.txt file to set up the environment.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -17,9 +17,10 @@ build:
 
 python:
   install:
-    - requirements: docs/requirements.txt
     - method: pip
       path: .
+      extra_requirements:
+        - docs
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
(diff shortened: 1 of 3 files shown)


vprusso:
The requirements are already specified in the `pyproject.toml` file, so why are we duplicating that in the `requirements.txt`? Additionally, why is `.idx` being added to the `.gitignore` file?

vprusso:
> The requirements.txt is complementary to pyproject.toml, not a duplicate. The pyproject.toml remains the primary project configuration for development and distribution, while binder/requirements.txt is specifically optimized for Binder's environment building process. Kind of like a best practice that can help avoid conflicts.
> 
> Also I am working on a nix environment and .idx is a config file for the environment.
> 
> If you have any questions, please ask. If you can suggest improvements, then it would be great.
> 
> Best wishes.

I don't think it makes sense to have two separate environment specifications. Ideally, we would use one for both use cases. Otherwise, this will introduce the need to maintain and update two separate dependency files which I don't see the value in.

Issue #1403, opened 2026-02-04
Broken equation on state distinguishability tutorial


natestemen:
Code here: https://github.com/vprusso/toqito/blob/74f29df6f948ed6f3886f1878da4cbb4fea1ca93/docs/examples/quantum_states/state_distinguishability.py#L82-L88

is rendering as

<img width="600" height="239" alt="Image" src="https://github.com/user-attachments/assets/bf2abf5f-43fb-41f0-af94-71ced543f47c" />

Live tutorial: https://toqito.readthedocs.io/en/latest/auto_examples/quantum_states/state_distinguishability.html#optimal-probability-of-distinguishing-a-quantum-state

PR #1432, opened 2026-02-24
equation rendering

natestemen:
## Description
I stumbled upon another instance of https://github.com/vprusso/toqito/issues/1403 and then decided to go searching.

## Changes
Aside from correctly rendering a bunch of equations, I don't believe I changed anything!

I did stumble upon the fact that the `toqito.state_opt.symmetric_extension_hierarchy.symmetric_extension_hierarchy` docstring uses `\circledvee` which is not on the [mathjax symbols list](https://docs.mathjax.org/en/latest/input/tex/macros/index.html), so that will need to be updated.


vprusso (approved):
🚀 

Assuming the pipelines pass, this looks good! Thanks, @natestemen !

Issue #1089, opened 2025-03-27
Implement the conditional Rényi entropies


tnemoz:
Note: Unless otherwise specified, all the definitions and properties are taken from [1504.00233](https://arxiv.org/abs/1504.00233).

**Is your feature request related to a problem? Please describe.**
Conditional Rényi entropies gained a lot of traction in quantum cryptography recently. Toqito could benefit from having them implemented.

There are two "main" types of conditional Rényi entropies: the Petz ones and the sandwiched ones, each coming in two flavours.

The Petz-Rényi conditional entropies are defined via the Petz-Rényi divergence, which is defined by, for $\alpha\in(0, 1)\cup(1,+\infty)$ and $\rho$ and $\sigma$ being PSD with $\rho\neq0$ as

$$\overline{D}_{\alpha}(\rho\parallel\sigma)=\frac{1}{\alpha-1}\log_2\left(\frac{\mathrm{Tr}\left[\rho^\alpha\sigma^{1-\alpha}\right]}{\mathrm{Tr}[\rho]}\right)$$

if we either have $\alpha<1\wedge\rho\not\perp\sigma$ or $\ker(\sigma)\subseteq\ker(\rho)$, and is defined as $+\infty$ if neither of these conditions are satisfied (Equation 4.86).

We can then define the "downarrow" Petz-Rényi conditional entropy of a PSD bipartite state $\rho_{AB}$ as

$$\overline{H}^{\downarrow}\_{\alpha}(A|B)\_{\rho\_{AB}}=-\overline{D}\_{\alpha}\left(\rho\_{AB}\parallel I\_A\otimes\rho\_B\right)$$

(Equation 5.16) while the uparrow version is defined as

$$\overline{H}^{\uparrow}\_{\alpha}(A|B)\_{\rho\_{AB}}=\sup\_{\substack{\sigma_B\succeq0\\\\\mathrm{Tr}[\sigma]=1}}-\overline{D}\_{\alpha}\left(\rho\_{AB}\parallel I\_A\otimes\sigma\_B\right)$$

(Equation 5.17). As it turns out, this optimization has a closed-form which will be handy for computing it, namely

[...]

PR #1454, opened 2026-03-14
Add Petz and sandwiched conditional Rényi entropies

arnavk23:
### Description
Towards #1089.

### Changes
- [x] Implements Petz and sandwiched conditional Rényi entropy functions.
- [x] Add parameterized tests for pure, mixed, maximally mixed, entangled, and invalid cases.
- [x] Adds references for key papers in refs.bib.

## Checklist

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -40,7 +40,7 @@ jobs:
           uv run mkdocs build --strict 2>&1 | tee build.log || true
 
       - name: Upload docs artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: docs-preview
           path: docs/site/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ repository = "https://github.com/vprusso/toqito"
 [dependency-groups]
 dev = [
     "pytest==9.0.0",
-    "pytest-cov==6.3.0",
+    "pytest-cov==7.1.0",
     "coverage==7.13.0",
     "distlib==0.4.0",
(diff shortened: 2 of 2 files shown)


arnavk23:
@vprusso Went through the guide,. I think this follows (checked ruff and mypy). Whenever you are free, take a look. Thanks!

arnavk23:
@purva-thakre I understand the concern, and I’ll explain the theory behind what I implemented in my own words.

For the **Petz conditional Rényi entropy**, the implementation is based on the Petz Rényi divergence, where the conditional entropy is defined by comparing the joint state with a reference constructed from the identity on subsystem A and the reduced state on subsystem B. In practice, this means I first compute the partial trace over subsystem A to obtain the marginal state on B. Then, I construct the tensor product of the identity on A with this marginal, and evaluate the trace expression involving matrix powers of both the joint state and this reference operator. Since all matrices involved are positive semidefinite, I compute fractional powers via eigen decomposition for numerical stability. The final value is obtained by applying the appropriate normalization and logarithm.

For the **sandwiched conditional Rényi entropy**, the implementation follows the sandwiched Rényi divergence, which differs structurally by introducing a symmetric “sandwich” of the reference operator around the joint state. As before, I compute the reduced state on subsystem B, take an appropriate fractional power of it and lift it to the joint space using the identity on subsystem A. I then construct an operator by placing the joint state between two copies of this transformed marginal operator. After that, I compute the trace of the resulting operator raised to the required power, again using eigendecomposition to ensure numerical stability. The key distinction from the Petz case is this symmetric sandwiching, which leads to stronger monotonicity properties under quantum operations, and this difference is directly reflected in how the operator is constructed in the implementation.

PR #1544, opened 2026-05-06
Implement the matrix geo mean cones from cvxquad

simsaidan:
First, implements and tests a function that calculates the t-weighted matrix geometric mean under matrix_ops. Then, implements the matrix geo mean hypo cone from cvxquad. Relies on a recursive construction of the constraints with base cases of t = 0, 1, and .5.

Finally, implements and tests the matrix geo mean epi cone. This cone calls the hypo cone with an inverted weight.

Imports all 3 in the init file and adds the matrix geo mean paper to bib

## Changes

  -  [x] Implement the matrix geometric mean
  -  [x] Implement the matrix geometric mean hypograph cone from cvxquad
  - [x] Implement the matrix geometric mean epigraph code from cvxquad

## Checklist


simsaidan:
Hi @vprusso ! Hope you have been well. 

First, I realize this is quite delayed from the original grant timeline; I want to apologize for that, and for the lack of communication on my end.

I had handed this project off to one of Mark's students, but after hearing he took it in a different direction, I took a much more successful stab at the original scope. This initial PR implements the matrix geometric mean along with its associated hypograph and epigraph cone constructions (both of which underly the rest of the CVXQuad functions).

If you’re open to it, I’d really appreciate your feedback.

vprusso (commented):
Hey @simsaidan — great to hear from you, and no apology needed. Really glad you took another swing at this; the structure here is much closer to what we want than the earlier attempt. Overall the recursion in `_matrix_geo_mean_cone_recursion` mirrors CVXQUAD's `op_geo_mean.m` faithfully and the trace-max / trace-min tests are a nice way to validate the cone semantics without re-deriving the geometric mean inside CVXPY.

Below is a full pass — I'd like to see the items in **Blocking** addressed before we merge, but the rest are mostly polish.

## Blocking

### 1. CI is failing on 3.14-ubuntu — `PD_JITTER = 1e-6` is too tight

The hypo/epi cone tests build random Wishart matrices and add a `1e-6 * I` jitter to push them PD before calling `matrix_geo_mean(...)` for the reference. On 3.14-ubuntu that jitter is not enough, and `is_positive_definite` is rejecting the reference call:

```
E           ValueError: The matrices must be positive definite.
toqito/matrix_ops/matrix_geo_mean.py:50: ValueError
```

(You can see the failures in the [check-build run](https://github.com/vprusso/toqito/actions/runs/25420219210/job/74801680833) — every `hermitian=True` case fails on 3.14-ubuntu while passing on every other matrix in the env grid.)

Two ways to fix it without making the test fragile:

- Bump the jitter to `max(1e-4, eps * largest_eig) * I`, or
- Construct the test matrices to be PD by construction (e.g. `A = X @ X.conj().T + I`) instead of relying on a Wishart-plus-tiny-jitter trick.

I'd lean toward the second — the Wishart sampler is overkill when all you actually need are arbitrary PD matrices.

### 2. mypy is being introduced alongside `ty`

The project standardizes on `ty` (Astral) as the single type checker — that's set up in `[tool.ty.environment]`, the `lint` group, and the CI pipeline. This PR adds:

- `mypy==1.15.0` to `[dependency-groups].lint`
- `[tool.mypy]` + `[[tool.mypy.overrides]]` blocks in `pyproject.toml`
- a mypy-specific `# type: ignore[import-untyped]` directive on the scipy import

`ty` already considers that comment unused and emits:

```
warning[unused-ignore-comment]: Unused blanket `type: ignore` directive
 --> toqito/matrix_ops/matrix_geo_mean.py:4:58
```

So we'd be running two type checkers that disagree about a directive on the same line. Could you drop mypy and the `# type: ignore` from this PR? If we want to relax type-checking on scipy specifically, the right place is `[tool.ty]` (or a per-module ignore for `ty`). I'd rather not introduce a parallel type checker as a side effect of a feature PR — happy to revisit standardization separately.

### 3. CVXQUAD attribution missing in source

The PR title ("Implement the matrix geo mean cones from cvxquad") and your comment make clear that this is a port of [CVXQUAD](https://github.com/hfawzi/cvxquad)'s `op_geo_mean.m` / `op_pwr_m.m`. The Fawzi–Saunderson 2015 paper is cited via the bib entry, but the code attribution should appear in the source files themselves — CVXQUAD is BSD-2 licensed, and a one-line note at the top of each new file lets future maintainers track upstream changes:

```python
# Adapted from CVXQUAD (https://github.com/hfawzi/cvxquad), BSD-2-Clause.
# Original implementation by Fawzi, Saunderson, et al.
```

Please add to `matrix_geo_mean_hypo_cone.py` and `matrix_geo_mean_epi_cone.py` (the `matrix_geo_mean.py` formula is from the paper directly, so for that one a paper-only citation is fine).

### 4. Module-level `assert` statements in `test_matrix_geo_mean.py`

```python
A_diag = np.diag([2.0, 4.0])
...
assert is_positive_definite(A_diag)
assert is_positive_definite(B_diag)
assert is_positive_definite(A_sym)
assert is_positive_definite(B_sym)
assert not np.allclose(A_sym @ B_sym, B_sym @ A_sym)
assert is_positive_definite(I_2)
assert is_positive_definite(B_scaled)
```

These run at collection time, not as actual tests — pytest won't report them as named cases, and if `is_positive_definite` ever changes shape they'll fail in a way that masks real failures with a confusing collection error. Either move them into a dedicated `test_fixture_invariants` test, or just delete them (these constants are obviously PD — we don't need to assert it).

## Non-blocking

### 5. `_symmetric_like_variable` is imported across modules

```python
# matrix_geo_mean_epi_cone.py
from toqito.matrix_ops.matrix_geo_mean_hypo_cone import (
    _symmetric_like_variable,
    matrix_geo_mean_hypo_cone,
)
```

Single-underscore names are a "private to this module" convention, so importing one across module boundaries is a code-smell. Two ways out:

- Drop the underscore (call it `symmetric_like_variable`) and re-export it from `matrix_ops` if it'll be used by future cone PRs.
- Move it (and `_is_power_of_two`, `_reduced_fraction_pq`) into a `toqito/matrix_ops/_cone_utils.py` private helper module — both files import from it, no cross-module privacy violation.

I'd take the second option since the rest of CVXQUAD will land in follow-up PRs and they'll want the same helpers.

### 6. Parameter naming `A`, `B`, `T`

Project convention in `matrix_props` / `matrix_ops` is lowercase (`mat`, `rho`, `phi`). The capitalization here matches the math, but it's the only place in the package that does this. For the cone constraint functions I'd accept the math-style names since `T` for the upper bound is genuinely clearer than `t_mat`. For `matrix_geo_mean(A, B, t)` though, `mat_a`/`mat_b` would match the rest of the package.

Up to you, but please pick one and apply it consistently.

### 7. Float equality on `t`

`if t == 0`, `if t == 1`, `if t == 1 / 2` all use exact equality. The recursion stays on power-of-two fractions, so this works for internal calls. But if a user passes `0.5000000000001` from outside, none of the base cases fire and we hit the `_is_power_of_two(q)` branch with `q == 1` for the reduced rational — which depends on `limit_denominator`'s default of 1e6. Probably fine in practice; would be more robust as `math.isclose(t, 0.5)` (or feed `t` through `Fraction(...).limit_denominator()` once at the entry point and then dispatch on `(p, q)` rather than `t`).

### 8. Misc small items

- `int(A.shape[0])` — `cvxpy.Expression.shape` returns a tuple of ints already, the casts are noise.
- `p, q = _reduced_fraction_pq(t)` is computed at the top of the recursion but unused on the base-case branches.
- `_is_power_of_two` test (`test_is_power_of_two_nonpositive_and_powers`) is testing a private helper directly; its behavior is implicitly exercised by the trace-max/min tests. Not wrong, but you can probably drop it.
- `test_matrix_geo_mean_symmetry_extended_t` has `-> None`; the parametrized tests in the same file don't. Pick one style.
- `non_hermitian = np.array([[1.0, 2.0], [0.0, 1.0]])` in `test_matrix_geo_mean_invalid_input` — the comment says "not Hermitian / not PD" but the matrix has eigenvalues `(1, 1)`. It's caught by the Hermitian check inside `is_positive_definite`, which is fine, just relabel the comment so future readers don't trip over "but the eigenvalues are positive".

### 9. Docstring nits

- `matrix_geo_mean` — "If the input matrices are scalars or they commute, then this reduces to A^{1-t} B^t" — true but the "scalars" bit is a degenerate special case of "commute". Drop "scalars" and just say "if `A` and `B` commute".
- `_matrix_geo_mean_cone_recursion` docstring says "the constraints are the A and B are positive semidefinite and A = T" — should be "the constraints are that".
- The `\[ G_t(A, B) = G_t_2(B, A) where t_2 = 1 - t. \]` block at the bottom of the same docstring isn't valid LaTeX (`t_2` needs braces); render it as `G_{t'}(A, B) = G_{1-t}(B, A)` or similar.

---

Take your time on the revisions — once 1–4 land I expect this to merge cleanly. Thanks again for picking this back up.

simsaidan:
Hey @vprusso ! Thank you for the review. I pushed new code which addresses all the blocking and non blocking issues as well as most of the copilot comments. Let me know if I should make any other changes.

vprusso:
Thanks for the fast turnaround @simsaidan — the blocking items all look good. Confirming what I see:

- mypy + `# type: ignore[import-untyped]` removed from `pyproject.toml` and `matrix_geo_mean.py` ✅
- CVXQUAD attribution headers added to both cone files ✅
- `_cone_utils.py` is exactly the shared-helper layout I was hoping for ✅
- module-level asserts gone from `test_matrix_geo_mean.py` ✅
- random matrices are now PD by construction (`X X* + 0.1 * I`), 3.14-ubuntu is green ✅
- the `if p == 0 / p == q / 2 * p == q` dispatch is a much cleaner way to handle the base cases than the float-equality version — nice ✅

Couple of follow-ups before I merge.

## File and function naming

The bigger one. The names `matrix_geo_mean*.py` don't quite fit toqito's conventions:

1. **The `matrix_` prefix is redundant** inside `matrix_ops/`. Look at the rest of the directory: `tensor.py`, `partial_trace.py`, `null_space.py`, `unvec.py`, `perturb_vectors.py`, `vectors_to_gram_matrix.py` — none carry a `matrix_` prefix because the package already implies it. The one file that does (`to_density_matrix.py`) only has it because "density matrix" is the noun. Same in `matrix_props/`: `is_hermitian.py`, `sk_norm.py`, `factor_width.py`, etc.

2. **`geo_mean` is more abbreviated than the rest of the package**. Most public names in toqito spell things out: `vectors_from_gram_matrix`, `nonnegative_rank`, `partial_transpose`. We do have a few short forms (`psd`, `npa`, `bb84`) but they're either widely-recognized initialisms or proper-name protocols. `geo` for `geometric` doesn't have that justification.

Concretely I'd like:

| current | proposed |
|---|---|
| `matrix_geo_mean.py` / `matrix_geo_mean(...)` | `geometric_mean.py` / `geometric_mean(...)` |
| `matrix_geo_mean_hypo_cone.py` / `matrix_geo_mean_hypo_cone(...)` | `geometric_mean_hypo_cone.py` / `geometric_mean_hypo_cone(...)` |
| `matrix_geo_mean_epi_cone.py` / `matrix_geo_mean_epi_cone(...)` | `geometric_mean_epi_cone.py` / `geometric_mean_epi_cone(...)` |

I'd keep `hypo_cone` / `epi_cone` as-is since "hypograph cone" / "epigraph cone" are the technical terms and the abbreviation is universal in convex-analysis literature (cvxquad uses the same).

## One leftover from the previous round

`matrix_geo_mean_epi_cone.py:52` still has the redundant `int()` cast that you cleaned up in the hypo-cone file:

```python
if int(A.shape[0]) != int(A.shape[1]):
    raise ValueError("The matrices must be square.")
```

Drop the casts to match the hypo file (line 143–144).

---

Once the rename + that one cast are done I'll merge. Thanks again for sticking with this.

simsaidan:
@vprusso I removed the cast and changed the file names.

vprusso:
Renames look good and the casts are gone — but the 3.11-ubuntu run is failing with the same kind of error as the original 3.14-ubuntu failure: every `hermitian=True` parametrization is hitting

```
E           ValueError: The matrices must be positive definite.
toqito/matrix_ops/geometric_mean.py:50: ValueError
```

The reason this slipped past the previous round is a pre-existing bug in `toqito/matrix_props/is_positive_definite.py`:

```python
if np.array_equal(mat, mat.conj().T):
```

That uses **bit-exact equality** to test Hermitian-ness, while `is_hermitian` correctly uses `np.allclose(..., rtol=1e-5, atol=1e-8)`. So when `_random_pd_matrix` returns

```python
mat = x @ x.conj().T + 0.1 * np.eye(dim, dtype=complex)
```

the result is mathematically Hermitian but Linux's complex BLAS produces tiny off-diagonal asymmetries (~1e-16) that `np.array_equal` rejects. macOS happens to give bit-symmetric output for `X @ X.conj().T`, which is why those jobs were green.

**Workaround for this PR** — symmetrize at the end of `_random_pd_matrix` in both test files:

```python
def _random_pd_matrix(dim: int, seed: int, *, hermitian: bool) -> np.ndarray:
    """Generate a well-conditioned positive-definite matrix by construction."""
    rng = np.random.default_rng(seed)
    if hermitian:
        x_mat = rng.standard_normal((dim, dim)) + 1j * rng.standard_normal((dim, dim))
    else:
        x_mat = rng.standard_normal((dim, dim))
    mat = x_mat @ x_mat.conj().T + PD_SHIFT * np.eye(dim, dtype=x_mat.dtype)
    return (mat + mat.conj().T) / 2  # force bit-exact Hermitian
```

That makes the matrix bit-exactly Hermitian (averaging out any FP asymmetry), so the `np.array_equal` test in `is_positive_definite` accepts it on every BLAS.

**Separately**, I'll open a follow-up issue for `is_positive_definite` — `np.array_equal` is the wrong tool for floating-point Hermitian-checking and it should call `is_hermitian` (or use `np.allclose`) for consistency with the rest of `matrix_props`. That's not your PR to fix, but it's the underlying reason this is fragile.

With the symmetrize workaround in place I expect the matrix to go fully green and we can merge.

simsaidan:
@vprusso I added the workaround

vprusso:
Merged — thanks @simsaidan, this was a clean port and the iteration loop was great. Looking forward to the rest of CVXQUAD landing on top of these cones.

PR #1547, opened 2026-05-11
Implement trace_mpower and lieb_ando from cvxquad

simsaidan:
Supports both numeric and affine branches. Tests mirror the file from cvxquad.

## Changes

  -  [x] Implement trace_mpower and lieb_ando

## Checklist

  -  [x ] Use `ruff` for errors related to code style and formatting.

### DIFF
--- a/toqito/matrix_ops/__init__.py
+++ b/toqito/matrix_ops/__init__.py
@@ -17,5 +17,5 @@ from toqito.matrix_ops.null_space import null_space
 from toqito.matrix_ops.geometric_mean import geometric_mean
 from toqito.matrix_ops.geometric_mean_epi_cone import geometric_mean_epi_cone
 from toqito.matrix_ops.geometric_mean_hypo_cone import geometric_mean_hypo_cone
-from toqito.matrix_ops.trace_power import trace_power
+from toqito.matrix_ops.trace_matrix_power import trace_matrix_power
 from toqito.matrix_ops.lieb_ando import lieb_ando
--- a/toqito/matrix_ops/__init__.py
+++ b/toqito/matrix_ops/__init__.py
@@ -17,3 +17,5 @@ from toqito.matrix_ops.null_space import null_space
 from toqito.matrix_ops.geometric_mean import geometric_mean
 from toqito.matrix_ops.geometric_mean_epi_cone import geometric_mean_epi_cone
 from toqito.matrix_ops.geometric_mean_hypo_cone import geometric_mean_hypo_cone
+from toqito.matrix_ops.trace_power import trace_power
+from toqito.matrix_ops.lieb_ando import lieb_ando
--- a/toqito/matrix_ops/_cone_utils.py
+++ b/toqito/matrix_ops/_cone_utils.py
@@ -1,8 +1,22 @@
 """Shared internal helpers for matrix geometric-mean cone constraints."""
 
 from fractions import Fraction
+from typing import Any
 
 import cvxpy
+
+
+def _require_2d(mat: Any, name: str) -> None:
+    r"""Raise ``ValueError`` unless ``mat`` is two-dimensional (``ndim == 2``)."""
+    if mat.ndim != 2:
+        raise ValueError(f"{name} must be 2D.")
+
+
+def _require_square_2d(mat: Any, name: str) -> None:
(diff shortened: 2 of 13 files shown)


simsaidan:
Hi @vprusso! Hope you had an enjoyable weekend. Following up the two cones from last week with two functions from cvxquad which rely on those cones. Looking forward to your review!

vprusso:
Thanks @simsaidan — glad to see the follow-up. A few things from my read:

## Naming

Echoing the same concern from #1544: the file/function names here aren't quite as clear as they could be, and there's a title/code mismatch.

- **`trace_power.py` / `trace_power(...)`** — the PR title says `trace_mpower` (matching CVXQUAD) but the file/function is `trace_power`. "Power" by itself is ambiguous (element-wise? scalar power of trace?). The function actually computes `tr(C · A^t)` — the trace of a *matrix* fractional power, optionally weighted by `C`. I'd prefer **`trace_matrix_power`** for clarity. It also resolves the title/file mismatch.
- **`lieb_ando.py` / `lieb_ando(...)`** — this is a proper-noun pattern, which we do use (e.g. `bb84`, `npa`), so I'm OK leaving it, but the docstring buries the formula. Please promote `f(A, B, K, t) = tr(K† A^{1−t} K B^t)` to the very first sentence of the function and module docstring so the name's meaning is immediate.

## Likely correctness issue in `lieb_ando` numeric branch for complex K

The docstring (and the SDP branch) use `K†` (conjugate transpose), but the numeric branch uses plain transpose:

```python
# lieb_ando.py:86  (numeric branch — both mat_a and mat_b are ndarray)
return float(np.real(np.trace(mat_k.T @ a_raised @ mat_k @ b_raised)))
```

Meanwhile the CVXPY branch builds `KvKv = Kvec @ Kvec.conj().T`, which encodes `K†` semantics. For complex `K` these two branches return different values. I checked with a random complex K (n=m=3, t=0.4, Hermitian PSD A, B):

```
K^T  formula:  8.644442
K^†  formula: 44.656174
```

The tests don't catch the gap:

- `test_lieb_ando_numeric_complex_rectangular_k` does exercise the numeric branch with complex K, but the reference `_numeric_lieb_ando_reference` *also* uses `mat_k.T` — they agree trivially, and neither matches the docstring's K† formula.
- `test_lieb_ando_constant_expressions_match_numeric` exercises the SDP branch with complex Hermitian A/B, but with `K = np.eye(dim)` (real). For real K, `K^T = K†`, so the divergence is invisible.

CVXQUAD's MATLAB `lieb_ando` uses `K'`, which is complex-conjugate transpose in MATLAB. So the numeric branch should use `mat_k.conj().T`, the test reference should match, and there should be a regression test with a genuinely complex (not real-identity) K that checks numeric and SDP agree.

## Broken docstring examples

Both files use `np.array([[1, 2], [3, 4]])` as the example matrix, which is **not PSD** — running the example as shown would raise `ValueError: The matrix mat_a must be positive semidefinite.` Please swap for an actual PSD matrix, e.g. `np.array([[2.0, 1.0], [1.0, 2.0]])`.

## Smaller things

- `test_trace_power.py` has a commented-out `_sdp_solve_settings(...)` helper (around the line just before the parametrized SDP tests). Please drop the dead code rather than leaving it.
- Input-type validation is asymmetric: `trace_power` raises `TypeError` if `mat_c` isn't `np.ndarray | None`, but `lieb_ando` doesn't validate `mat_k` type at all (just `.ndim` / `.shape`). I'd lean toward both validating.
- The "must be 2D / must be square" stanza is now duplicated across `geometric_mean.py`, `trace_power.py`, and `lieb_ando.py`. Worth a small helper in `_cone_utils.py` (e.g. `_require_square_2d(mat, name)`) to keep these consistent as you add more cvxquad ports on top.

CI is green and the test coverage is broad — the bones are good. The K† fix and the naming are the items I'd want resolved before merge; the rest are polish.

simsaidan:
Hi @vprusso ! I pushed new code which should address each comment.

vprusso:
Merged — thanks @simsaidan, clean revision and great call adding the genuinely-complex-K SDP-vs-numeric regression test. Looking forward to whatever cvxquad piece is next.

PR #1551, opened 2026-05-14
Implement the last cvxquad cone and trace logm.

simsaidan:
Also, surface the lieb ando equation

## Changes

  -  [x] Implement the operator relative entropy epigraph cone
  -  [x] Implement trace logm
  -  [x] Surface the lieb ando equation

## Checklist

### DIFF
--- a/docs/content/refs.bib
+++ b/docs/content/refs.bib
@@ -496,6 +496,21 @@
   url = {https://arxiv.org/abs/1407.4095},
   publisher = {Springer},
 }
+
+@article{fawzi2017matrixlogarithm,
+  title = {Semidefinite approximations of the matrix logarithm},
+  author = {Fawzi, Hamza and Saunderson, James and Parrilo, Pablo A.},
+  journal = {Foundations of Computational Mathematics},
+  volume = {19},
+  number = {2},
+  pages = {259--296},
+  year = {2019},
+  doi = {10.1007/s10208-018-9385-0},
+  url = {https://arxiv.org/abs/1705.00812},
+  eprint = {1705.00812},
+  archivePrefix = {arXiv},
+  primaryClass = {math.OC},
+}
 % Keys beginning with G
 @article{gisin1996hidden,
   title = {Hidden quantum nonlocality revealed by local filters},
@@ -1370,6 +1385,18 @@
   year = {2013},
   month = {oct},
   pages = {103002},
+}
+
+@article{trefethen2008gauss,
+  author = {Trefethen, L. N.},
+  title = {Is {G}auss quadrature better than {C}lenshaw-{C}urtis?},
+  journal = {SIAM Review},
+  volume = {50},
+  number = {1},
+  pages = {67--87},
+  year = {2008},
+  doi = {10.1137/060659831},
+  url = {https://doi.org/10.1137/060659831},
 }
 
 % Keys beginning with V
--- a/toqito/matrix_ops/__init__.py
+++ b/toqito/matrix_ops/__init__.py
@@ -19,3 +19,6 @@ from toqito.matrix_ops.geometric_mean_epi_cone import geometric_mean_epi_cone
 from toqito.matrix_ops.geometric_mean_hypo_cone import geometric_mean_hypo_cone
 from toqito.matrix_ops.trace_matrix_power import trace_matrix_power
 from toqito.matrix_ops.lieb_ando import lieb_ando
+from toqito.matrix_ops.operator_relative_entropy_epi_cone import (
+    operator_relative_entropy_epi_cone,
+)
--- a/toqito/matrix_ops/__init__.py
+++ b/toqito/matrix_ops/__init__.py
@@ -22,3 +22,4 @@ from toqito.matrix_ops.lieb_ando import lieb_ando
 from toqito.matrix_ops.operator_relative_entropy_epi_cone import (
     operator_relative_entropy_epi_cone,
 )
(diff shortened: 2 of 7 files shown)


simsaidan:
Hi @vprusso ! Back again with a PR which implements the last cone from cvxquad as well as the function that calculates the trace of the log of a matrix. I also made the lieb ando module string change that I missed from the last PR.

Let me know what feedback you have!

vprusso:
Thanks for closing out the CVXQUAD cone series with this PR — really nice to have `op_rel_entr_epi_cone` and `trace_logm` in the same shape as the earlier ones, and the Gauss-Legendre / Gauss-Radau scaffolding is clean. The test parameter grid (dim × m,k × apx × hermitian) is thorough and pinning TAU to `A^{1/2} log(A^{1/2} B^{-1} A^{1/2}) A^{1/2}` via `_d_op_reference` is exactly the right reference to compare against.

A few things to fix before this goes in — the math one matters, the rest are housekeeping.

### Correctness

1. **`operator_relative_entropy_epi_cone` docstring math is wrong.** The docstring claims the cone enforces

   ```
   X^{-1/2} log(X^{-1/2} Y^{1/2} X^{-1/2}) X^{1/2} ⪯ TAU
   ```

   That's not Hermitian (the outer factors are `X^{-1/2}` on the left and `X^{1/2}` on the right) and the argument inside the log has `Y^{1/2}`, neither of which match CVXQUAD or the test's own reference. The actual cone the code enforces — and what `_d_op_reference` in the test computes — is the standard operator relative entropy

   ```
   X^{1/2} log(X^{1/2} Y^{-1} X^{1/2}) X^{1/2} ⪯ TAU
   ```

   (equivalently `-X^{1/2} log(X^{-1/2} Y X^{-1/2}) X^{1/2}`). I sanity-checked this numerically — the docstring formula gives a non-Hermitian matrix that differs from the reference by a Frobenius norm of ~4.5 on a 3×3 random test. Please fix the docstring to match the implementation; the code itself is correct.

2. **Circular import will reappear.** `trace_matrix_log.py` has

   ```python
   ```

   at module top level. This is exactly the pattern that's currently breaking docs-deploy on master and that PR #1550 fixes for the sibling files (`geometric_mean.py`, `trace_matrix_power.py`, `lieb_ando.py`). When this PR is merged on top of #1550, the new file will reintroduce the same loop (`matrices.pauli → matrix_ops/__init__ → … → matrix_props → … → matrices.pauli`). Two options: rebase on #1550 once it's in, or apply the same fix here — move the import inside the function body with `# noqa: PLC0415`. I'd prefer doing it in this PR so we don't ship the regression even briefly.

### Surface / API

3. **`trace_matrix_log` isn't re-exported from `toqito/matrix_ops/__init__.py`.** `operator_relative_entropy_epi_cone` is added there but `trace_matrix_log` isn't, even though sibling `trace_matrix_power` is. Looks like an oversight — please add the import so users can do `from toqito.matrix_ops import trace_matrix_log`.

4. **`mat_c` should default to `None`** in `trace_matrix_log`. The docstring says "If `mat_c` is not provided, it is set to the identity matrix" but the signature is

   ```python
   def trace_matrix_log(
       mat_x: np.ndarray | cvxpy.Expression,
       mat_c: np.ndarray | None,
       ...
   )
   ```

   with no default, so callers have to write `trace_matrix_log(X, None)`. Add `= None`.

5. **Document what `apx` values mean.** Both new functions have `apx: int = 0` with the docstring line "The approximation of the logarithm to use." Reading the code I can see `apx=0` → Gauss-Legendre (midpoint, two-sided), `apx=1` → Gauss-Radau with right endpoint (upper bound on log), `apx=-1` → left endpoint (lower bound on log) — but a caller shouldn't have to read the source to pick a value. A one-liner like `apx: -1 for a lower bound, 0 for two-sided Gauss-Legendre, 1 for an upper bound on the logarithm` would do it.

### Tests

6. **No happy-path coverage of non-identity `e`.** Every parametrized success test calls `operator_relative_entropy_epi_cone(..., e=np.eye(dim), ...)`, so the whole point of the `e` parameter (project onto an `r`-dimensional subspace, so `TAU` is `r × r` with `r < n`) is only exercised by the `ValueError` tests. Worth adding at least one passing test with `r < n` so the projection path is actually validated.

7. **`apx=-apx` in `trace_matrix_log`** deserves a one-line comment explaining the sign flip — it's the conversion from "epigraph of `D_op(I, X) = -log X`" to "lower bound on `log X`", and without that note it reads like a bug.

### Nits

8. `operator_relative_entropy_epi_cone` checks `if wi == 0: raise ValueError(...)` inside the quadrature loop. The shipped `_gauss_legendre` / `_gauss_radau` can never produce a zero weight, so this branch only exists to cover the corresponding `patch.object(..., _gauss_legendre, return_value=...)` test. Either drop the check (and the test) or pull it outside the loop and frame it as a precondition. Not a blocker.

The `lieb_ando` docstring refactor (moving the equation into the first line) reads well and surfaces the formula nicely.

Once items 1, 2, and 3 are addressed I'm happy to merge.

simsaidan:
@vprusso Thanks for the feedback and sorry for the sloppy docstring. I pushed new code that addresses all 8 points. Let me know if I should change anything else.

vprusso:
Thanks for the quick turnaround on all eight items — the corrected docstring math, the deferred import, the subspace test, and the `apx`/`mat_c` polish all look good. CI is green across the matrix. Merging now — nice work closing out the CVXQUAD cone trilogy.

PR #1552, opened 2026-05-18
implementation of three cvxquad entropy files

simsaidan:
Implements the quantum entropy, quantum conditional entropy, and quantum relative entropy

## Changes

  -  [ ] Change 1

## Checklist

### DIFF
--- a/toqito/matrix_ops/__init__.py
+++ b/toqito/matrix_ops/__init__.py
@@ -23,3 +23,4 @@ from toqito.matrix_ops.operator_relative_entropy_epi_cone import (
     operator_relative_entropy_epi_cone,
 )
 from toqito.matrix_ops.trace_matrix_log import trace_matrix_log
+from toqito.matrix_ops.ln_quantum_entropy import ln_quantum_entropy
--- a/toqito/matrix_ops/ln_quantum_entropy.py
+++ b/toqito/matrix_ops/ln_quantum_entropy.py
@@ -1,0 +1,101 @@
+r"""Quantum (von Neumann) entropy with natural logarithm: \(-\operatorname{tr}(X \log X)\) for PSD \(X\)."""
+
+# Adapted from CVXQUAD (https://github.com/hfawzi/cvxquad), BSD-2-Clause.
+# Original implementation by Fawzi, Saunderson, et al.
+
+import cvxpy
+import numpy as np
+from scipy.linalg import logm
+
+from toqito.matrix_ops._cone_utils import _require_square_2d
+from toqito.matrix_ops.operator_relative_entropy_epi_cone import (
+    operator_relative_entropy_epi_cone,
+)
+
+
+def ln_quantum_entropy(
+    mat_x: np.ndarray | cvxpy.Expression, m: int = 3, k: int = 3, apx: int = 0
+) -> float:
+    r"""Compute the quantum entropy \(-\operatorname{tr}(X \log X)\) for PSD \(X\).
+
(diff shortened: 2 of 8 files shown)


simsaidan:
Hey @vprusso ! Hope you had an enjoyable weekend. I'm opening this PR, which covers the quantum entropy, quantum relative entropy, and quantum conditional entropy. However, there are several design questions that I want to consult you on. 

For the quantum entropy file, I know there is already a von_neumann.py file in toqito. For simplicity, I put my code in its own file, but would it be better to just build on top of the existing one? One main difference between toqito and cxvquad is the base for the logarithm, but I don't think it would be too hard to calculate with ln and then convert the value for bits. 

I looked for relative_entropy and conditional entropy but didn't see them, so they are in their own files. The question about whether to use bits or nats applies to those two files too. Also, I am not sure that the placement of the three files is ideal, nor their name. 

Looking forward to hearing what you think. Also, for context, between the files already merged and these three, 90% of cvxquad has been ported already.

vprusso:
CI failure here is a single ruff lint error in `quantum_relative_entropy.py`:

```
toqito/state_props/quantum_relative_entropy.py:55:5: PLC0415 `import` should be at the top-level of a file
```

The function-body import + `# noqa: PLC0415` was a defensive copy of the pattern from #1551, but **it isn't needed**: the underlying circular import that motivated that pattern was broken by #1550 (merged before this PR), so a normal top-level import works on master today. I verified locally — `from toqito.matrix_props import is_hermitian, is_positive_semidefinite` at module top level imports cleanly with no cycle.

**Fix**: hoist the import to the top of `quantum_relative_entropy.py` and drop the `# noqa`. Same goes for `ln_quantum_entropy.py:44` — that deferred `is_positive_semidefinite` import can also be hoisted to the top. The real cleanup is a normal top-level import; the noqa is just hiding the underlying refactor.

Once that's in, `code-style` should go green (pytest is already passing across the matrix).

simsaidan:
@vprusso I moved the import up, and I'm now seeing a circular import issue in `ln_quantum_entropy`.

vprusso:
Apologies — my "hoist both imports" advice was wrong. There's a real structural cycle on master: `matrix_props/__init__.py` eagerly loads `is_orthonormal` (which had a redundant call into `state_props`) and `is_block_positive` (which top-imports `sk_operator_norm`, dragging `sk_norm` → `state_props` in at package-load time). With your PR also adding `quantum_relative_entropy` to `state_props`, that finally closes the loop back through `matrix_ops.ln_quantum_entropy`.

I've opened #1559 to fix this at the source — drops the redundant `is_mutually_orthogonal` call, makes `sk_operator_norm` lazy in `matrix_props/__init__.py`, and defers the one call-site in `is_block_positive`. After #1559 lands, your PR can keep the top-level matrix_props imports in both files (no `# noqa` needed) and CI should go green.

If you want to unblock yourself before #1559 merges, the tactical option is to revert just `ln_quantum_entropy.py`'s matrix_props import back to a deferred call (with `# noqa: PLC0415`). Leave `quantum_relative_entropy.py` as you have it now. But I'd suggest just waiting on #1559 — should be a quick review.

simsaidan:
@vprusso Sounds good, I will leave as is and wait for the other PR to merge.

vprusso:
#1559 is merged — please rebase on master and CI should go green with the top-level imports as-is.

vprusso:
Thanks for the rebase — CI is green across the full matrix. Merging now.

PR #1558, opened 2026-05-27
Implement the channel relative entropy SDP formulation

simsaidan:
Supports both upper and lower bound estimation and direct value calculation

## Changes

  -  [x] Implement the relative entropy of channels SDP from https://arxiv.org/pdf/2410.16362 

## Checklist


simsaidan:
Hey @vprusso ! Hope you've been well. This PR implements the SDP for the relative entropy of channels. It is not directly related to the cvxquad files, but I had worked on this while I was still a student of Mark's. I recently polished it up and added tests. 

The paper behind this code is https://arxiv.org/pdf/2410.16362. Essentially, the SDP comes from the integral form of the relative entropy, and the implementation optimizes over a grid of points. 

Looking forward to your review!

vprusso (commented):
@simsaidan — apologies for the delay on this one. I'd drafted a review back on June 3 but never sent it; pulling on that thread now.

The Koßmann–Wilde paper isn't the easiest to translate into a clean SDP and the discretized integral grid + δ/γ coefficients look right against §III. The qubit dephasing-vs-depolarizing parametrized test (30 points across `p ∈ [0.01, 0.10]`) is a strong correctness anchor, and the bib entry is well-formed (DOI + arXiv + venue). CI is green across all five Python versions.

Three correctness items I'd want addressed before merging — the third one is a two-in-one fix that's a clear win.

### Correctness

1. **No solver-status check on either SDP.** Both `lower_prob.solve(solver=cvx.SCS, ...)` and `upper_prob.solve(...)` are followed by `.value` reads, but neither inspects `prob.status`. If SCS returns `OPTIMAL_INACCURATE`, `INFEASIBLE`, or fails outright, the subsequent reads on `rho_a.value`, `qs[k].value`, and `x_var.value` produce `None`, and the lower/upper computation then throws a confusing `TypeError` deep in CVXPY. Suggest:

   ```python
   if lower_prob.status not in (cvx.OPTIMAL, cvx.OPTIMAL_INACCURATE):
       raise RuntimeError(f"Lower-bound SDP failed: {lower_prob.status}")
   if lower_prob.status == cvx.OPTIMAL_INACCURATE:
       warnings.warn("Lower-bound SDP returned OPTIMAL_INACCURATE; result may be off.")
   ```

   Same shape for the upper-bound SDP. The existing `warnings.catch_warnings()` block currently silences SCS's "Solution may be inaccurate" message, which is exactly the signal a user would want to see.

2. **`_make_grid` has no guard** against `mu ≤ 0` or `lam ≤ mu`. If the two channels are very close (but not `np.allclose` enough for the early return), `_find_mu` can return values near 0 — or slightly negative from solver imprecision — making `np.sqrt(epsilon * mu * 8)` either zero (infinite loop on the `while curr < lam` check) or `nan` (silent corruption). Either:

   - Add a precondition `if mu <= 0 or lam <= mu: raise ValueError(...)` with a clear message, or
   - Clamp `mu = max(mu, _MIN_FLOOR)` for some small floor and document the behavior.

   Either way the docstring of `channel_relative_entropy` should state the implicit assumption that `channel_1` is not in the kernel of `channel_2` (i.e., the channels are not too close to identical beyond the `np.allclose` early-exit).

3. **Return type is `np.complex128`, not `float`** as the signature claims. The signature is `-> float | tuple[float, float]`, but `lower` and `upper` are computed by manually rebuilding the objective via `cvx.trace(...).value` on complex Hermitian variables and never `.real`-ed. The Hermitian structure means the imaginary part is ~0 numerically, but the return type is still complex.

   Two-in-one fix: instead of recomputing the objective expression by hand on lines ~250-261, just read it from the solver:

   ```python
   lower = lower_prob.value + np.log(lam) + 1 - lam
   upper = upper_prob.value + np.log(lam) + 1 - lam
   ```

   This matches the objective (which already does `cvx.real(...)`), produces a real `float`, and removes ~15 lines of redundant CVXPY trace recomputation. Strict win.

### Smaller items (non-blocking)

- **`hamiltonian` and `energy` should default.** Every test passes `hamiltonian=np.zeros((in_dim, in_dim), dtype=complex)` and `energy=0.0`. A user just wanting the lower bound shouldn't have to construct a zero Hamiltonian. Suggest `hamiltonian: np.ndarray | None = None` (default to `np.zeros((in_dim, in_dim), dtype=complex)` inside the body) and `energy: float = 0.0`.
- **`epsilon_dec` should default.** It's a tuning knob, not a required input. The paper-example test uses `1e-2`, which would be a sensible default.
- **30-parameter slow test** runs `30 × 2 = 60` SDP solves at `epsilon_dec=1e-2`. Trimming to ~5–8 representative `p` values would preserve correctness coverage at much lower CI cost. Both `mean=True` and `mean=False` are called per parametrization, doubling the solver work for no value — the test could compute the mean itself from a single call.

The lifted SDP framing and the qubit-channel regression are the load-bearing parts and they're correct. Items 1–3 are real but small fixes; once those are in I'm happy to merge.

simsaidan:
@vprusso No worries, I've been preoccupied with UnitaryHack lately. I just pushed new code which should address all 6 points you made. Let me know if I should make any other changes!

toqito/channel_metrics/channel_relative_entropy.py
```diff
@@ -0,0 +1,299 @@
...
+    hamiltonian: np.ndarray | None = None,
+    energy: float = 0.0,
+    epsilon_dec: float = 1e-2,
+    mean: bool = False,
+) -> float | tuple[float, float]:
+    r"""Estimate the quantum relative entropy of two channels [@kossmann2025channelrelativeentropy].
+
+    For channels $\mathcal{N}$ and $\mathcal{M}$,
+
+    $$
```
vprusso: Minor consistency nit (not blocking): the docstrings use `$$...$$` / `$...$` for math, whereas toqito's convention (and the contributing guide) is `\[...\]` for display and `\(...\)` inline (used by ~166 source files vs 2 with `$$`). I ran `mkdocs build --strict` and it renders fine, so this is purely for consistency with the rest of the API docs.

toqito/channel_metrics/channel_relative_entropy.py
```diff
@@ -0,0 +1,299 @@
...
+        energy: Energy parameter ``E`` in the upper-bound SDP. Defaults to ``0.0``.
+        epsilon_dec: Grid refinement parameter for the discretized integral.
+            Defaults to ``1e-2``.
+        mean: If ``True``, return the average of the lower and upper bounds; otherwise
+            return ``(lower, upper)``.
+
+    Returns:
+        Either the midpoint of the bounds or the pair ``(lower, upper)``.
+
+    Raises:
```
vprusso, line 207: Per the contributing guide, a new feature's docstring should include an `Examples:` subsection (ideally an executable ```python exec="1" source="above"``` block) and a `References:` subsection. This docstring has Args/Returns/Raises but neither, so the function won't be demonstrated in the rendered gallery. Could you add a short runnable example and a References section citing [@kossmann2025channelrelativeentropy]?

toqito/channel_metrics/channel_relative_entropy.py
```diff
@@ -0,0 +1,299 @@
...
+        sigma: The Choi matrix of the second channel.
+
+    Returns:
+        The starting point $\mu$.
+
+    """
+    mu = cvx.Variable()
+    problem = cvx.Problem(cvx.Maximize(mu), [sigma - mu * rho >> 0])
+    problem.solve(verbose=False)
+    return float(mu.value)
```
vprusso, line 57: `_find_mu` / `_find_lambda` call `problem.solve(...)` with the default solver and don't check status before `float(mu.value)`. If either auxiliary SDP fails (infeasible/unbounded/solver error), `.value` is `None` and this raises a bare `TypeError` rather than the clear `ValueError` the rest of the function gives. Consider checking `problem.status` (and/or `None`) here too.

toqito/channel_metrics/channel_relative_entropy.py
```diff
@@ -0,0 +1,299 @@
...
+                    [
+                        cvx.trace(qs[k] @ (alpha[k] * choi_1 + beta[k] * choi_2))
+                        for k in range(r - 1)
+                    ]
+                )
+            )
+        ),
+        cons,
+    )
+    lower_prob.solve(solver=cvx.SCS, eps=1e-8, verbose=False)
```
vprusso: The solver is hardcoded to `cvx.SCS` with `eps=1e-8` in both SDPs (and the auxiliary solves use the default solver). Most toqito routines expose `solver`/`**kwargs` so callers can pick a solver. Consider threading a `solver` argument through for consistency and so users with MOSEK etc. can use it.

vprusso (commented):
Thanks for this, it's a solid, well-tested contribution. I went through the implementation and the CI, and separately ran the docs build, and it's close to merge-ready. Notes below.

**What I verified**
- All `toqito-pytest` jobs pass, and the `3.12 / ubuntu-latest` job runs with `--runslow`, so the slow correctness tests actually execute, including `test_channel_relative_entropy_paper_example`, which checks the numeric output against the values in the source paper (arXiv 2410.16362). Good validation that the SDP is correct, not just that it runs.
- I ran `mkdocs build --strict` on this branch merged with current `master`. The new module is strict-clean. The build's single strict warning is a pre-existing, unrelated one in `channel_exclusion.py` (the `**kwargs` annotation, already being fixed in #1582), not from this PR. So this PR does not break the docs build.

**Please address before merge**
- Docstring is missing `Examples:` and `References:` subsections. The contributing guide requires both for a new feature; an executable example would also surface it in the rendered gallery. (inline)

**Non-blocking / nice to have**
- `_find_mu` / `_find_lambda` don't check solve status before reading `.value` (None -> `TypeError` instead of a clean error). (inline)
- Solver is hardcoded to `cvx.SCS`; most toqito routines expose `solver`/`**kwargs`. (inline)
- Math delimiters use `$$`/`$` rather than the `\[\]`/`\(\)` convention (renders fine, consistency only). (inline)

simsaidan:
@vprusso I added a new commit adding an example and all the nice-to-haves. I have the inline citation for the Kossman paper, is that enough?

vprusso:
Thanks @simsaidan, this is close. One CI job is red and it's blocking merge:

`toqito-pytest (3.12, ubuntu-latest)` fails in the new test:

```
FAILED toqito/channel_metrics/tests/test_channel_relative_entropy.py::test_bounds_order_for_distinct_channels
assert np.float64(0.6762907317388205) >= np.float64(0.6762907318552069)
```

The two values are equal to about 1e-10, so `assert upper >= lower` is flipping on SDP solver noise. For these channels (`depolarizing(2, 1)` vs `depolarizing(2, 0.2)` with `epsilon_dec=0.2`) the lower and upper bounds essentially coincide, so the strict ordering can't be relied on at full float precision.

Could you add a small absolute tolerance to that assertion, e.g.

```python
assert upper >= lower - 1e-8
```

(or use `pytest.approx`/`np.isclose`)? The other jobs and codecov are green, so once this is tolerant the PR should be good to merge.

simsaidan:
@vprusso i made a change and ci is green now!

vprusso:
Merged, thanks @simsaidan! Nice addition, and thanks for working through all the review points and the CI fix. Appreciate the contribution.

PR #1564, opened 2026-06-03
Added closed-form antidistinguishability criterion for circulant Gram matrices

### LINKED ISSUE #1522: Implement closed-form eigenvalue criterion for antidistinguishability of circulant sets of pure states
The function `is_antidistinguishable` currently works by solving a full SDP via `state_exclusion`. For the important special case of **circulant sets of pure states**, Theorem 5.1 of Johnston, Russo, and Sikora ([arXiv:2501.03987](https://arxiv.org/abs/2501.03987)) provides an eigenvalue criterion that determines antidistinguishability in closed form — no SDP required.

## Background

A set of $n$ pure states $\{|\psi_0\rangle, \ldots, |\psi_{n-1}\rangle\}$ is **circulant** if its Gram matrix $G$ with entries $G_{jk} = \langle \psi_j | \psi_k \rangle$ is a circulant matrix. Circulant Gram matrices are diagonalized by the DFT matrix, so their eigenvalues $\lambda_0, \ldots, \lambda_{n-1}$ can be computed in $O(n \log n)$ time via FFT.

Theorem 5.1 of [arXiv:2501.03987](https://arxiv.org/abs/2501.03987) states that such a set is antidistinguishable if and only if the eigenvalues of $G$ satisfy a specific condition involving the inverse DFT of $1/\lambda_k$. More precisely, the optimal exclusion value has a closed form in terms of these eigenvalues, and antidistinguishability holds iff this value equals zero.

`|toqito⟩` already has the building blocks:
- `is_circulant` checks whether a matrix is circulant
- `random_circulant_gram_matrix` generates random circulant Gram matrices
- `is_antidistinguishable` checks antidistinguishability via SDP
- There is even an [example script](https://github.com/vprusso/toqito/blob/master/docs/content/examples/quantum_states/circulant_states.py) demonstrating Theorem 5.1

What is missing is a **function that implements the closed-form criterion itself**.

## What should be added

A function (suggested name: `is_antidistinguishable_circulant` or similar, in `toqito/state_props/`) that:

1. Takes a set of pure state vectors (or a Gram matrix directly)
2. Verifies the Gram matrix is circulant (or accepts a flag to skip the check)
3. Computes the eigenvalues via FFT
4. Applies the eigenvalue criterion from Theorem 5.1 to determine antidistinguishability
5. Optionally returns the closed-form optimal exclusion probability

### Relevant code

- [`toqito/state_props/is_antidistinguishable.py`](https://github.com/vprusso/toqito/blob/master/toqito/state_props/is_antidistinguishable.py) — the SDP-based check
- [`toqito/matrix_props/is_circulant.py`](https://github.com/vprusso/toqito/blob/master/toqito/matrix_props/is_circulant.py) — circulant check
- [`toqito/rand/random_circulant_gram_matrix.py`](https://github.com/vprusso/toqito/blob/master/toqito/rand/random_circulant_gram_matrix.py) — random circulant Gram matrix generation
- [`docs/content/examples/quantum_states/circulant_states.py`](https://github.com/vprusso/toqito/blob/master/docs/content/examples/quantum_states/circulant_states.py) — existing example using Theorem 5.1
- [arXiv:2501.03987](https://arxiv.org/abs/2501.03987) — the paper containing Theorem 5.1

### Why this is not trivial

The eigenvalue criterion involves:
- Correctly handling the case where some eigenvalues of $G$ are zero (the states span a proper subspace)
- Understanding when the inverse-DFT formula applies vs. when the set is trivially antidistinguishable due to linear dependence
- Numerical tolerances for eigenvalue comparisons (circulant matrices can have eigenvalues very close to zero)

A contributor must read Theorem 5.1 and its proof to understand the edge cases. The existing example script in the docs demonstrates the *usage* but does not implement the criterion as a reusable function.

### Acceptance criteria

- New function computes antidistinguishability of circulant state sets without calling any SDP solver
- Results agree with `is_antidistinguishable` on a battery of test cases (random circulant Gram matrices of various sizes)
- At least one test with a known closed-form value from the paper (e.g., the trine states example)
- Performance test: the closed-form function is faster than the SDP path for $n \geq 10$
- Edge cases handled: linearly dependent states, rank-deficient Gram matrices, non-circulant input (should raise `ValueError`)
- Docstring cites Theorem 5.1 and includes a worked example

### Out of scope

- Extending the criterion to mixed states
- General (non-circulant) closed-form results
- Modifying the SDP-based `is_antidistinguishable` to auto-detect circulant input (fine as a follow-up)

tnemoz:
## Description

Closes #1522. A new file `is_antidistinguishable_circulant.py` has been added, that contains the new `is_antidistinguishable_circulant` function that implements the criterion for antidistinguishability for circulant matrices derived in [arXiv:2311.17047](https://arxiv.org/abs/2311.17047). It accepts both a Gram matrix, or a set of pure quantum states that it transforms into a Gram matrix before checking the aforementioned criterion.

This function accepts a `skip_circulant_check` optional parameter that is `False` by default (which I think is pretty self-explanatory :p).

The eigenvalues are computed using the FFT. Eigenvalues are clipped to be positive, since we know the Gram matrix is PSD. This avoids close-to-zero eigenvalues being negative when taking their square root.

The function returns `True` if either the criterion is satisfied, or if it is _almost_ satisfied, to be consistent with how `is_antidistinguishable` tests for antidistinguishability.

This function is tested
 - for the trine states, which we know are antidistinguishable,
 - against `is_antidistinguishable` for random circulant matrices,
 - for complex-valued Gram matrices,
 - using Example 3.3 of [arXiv:2311.17047](https://arxiv.org/abs/2311.17047).

Invalid inputs are also tested.

## Changes

  -  [x] Adds `toqito/state_props/is_antidistinguishable.py` and `toqito/state_props/tests/test_is_antidistinguishable.py`.

## Checklist

Claude Code was used to help me make `ty` pass.


vprusso (commented):
Thanks for picking this up — the core implementation is tight (~25 lines), and the test selection lands on exactly the right boundary cases: trine (Example 2.1), random circulant Grams regressed against the SDP path, and the `γ`-on-off-diagonal family (which is the Corollary 5.4 boundary at `γ = (n-2)/(n-1)`).

A few items before merging.

### Docstring bugs

1. The eigenvalue-ordering LaTeX has a typo: `\geqlant` → `\geqslant`.
2. The trine subscripts are duplicated — currently `u_1 = ..., u_1 = ..., u_2 = ...`. Should be `u_0`, `u_1`, `u_2` (matching Example 2.1 of the paper).
3. The boundary calculation in the docstring reads `\sqrt{3/2} ≤ \sqrt{1/3} + \sqrt{0}`. The eigenvalues you list one line above are `3/2, 3/2, 0`, so the RHS should be `\sqrt{3/2} + \sqrt{0}`.

### `skip_circulant_check` dropped on the list path

When called with a list of states, the recursive call

```python
return is_antidistinguishable_circulant(
    vectors_to_gram_matrix(cast(...))
)
```

doesn't forward `skip_circulant_check`. A user who passes `skip_circulant_check=True` with a list expects the check to be skipped on the recursion. Not a correctness bug (the answer is still right), just a surprise. Suggest:

```python
return is_antidistinguishable_circulant(
    vectors_to_gram_matrix(states),
    skip_circulant_check=skip_circulant_check,
)
```

### Type-hint scope

The new shape-precise hints (`list[np.ndarray[tuple[int, Literal[1]], np.dtype[np.inexact[Any]]]]`) on `vectors_to_gram_matrix` and `trine` are inconsistent with the rest of `toqito`, which uses `list[np.ndarray]` for public APIs. The ripple effect is that any other callsite that constructs `[v.flatten() for v in trine()]` (or similar) will start tripping `ty` even though it was fine before.

My recommendation: keep the strict shape types inside `is_antidistinguishable_circulant.py` only, and revert `vectors_to_gram_matrix.py` and `trine.py` to their previous signatures. The .astype(float) fix in `trine.py` is genuine and worth keeping. The type-system tightening is best as a separate, project-wide PR.

### Missing per the issue's acceptance criteria

4. **No `skip_circulant_check=True` test** — should verify the bypass path is exercised at least once.
5. **No complex-Gram test.** `random_circulant_gram_matrix` returns *real* symmetric circulant matrices (the docstring explicitly says so: "the circulant matrix... is real due to its origin from the DFT of a real diagonal matrix"). So the entire complex Hermitian-circulant path is uncovered. Worth one test with a genuinely complex circulant set (e.g., `|ψ_k⟩ = U^k |ψ_0⟩` for some non-real unitary `U`).
6. **No performance test.** The issue asked for "the closed-form function is faster than the SDP path for `n ≥ 10`". A short timing assertion (or a `@pytest.mark.benchmark`) would close that bullet.

### Subtle: `ifft` vs `fft`

The eigenvalue computation uses `np.fft.ifft(states[0])`. The standard convention for a circulant matrix `C = scipy.linalg.circulant(c)` is `eigenvalues = np.fft.fft(c)`. These differ by a factor of `1/n` (and, for general complex inputs, an index-reversal that doesn't matter after sorting). The criterion `√λ₀ ≤ Σ √λⱼ` is invariant under positive scaling of all eigenvalues, so the answer is correct, but the choice is unconventional. Either:

- switch to `np.fft.fft(states[0])` (giving the actual eigenvalues), or
- add a one-line comment that the `1/n` scaling is absorbed by the inequality's scale-invariance.

Not a blocker.

---

Once 1–3 are addressed and the type-hint scope is rolled back, I'm happy to merge. 4–6 would be valuable but the closed-form criterion is the load-bearing part and it's correct.

tnemoz:
Hm, I thought using the `benchmark` fixture of `pytest-benchmark` was the thing to do here so that we don't measure the time taken to create the rnadom circulant matrix, but `pytest-benchmark` isn't installed when running the tests (though it is used to run the benchmarks).

What is the best course of actions? Use `@pytest.mark.benchmark` and measure the whole function, or install `pytest-benchmark` before running the tests?

vprusso:
Good question — looking at the repo setup, you've spotted a real friction point. The project's benchmarks actually live in a separate repo ([`vprusso/toqito-bench`](https://github.com/vprusso/toqito-bench)), and the `benchmark.yml` workflow checks that out + installs `pytest-benchmark` only for the benchmark job. `pytest-benchmark` isn't in the main toqito test deps on purpose.

So I'd just drop the performance test from this PR. Two reasons:

1. The existing `test_is_antidistinguishable_random_circulant_gram[dim=10]` test already runs both paths and asserts they agree — that's enough regression coverage to catch any future "oops, I'm calling the SDP" mistake. The implementation is structurally non-SDP (FFT + a sum-of-square-roots inequality), so the "is it faster than the SDP?" question doesn't have a realistic regression path.
2. Adding `pytest-benchmark` to the main test deps for one test would cut against the existing split where benchmarks live elsewhere.

If you want a perf signal eventually, the right home for it is a follow-up in `toqito-bench`, not here. Strike that bullet from this PR.

tnemoz:
Hey @vprusso ! Is there anything else to do in this PR?

vprusso:
Almost there — every code/test item from my earlier review is addressed: `skip_circulant_check` now forwards on the list path, the docstring math is correct (right indices on the trine, `√(3/2) + √0` boundary, `\geqslant`), the complex-Gram test on `{|+⟩, |i⟩, |−⟩, |−i⟩}` covers the Hermitian-circulant path, `skip_circulant_check=True` is parametrized, and the `fft` switch is in. The math + tests are good to merge.

Last thing before I can hit the button: **`uv.lock` has unrelated dependency drift** (269 lines, `cvxpy 1.8.1 → 1.9.0` plus the transitive bumps). That's separate from this feature and shouldn't ride with it. Same fix as on a couple of other PRs this round — please revert just that file to upstream:

```bash
git remote add upstream https://github.com/vprusso/toqito.git   # if you don't have it
git fetch upstream master
git checkout upstream/master -- uv.lock
git commit -m "Revert unrelated uv.lock dep bumps"
git push
```

After that the diff should drop to ~`+~280/-~45` across just the four real files (`is_antidistinguishable_circulant.py`, its test, the `__init__.py` export, and the tiny `trine.py` `.astype(float)` fix). Once that's clean I'll merge.

tnemoz:
@vprusso Done!

vprusso (approved):
Clean — `uv.lock` is back at 0-line diff and the only changes are the four files this PR is about (`is_antidistinguishable_circulant.py`, its test, the export, and the `trine.py` `.astype(float)` fix). Approving. Will merge once the pytest matrix finishes (code-style is already green).

PR #1565, opened 2026-06-03
fix: check hermiticity using is_hermitian instead of np.array_equal

### LINKED ISSUE #1546: is_positive_definite uses bit-exact equality for Hermitian check (BLAS-fragile)
## Summary

`is_positive_definite` checks Hermitian-ness with `np.array_equal(mat, mat.conj().T)` — a bit-exact comparison — while the rest of `matrix_props` uses `np.allclose` (with `rtol=1e-5, atol=1e-8`). This means matrices that are *mathematically* Hermitian but built via floating-point matrix products can be rejected as "not PD" purely because of BLAS-induced off-diagonal asymmetries (~1e-16). Whether this triggers depends on the BLAS implementation: macOS Accelerate produces bit-symmetric output for `X @ X.conj().T`, while Linux OpenBLAS / MKL on some configurations does not.

This was the root cause of the CI flake we hit during #1544 — the contributor had to add `mat = (mat + mat.conj().T) / 2` to test fixtures as a workaround.

## Where

`toqito/matrix_props/is_positive_definite.py:61`

```python
def is_positive_definite(mat: np.ndarray) -> bool:
    if np.array_equal(mat, mat.conj().T):     # <-- bit-exact, wrong tolerance
        try:
            np.linalg.cholesky(mat)
            return True
        except np.linalg.LinAlgError:
            return False
    else:
        return False
```

Compare to `is_hermitian.py`:

```python
def is_hermitian(mat: np.ndarray, rtol: float = 1e-05, atol: float = 1e-08) -> bool:
    return np.allclose(mat, mat.conj().T, rtol=rtol, atol=atol)
```

## Proposed fix

Replace the `np.array_equal` check with `is_hermitian(mat)` so all Hermitian-ness checks in `matrix_props` use the same tolerant comparison:

```python
from toqito.matrix_props.is_hermitian import is_hermitian

def is_positive_definite(mat: np.ndarray) -> bool:
    if not is_hermitian(mat):
        return False
    try:
        np.linalg.cholesky(mat)
        return True
    except np.linalg.LinAlgError:
        return False
```

Optionally surface `rtol`/`atol` parameters on `is_positive_definite` mirroring `is_hermitian` so callers can tighten/loosen as needed.

## Test plan

- Add a regression test that constructs `mat = X @ X.conj().T + 0.1 * I` for complex `X`, asserts `is_positive_definite(mat)` is `True` regardless of BLAS-induced asymmetry.
- Verify existing `is_positive_definite` tests still pass (the tolerance is loose enough that genuinely non-Hermitian matrices still fail).

## Context

Discovered while reviewing #1544 (matrix geometric mean / CVXQUAD cones). See [this comment](https://github.com/vprusso/toqito/pull/1544#issuecomment-4407737889) for the failure trace.

tnemoz:
## Description
Fixes #1546. The `is_positive_definite` function now uses the `is_hermitian`  function instead of `np.array_equal` to deal with potential inaccuracies of BLAS implementations. The new function also has an `rtol` and an `atol` parameters that are passed down to `is_hermitian`.

## Changes

  - [x] `is_positive_definite` now uses `is_hermitian`, with newly added tolerance parameters.
  - [x] A test that doesn't pass the current implementation of `is_positive_definite` but passes with this new implementation has been added.
  - [x] The workarounds for the tests written in #1544 have been removed. This simplifies the code, and ensures that this workaround isn't needed anymore with this new implementation.

## Checklist

I haven't used an AI for this PR, neither for the code, nor for the PR description.

### DIFF
--- a/toqito/cones/tests/test_geometric_mean_epi_cone.py
+++ b/toqito/cones/tests/test_geometric_mean_epi_cone.py
@@ -33,8 +33,7 @@ def _random_pd_matrix(dim: int, seed: int, *, hermitian: bool) -> np.ndarray:
         x_mat = rng.standard_normal((dim, dim)) + 1j * rng.standard_normal((dim, dim))
     else:
         x_mat = rng.standard_normal((dim, dim))
-    mat = x_mat @ x_mat.conj().T + PD_SHIFT * np.eye(dim, dtype=x_mat.dtype)
-    return (mat + mat.conj().T) / 2
+    return x_mat @ x_mat.conj().T + PD_SHIFT * np.eye(dim, dtype=x_mat.dtype)
 
 
 @pytest.mark.parametrize("dim", DIMS)
--- a/toqito/cones/tests/test_geometric_mean_hypo_cone.py
+++ b/toqito/cones/tests/test_geometric_mean_hypo_cone.py
@@ -30,8 +30,7 @@ def _random_pd_matrix(dim: int, seed: int, *, hermitian: bool) -> np.ndarray:
         x_mat = rng.standard_normal((dim, dim)) + 1j * rng.standard_normal((dim, dim))
     else:
         x_mat = rng.standard_normal((dim, dim))
-    mat = x_mat @ x_mat.conj().T + PD_SHIFT * np.eye(dim, dtype=x_mat.dtype)
-    return (mat + mat.conj().T) / 2
+    return x_mat @ x_mat.conj().T + PD_SHIFT * np.eye(dim, dtype=x_mat.dtype)
 
 
 @pytest.mark.parametrize("dim", DIMS)
--- a/toqito/matrix_props/is_positive_definite.py
+++ b/toqito/matrix_props/is_positive_definite.py
@@ -62,14 +62,13 @@ def is_positive_definite(mat: np.ndarray, rtol: float = 1e-05, atol: float = 1e-
             [`is_positive_semidefinite`][toqito.matrix_props.is_positive_semidefinite.is_positive_semidefinite]
 
     """
-    if is_hermitian(mat, rtol, atol):
-        try:
-            # Cholesky decomp requires that the matrix in question is
-            # positive-definite. It will throw an error if this is not the case
-            # that we catch here.
-            np.linalg.cholesky(mat)
-            return True
-        except np.linalg.LinAlgError:
-            return False
-    else:
+    if not is_hermitian(mat, rtol, atol):
+        return False
+    try:
+        # Cholesky decomp requires that the matrix in question is
+        # positive-definite. It will throw an error if this is not the case
+        # that we catch here.
+        np.linalg.cholesky(mat)
+        return True
+    except np.linalg.LinAlgError:
         return False
--- a/toqito/matrix_props/is_positive_definite.py
+++ b/toqito/matrix_props/is_positive_definite.py
@@ -2,12 +2,16 @@
 
 import numpy as np
 
+from toqito.matrix_props import is_hermitian
 
-def is_positive_definite(mat: np.ndarray) -> bool:
+
+def is_positive_definite(mat: np.ndarray, rtol: float = 1e-05, atol: float = 1e-08) -> bool:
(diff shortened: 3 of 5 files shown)


vprusso (commented):
Thanks for picking this up — the core fix is exactly what #1546 asked for: swap `np.array_equal` for `is_hermitian` and surface `rtol`/`atol`. The BLAS-accuracy regression test is the right shape (uses a Linux-like OpenBLAS non-bit-symmetric `X @ X.conj().T` to exercise the bug).

A few things to fix before merging.

### Stale branch — needs a rebase on master

The biggest one: the PR is based off a pre-#1561 master. The two files

- `toqito/matrix_ops/tests/test_geometric_mean_epi_cone.py`
- `toqito/matrix_ops/tests/test_geometric_mean_hypo_cone.py`

no longer exist on master. They were moved to `toqito/cones/tests/` as part of [#1561](https://github.com/vprusso/toqito/pull/1561) (the CVXQUAD subpackage split). Right now, this PR effectively re-creates the old files in their old locations, which (a) doesn't actually remove the workaround you wanted to remove — the surviving copies in `toqito/cones/tests/` still have it — and (b) leaves duplicate test files after merge.

After rebasing on current master, the workaround-removal edits should be applied to:

- `toqito/cones/tests/test_geometric_mean_epi_cone.py`
- `toqito/cones/tests/test_geometric_mean_hypo_cone.py`

(specifically dropping the `mat = ...; return (mat + mat.conj().T) / 2` symmetrization lines, which is the exact fix you intended.)

### `test_is_not_positive_definite` mixes two cases

The modification appends a 4×4 Gram matrix (Example 6.1 from arXiv:2311.17047) into the existing `test_is_not_positive_definite`. Two concerns:

1. **`eps = 0` makes the `gram_eps` arithmetic dead.** With `eps = 0`, `1 / (1 - 2*eps) * (gram + eps * (v v* + w w* - 3I)) = gram`. The `v_vec`, `w_vec`, and `np.identity(4)` constructions are inert. Either remove them and just use `gram`, or move them into a parameterized test that actually varies `eps`.
2. **It's two unrelated assertions in one test.** The `[[-1,-1],[-1,-1]]` matrix and the rank-deficient 4×4 Gram are testing different things (negative-definite-on-real vs. PSD-but-singular-on-complex). Suggest splitting into a separate `test_is_positive_definite_rank_deficient` so the failure mode is clear if either ever regresses.

(I verified your assertion: this `gram` has eigenvalues ≈ {0, 0.382, 1.0, 2.618}, so it's PSD-but-not-PD and the assertion is correct. Just a structural quibble.)

### Tiny code-style nit

The new `is_positive_definite` could flatten the nested else:

```python
if not is_hermitian(mat, rtol, atol):
    return False
try:
    np.linalg.cholesky(mat)
    return True
except np.linalg.LinAlgError:
    return False
```

Same behavior, one less indent level. Not blocking.

---

Once the rebase is in and the workaround-removal lands in the actual `toqito/cones/tests/` paths, this is ready to merge. The other items are polish.

vprusso:
The four file changes look right and CI is green — but there's a `uv.lock` issue blocking merge. Your branch last picked up master on June 4 (before #1562, #1568, #1571, and #1576 landed), so the lockfile carries pre-June-4 state. Specifically, `uv.lock` here has `ipython 9.13.0` whereas current master has `9.14.0` (bumped by #1562). I simulated the squash locally and the working tree ends up with the older ipython, so merging would silently downgrade it.

Same fix you did on #1564:

```bash
git fetch upstream master
git checkout upstream/master -- uv.lock
git commit -m "Revert unrelated uv.lock dep drift"
git push
```

Once that's done the diff will be exactly the four real files (`is_positive_definite.py`, its test, the two cone test workaround removals) and it's ready to merge.

The math/code itself is in good shape — your `is_hermitian` swap with `rtol`/`atol` plumbed through is exactly what #1546 asked for, the BLAS-accuracy regression test is the right shape, the test split (`negative_definite` vs `rank_deficient`) is cleaner, and the cone test workaround removals land at the correct `toqito/cones/tests/` paths now that #1561's reorganization is in master.

PR #1567, opened 2026-06-04
Add state exclusion to the symmetric extension hierarchy SDP

### LINKED ISSUE #1523: Add symmetric extension hierarchy for state exclusion (separable measurement bounds on antidistinguishability)
`toqito.state_opt.symmetric_extension_hierarchy` computes upper bounds on the probability of state **distinguishability** under separable measurements, using the symmetric-extension SDP technique of Navascués ([arXiv:0707.4398](https://arxiv.org/abs/0707.4398), *Pure State Estimation and the Characterization of Entanglement*, Phys. Rev. Lett. **100**, 070503, 2008) as applied to state discrimination in Cosentino's thesis (Equation 4.55 of [*Quantum State Local Distinguishability via Convex Optimization*](https://uwspace.uwaterloo.ca/handle/10012/9572), 2015; see also Cosentino's PRA paper [arXiv:1205.1031](https://arxiv.org/abs/1205.1031) for the related PPT-only SDP). No analogous hierarchy exists for state **exclusion**.

This is a significant gap: the hierarchy for distinguishability gives upper bounds on the SEP value, and at level 1 it recovers the PPT bound exactly. The exclusion analogue would provide upper bounds on how well Bob can *exclude* states using separable measurements — directly bounding what is achievable under LOCC for antidistinguishability.

## Background

For state exclusion under separable measurements, the SDP hierarchy has the same symmetric extension structure but with the objective flipped: instead of maximizing the sum over $k$ of $p_k \operatorname{Tr}(\rho_k \mu_k)$, we minimize it. The constraints on the extended variables $X_k$ (positivity, symmetric extension, PPT across all cuts) remain the same.

Concretely, the hierarchy at level $s$ for exclusion is:

> **minimize** the sum over $k$ from $1$ to $N$ of $p_k \langle \rho_k, \mu(k) \rangle$, **subject to** the same constraints as Eq. 4.55 of Cosentino's thesis.

At level 1 this gives the PPT exclusion bound. As $s \to \infty$, it converges to the true separable exclusion value.

This is directly relevant to the question: **when are quantum states antidistinguishable under LOCC/separable measurements?** For bipartite states with entanglement across Alice and Bob's labs, the global, PPT, and SEP exclusion values can all differ. Having this hierarchy in `|toqito⟩` would enable systematic numerical exploration of these gaps.

## What should be added

A function `symmetric_extension_hierarchy_exclusion` (or add an `objective` parameter to the existing function) that computes the exclusion analogue.

### Relevant code

- [`toqito/state_opt/symmetric_extension_hierarchy.py`](https://github.com/vprusso/toqito/blob/master/toqito/state_opt/symmetric_extension_hierarchy.py) — the distinguishability version to use as template
- [`toqito/state_opt/state_exclusion.py`](https://github.com/vprusso/toqito/blob/master/toqito/state_opt/state_exclusion.py) — the global exclusion SDP
- [Cosentino's thesis](https://uwspace.uwaterloo.ca/handle/10012/9572) — Chapter 4 for the hierarchy applied to discrimination, Chapter 5 for exclusion
- [arXiv:2206.08313](https://arxiv.org/abs/2206.08313) — Russo and Sikora, *A note on the inner products of pure states and their antidistinguishability* (background motivation)

### Why this is not a simple copy-paste

While the constraints are structurally the same, there are subtleties:
- The exclusion objective minimizes rather than maximizes — this affects SDP solver behavior and numerical stability
- The dual problem has a different structure (the dual of a minimization over PSD variables with symmetric extension constraints)
- The relationship between levels must be verified: level $s+1$ should give a tighter (higher) lower bound on exclusion error than level $s$
- For the distinguishability case, the hierarchy gives *upper* bounds that decrease with level. For exclusion, the hierarchy gives *lower* bounds on the error that increase with level. The monotonicity direction must be verified in tests.

### Acceptance criteria

- Function computes exclusion bounds at any level $s \geq 1$
- Level 1 output matches the PPT exclusion value (once issue #1520 is implemented, or verified independently)
- Tests include:
  - Three Bell states tensored with a resource state $|\tau_\epsilon\rangle$ (the same example used in the existing `symmetric_extension_hierarchy` docstring) — verify that the exclusion hierarchy values are consistent with the distinguishability values (exclusion value + distinguishability value relationships)
  - Monotonicity check: level 2 gives a value $\geq$ level 1 for the exclusion error
  - A case where SEP exclusion is strictly worse than global exclusion (demonstrates the hierarchy is doing something nontrivial)
- Docstring includes the SDP formulation and at least one worked example with known values

### Design options

Two reasonable approaches:
1. **New function** `symmetric_extension_hierarchy_exclusion` alongside the existing one
2. **Add a parameter** like `objective="distinguish"` / `objective="exclude"` to the existing `symmetric_extension_hierarchy`

Either is acceptable. Option 2 avoids code duplication but changes the existing API (the parameter should default to `"distinguish"` for backward compatibility).

### Out of scope

- Implementing the full Cosentino hierarchy for channels
- Proving convergence of the hierarchy (this is a known theoretical result)
- Combining with the eigenvalue criterion for circulant states

### Related issues

- #1520 — PPT state exclusion. The level-1 value of this hierarchy should match the PPT exclusion value implemented there. Implementations are independent: this issue can be done first and use a hand-coded PPT comparison, or it can wait on #1520. Either order is fine.

### Before starting

If you're picking this up, please consider commenting with a short design sketch (new function vs. `objective=` parameter, dual structure) before going deep. Not required — direct PRs welcome.

### Notation

A good PR should include a short note in the docstring or PR description explaining how the paper's / thesis's notation (e.g., $X_k$, $\mu(k)$, the symmetric projector $\Pi$) maps to the variables used in `toqito`.

pandey-tushar:
Closes #1523.

`symmetric_extension_hierarchy` previously computed only upper bounds on the separable (SEP) value for state *distinguishability*. This adds an `objective` parameter so the same hierarchy can also bound state *exclusion*.

With `objective="exclude"` the SDP keeps the identical feasible region (the symmetric-extension and PPT constraints are unchanged) and only flips the objective from `Maximize` to `Minimize`, which matches the global exclusion SDP in `state_exclusion.py`. At level 1 this equals the PPT exclusion value, and the bound is non-decreasing in the level (it converges to the separable value from below), mirroring the distinguishability hierarchy that converges from above.

`objective` defaults to `"distinguish"`, so existing behaviour is unchanged.

The docstring includes the exclusion SDP formulation, the monotonicity direction, the mapping from the thesis notation to the code variables (`mu(k)` -> `meas[k]`, `X_k` -> `x_var[k]`, the symmetric projector -> `sym`), and a runnable worked example.

### Tests

- level 1 of the exclusion hierarchy equals a directly computed PPT exclusion SDP;
- monotonicity in the level;
- a case where separable exclusion is strictly worse than global exclusion, with the gap certified independently by a direct PPT-vs-global comparison (so it is not a solver artifact);
- orthogonal states are perfectly excludable (value 0);
- an invalid `objective` raises a clear error.

All new and existing tests pass, and `ruff check`/`ruff format` are clean.

---

AI disclosure: I used Claude Code to help implement and test this change. I reviewed, ran, and verified all of the code and tests myself.

### DIFF
--- a/toqito/state_opt/symmetric_extension_hierarchy.py
+++ b/toqito/state_opt/symmetric_extension_hierarchy.py
@@ -14,6 +14,7 @@ def symmetric_extension_hierarchy(
     probs: list[float] | None = None,
     level: int = 2,
     dim: int | list[int] | None = None,
+    objective: str = "distinguish",
 ) -> float:
     r"""Compute optimal value of the symmetric extension hierarchy SDP [@navascues2008pure].
 
@@ -58,14 +59,35 @@ def symmetric_extension_hierarchy(
         \end{equation}
     \]
 
+    For state **exclusion**, set `objective="exclude"`. The feasible region is identical; only the
+    objective is *minimized* rather than maximized,
+
+    \[
+        \begin{equation}
+            \text{minimize:} \quad \sum_{k=1}^N p_k \langle \rho_k, \mu(k) \rangle
+            \quad \text{subject to the same constraints.}
+        \end{equation}
+    \]
+
+    At level 1 this recovers the PPT exclusion value, and as the level increases it gives a
+    non-decreasing lower bound on the separable exclusion error -- in contrast to the
+    distinguishability case, where the hierarchy yields an upper bound that decreases with the
+    level. In the implementation, `meas[k]` is the measurement operator \(\mu(k)\), `x_var[k]` is
+    its symmetric extension \(X_k\), and `sym` is the symmetric projector \(\Pi\).
+
     Args:
         states: A list of states provided as either matrices or vectors.
         probs: Respective list of probabilities each state is selected.
         level: Level of the hierarchy to compute.
         dim: The default has both subsystems of equal dimension.
+        objective: Either `"distinguish"` (default) to maximize the success probability of state
+            distinguishability, or `"exclude"` to minimize the error probability of state
+            exclusion. The constraints are identical for both objectives.
 
     Returns:
-        The optimal probability of the symmetric extension hierarchy SDP for level `level`.
+        The optimal value of the symmetric extension hierarchy SDP at level `level`: the
+        distinguishability value for `objective="distinguish"`, or the exclusion error for
+        `objective="exclude"`.
 
     Examples:
         It is known from [@cosentino2015quantum] that distinguishing three Bell states along with a resource
@@ -130,12 +152,35 @@ def symmetric_extension_hierarchy(
         print(f"True separable value: {np.around(true_sep_val, decimals=2)}")
         ```
 
+        The exclusion analogue is selected with `objective="exclude"`. Here three
+        non-antidistinguishable two-qubit states have a strictly positive separable exclusion
+        error already at the first level of the hierarchy.
+
+        ```python exec="1" source="above" result="text"
+        from toqito.state_opt import symmetric_extension_hierarchy
+        import numpy as np
+
+        # Three states (computational basis ordered as |00>, |01>, |10>, |11>).
+        vecs = [
+        np.array([[0], [1], [1], [1]], dtype=complex),
+        np.array([[0], [0], [1], [1]], dtype=complex),
+        np.array([[1], [0], [1], [1]], dtype=complex),
+        ]
+        states = [v @ v.conj().T / (v.conj().T @ v).real.item() for v in vecs]
+
+        # The first level of the hierarchy equals the PPT exclusion error.
+        val = symmetric_extension_hierarchy(states=states, level=1, objective="exclude")
+        print(f"Separable exclusion error (level 1): {np.around(val, decimals=2)}")
+        ```
+
     """
     obj_func = []
     meas = []
     x_var = []
     constraints = []
 
+    if objective not in ("distinguish", "exclude"):
+        raise ValueError("Argument `objective` must be either 'distinguish' or 'exclude'.")
     if not states:
         raise ValueError("At least one state must be provided.")
     if probs is None:
@@ -186,10 +231,14 @@ def symmetric_extension_hierarchy(
 
     constraints.append(sum(meas) == np.identity(dim_xy))
 
-    objective = cvxpy.Maximize(cvxpy.real(sum(obj_func)))
+    # For distinguishability we maximize the success probability; for exclusion we
+    # minimize the error probability. The feasible region (symmetric extension + PPT
+    # constraints) is identical in both cases.
+    obj_expr = cvxpy.real(sum(obj_func))
+    problem_objective = cvxpy.Maximize(obj_expr) if objective == "distinguish" else cvxpy.Minimize(obj_expr)
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", message="Constraint.*subexpressions")
-        problem = cvxpy.Problem(objective, constraints)
+        problem = cvxpy.Problem(problem_objective, constraints)
         sol_default = problem.solve()
 
     return sol_default
--- a/toqito/state_opt/symmetric_extension_hierarchy.py
+++ b/toqito/state_opt/symmetric_extension_hierarchy.py
@@ -162,11 +162,11 @@ def symmetric_extension_hierarchy(
 
         # Three states (computational basis ordered as |00>, |01>, |10>, |11>).
         vecs = [
-        np.array([[0], [1], [1], [1]], dtype=complex),
-        np.array([[0], [0], [1], [1]], dtype=complex),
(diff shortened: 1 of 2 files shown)


vprusso (commented):
Nice — clean implementation. The math is right (same feasible region, flipped objective; level 1 = PPT exclusion since the symmetric-extension constraint trivializes at `s=1`), and the monotonicity-direction claim in the docstring matches the theory.

A few items.

### Missing per issue acceptance criteria

The issue asked specifically for the **three Bell states tensored with a resource state `|τ_ε⟩`** test (mirroring the existing distinguishability docstring example), verifying that the exclusion hierarchy values are consistent with the distinguishability values. The PR's tests use a custom 4-dim ensemble instead. The Bell-tensor-resource case is the canonical comparison and worth including — the gap between distinguishability and exclusion has a clean physical interpretation there.

### Monotonicity test is weak

```python
np.testing.assert_equal(lvl_2 >= lvl_1 - 1e-5, True)
```

The comment notes that PPT is tight for this ensemble, so level 2 equals level 1 to solver precision. That means this test only verifies *non-decrease tolerant to solver noise* — it would pass equally well if the levels happened to coincide for the wrong reason. If there's an ensemble (even small/synthetic) where SEP > PPT strictly, that's the one that would actually exercise the hierarchy's monotonicity. Finding one is non-trivial but worth a try — otherwise rename the test to something like `test_exclusion_level_two_does_not_decrease` so future readers know what it's really asserting.

### `_direct_exclusion_sdp` will become redundant

Once #1566 (which adds `measurement="ppt"` to `state_exclusion`) lands, this test-local SDP can be replaced by a call to `state_exclusion(..., measurement="ppt", subsystems=[0], dimensions=[2, 2])`. Not a blocker — happy to leave the local reference SDP here and clean up in a follow-up after #1566 merges. Just flagging the future overlap.

### Minor

- The docstring example has off-indented list items inside the `vecs = [...]` literal (legal Python, just visually odd). One-tab indent for the array elements would match the surrounding examples.
- `v @ v.conj().T / (v.conj().T @ v).real.item()` works but could be `v @ v.conj().T / float(np.linalg.norm(v)**2)` — same result, fewer hoops.

---

Solid PR. Once the Bell-tensor-resource test goes in and the monotonicity test gets either a strict-inequality ensemble or a rename, I'm happy to merge.

pandey-tushar:
Thanks for the review! Addressed:

- Added `test_exclusion_three_bell_with_resource_state` on the canonical three-Bell-plus-resource
  ensemble. The states are mutually orthogonal, so the exclusion error is 0 while the PPT
  distinguishability value is ~0.99, which makes the exclusion and distinguishability hierarchy
  values directly comparable on the known example.
- Renamed the monotonicity test to `test_exclusion_level_two_does_not_decrease` and documented that
  the two levels coincide here (PPT is tight), so it only asserts the non-decreasing direction. I
  looked for a small-dimension ensemble with SEP strictly above PPT for exclusion but could not find
  one, consistent with your "non-trivial" note; happy to add it if you know of a good example.
- Indented the docstring list and switched the normalization to `np.linalg.norm`.
- On the `_direct_exclusion_sdp` overlap: `measurement="ppt"` support for `state_exclusion` isn't in
  toqito yet (PR #1566 was closed; the feature is still tracked by the open issue #1520), so I kept
  the inline reference SDP self-contained for now. Happy to switch it to
  `state_exclusion(..., measurement="ppt", ...)` as a follow-up once that support lands.

vprusso (approved):
Nice updates. All four points addressed:

- **Bell-tensor-resource test** is the right shape: mutual orthogonality gives `exclude ≈ 0` (perfectly antidistinguishable), `distinguish > 0.9` (close to but strictly less than 1, which is the classical NLWE asymmetry). One follow-up idea — tightening the distinguishability assertion to also cap it from above (`distinguish < 1 - 1e-3` or similar) would make the asymmetry sharper in the test, but it's already meaningful as-is.
- **Monotonicity test rename** + docstring is honest. Your note about not finding a SEP-strictly-above-PPT small ensemble matches what I've seen too; we'll add it if someone publishes one.
- **Docstring indentation** + **`np.linalg.norm` normalization** look clean.
- **`_direct_exclusion_sdp`**: good call — you're right, #1566 was closed, so the PPT path for `state_exclusion` isn't available yet (#1520 is still open). Keeping the inline reference SDP is the right call for now; we can revisit when PPT support lands.

Approving — happy to merge once the pytest matrix finishes.

pandey-tushar:
@vprusso all green

vprusso:
Merged as 8e453aa1, #1523 auto-closed. Nice work — symmetric extension hierarchy now ships exclusion alongside distinguishability.

PR #1568, opened 2026-06-04
Add one-way LOCC state exclusion via see-saw optimization

### LINKED ISSUE #1525: Implement SDP bounds for quantum state exclusion under LOCC constraints via see-saw optimization
State exclusion under local operations and classical communication (LOCC) is central to the study of quantum nonlocality without entanglement (see [arXiv:2412.11eft](https://arxiv.org/abs/2412.12889) — Russo, Sikora, *Quantum Nonlocality without Entanglement and State Discrimination Measures*). Currently, \`|toqito⟩\` has no way to compute exclusion probabilities under LOCC constraints.

## Background

For bipartite states $\{\rho_1, \ldots, \rho_n\}$ shared between Alice and Bob, LOCC exclusion asks: what is the minimum error probability when Alice and Bob can only perform local operations and communicate classically, and must identify which state was *not* sent?

Exact optimization over LOCC is intractable, but two tractable approaches exist:

1. **See-saw (alternating optimization) for LOCC lower bounds:** Fix Alice's measurement, optimize Bob's, then alternate. This is already the approach used in \`NonlocalGame.quantum_value_lower_bound\` and has a direct analogue for state exclusion under one-way LOCC.

2. **One-way LOCC exclusion SDP:** Under one-way LOCC (Alice measures first, communicates outcome to Bob, Bob measures), the problem decomposes. For each of Alice's measurement outcomes $a$, Bob receives a conditional ensemble and performs state exclusion on it. The overall optimization over Alice's POVM and Bob's conditional POVMs can be cast as an SDP.

The one-way LOCC SDP for state *distinguishability* is well-studied (see Section 4.2 of Cosentino's thesis [arXiv:1503.02181](https://arxiv.org/abs/1503.02181)). The exclusion analogue follows the same LOCC decomposition structure but with the exclusion objective.

## What should be added

A function for computing state exclusion under one-way LOCC constraints, accepting bipartite states with specified subsystem dimensions.

### Relevant code

- [\`toqito/state_opt/state_exclusion.py\`](https://github.com/vprusso/toqito/blob/master/toqito/state_opt/state_exclusion.py) — global exclusion SDP
- [\`toqito/state_opt/state_distinguishability.py\`](https://github.com/vprusso/toqito/blob/master/toqito/state_opt/state_distinguishability.py) — has PPT constraints, could serve as template for subsystem-aware measurements
- [\`toqito/nonlocal_games/nonlocal_game.py\`](https://github.com/vprusso/toqito/blob/master/toqito/nonlocal_games/nonlocal_game.py) — see-saw optimization pattern (\`__optimize_alice\` / \`__optimize_bob\`)
- [arXiv:1503.02181](https://arxiv.org/abs/1503.02181) — Cosentino's thesis, one-way LOCC discrimination SDP
- [arXiv:2412.12889](https://arxiv.org/abs/2412.12889) — Russo and Sikora, nonlocality and discrimination measures

### Why this is not straightforward

The see-saw for exclusion under LOCC has several subtleties:

- **Exclusion vs. distinguishability objective:** In LOCC distinguishability, Alice's goal (in the see-saw inner loop) is to steer Bob toward states that are easy to distinguish. In LOCC exclusion, Alice must steer Bob toward states where at least one can be reliably excluded. The post-measurement ensembles have a different optimization landscape.

- **One-way vs. two-way:** One-way LOCC (Alice→Bob) is an SDP. Two-way LOCC requires the see-saw. The implementation should clearly document which is being computed.

- **Convergence:** The see-saw converges to a local optimum. Multiple random initializations are needed, following the pattern in \`NonlocalGame.quantum_value_lower_bound\`.

- **Bipartite structure:** The function must correctly handle the tensor product structure $\mathcal{H}_A \otimes \mathcal{H}_B$ and partial traces over the appropriate subsystems.

### Acceptance criteria

- Implements at least one-way LOCC exclusion (Alice measures first)
- Accepts bipartite states with subsystem dimensions
- Tests include:
  - Domino states (known to be locally indistinguishable but globally distinguishable) — verify that LOCC exclusion gives a different value than global exclusion
  - Product states where LOCC exclusion should match global exclusion
  - A case from the quantum nonlocality without entanglement literature where the LOCC gap is known
- At least 3 random restarts for the see-saw, returning the best value found
- Docstring with SDP formulation and worked example

### Out of scope

- Full two-way LOCC (computationally intractable to certify optimality)
- Combining with PPT or symmetric extension bounds
- Resource-assisted LOCC exclusion

pandey-tushar:
Closes #1525.

Adds `state_exclusion_locc`, which computes the minimum-error probability of quantum state exclusion under one-way LOCC: Alice measures her subsystem, communicates her outcome to Bob, and Bob announces a state to exclude.

The induced global measurement operator for guessing `k` is `M_k = sum_a A_a (x) B^a_k`, and the error `sum_k p_k sum_a Tr[(A_a (x) B^a_k) rho_k]` is bilinear in Alice's and Bob's operators, so it is minimized by a see-saw of two SDPs:

- fix Alice's POVM and optimize Bob's conditional POVMs on the steered ensembles `sigma^a_k = Tr_A[(A_a (x) I) rho_k]`;
- fix Bob's POVMs and optimize Alice's POVM.

This mirrors the see-saw in `NonlocalGame.quantum_value_lower_bound`, including the random restarts. The value returned is the error of a concrete one-way LOCC strategy, hence at least the global exclusion error.

The docstring gives the two-SDP formulation, the notation mapping, and a runnable worked example.

### Tests

- orthogonal product states are perfectly LOCC-excludable (matching the global value);
- a fixed entangled ensemble where LOCC exclusion is strictly worse than global, with the gap certified by a direct PPT lower bound (PPT exclusion is a lower bound on LOCC exclusion, so PPT > global proves the gap is real rather than a see-saw artifact);
- reproducibility under a fixed seed;
- input validation.

### A note on the suggested domino-states test

The issue suggested that domino states would show a global-vs-LOCC exclusion gap. They do not: domino states are an orthonormal product basis, so although they are not LOCC-*distinguishable*, they are perfectly LOCC-*excludable* (exclusion only needs to rule out one state). I included a test confirming the domino exclusion error is 0, and demonstrated the strict global-vs-LOCC gap with the entangled ensemble above instead. Happy to adjust if you prefer a different example.

`ruff check`/`ruff format` are clean and all tests pass.

---

AI disclosure: I used Claude Code to help implement and test this change. I reviewed, ran, and verified all of the code and tests myself.

### DIFF
--- a/toqito/state_opt/__init__.py
+++ b/toqito/state_opt/__init__.py
@@ -7,6 +7,7 @@ from toqito.state_opt.optimal_clone import optimal_clone
 from toqito.matrix_props import has_same_dimension as _has_same_dimension  # noqa: F401
 from toqito.state_opt.state_distinguishability import state_distinguishability
 from toqito.state_opt.state_exclusion import state_exclusion
+from toqito.state_opt.state_exclusion_locc import state_exclusion_locc
 from toqito.state_opt.npa_hierarchy import npa_constraints, bell_npa_constraints
 from toqito.state_opt.symmetric_extension_hierarchy import symmetric_extension_hierarchy
 from toqito.state_opt.bell_inequality_max import bell_inequality_max, bell_inequality_max_qubits
--- a/toqito/state_opt/state_exclusion_locc.py
+++ b/toqito/state_opt/state_exclusion_locc.py
@@ -10,7 +10,7 @@ from toqito.rand import random_povm
 def state_exclusion_locc(
     states: list[np.ndarray],
     probs: list[float] | None = None,
-    dims: list[int] | None = None,
+    dim: list[int] | None = None,
     *,
     num_alice_outcomes: int | None = None,
     reps: int = 5,
@@ -67,7 +67,7 @@ def state_exclusion_locc(
         states: Bipartite states, each given as a density matrix (or a vector, which is converted
             to a density matrix) on the \(\mathcal{H}_A \otimes \mathcal{H}_B\) system.
         probs: Respective selection probabilities. Defaults to the uniform distribution.
-        dims: The two subsystem dimensions `[dim_A, dim_B]`. Their product must equal the dimension
+        dim: The two subsystem dimensions `[dim_A, dim_B]`. Their product must equal the dimension
             of each state.
         num_alice_outcomes: Number of measurement outcomes available to Alice. Defaults to the
             number of states. Allowing more outcomes can only tighten (lower) the value.
@@ -89,21 +89,21 @@ def state_exclusion_locc(
         import numpy as np
 
         vecs = [
-        np.array([[0], [1], [1], [1]], dtype=complex),
-        np.array([[0], [0], [1], [1]], dtype=complex),
-        np.array([[1], [0], [1], [1]], dtype=complex),
+            np.array([[0], [1], [1], [1]], dtype=complex),
+            np.array([[0], [0], [1], [1]], dtype=complex),
+            np.array([[1], [0], [1], [1]], dtype=complex),
         ]
-        states = [v @ v.conj().T / (v.conj().T @ v).real.item() for v in vecs]
+        states = [v @ v.conj().T / float(np.linalg.norm(v) ** 2) for v in vecs]
 
-        val = state_exclusion_locc(states, dims=[2, 2], reps=3, seed=1)
+        val = state_exclusion_locc(states, dim=[2, 2], reps=3, seed=1)
         print(f"One-way LOCC exclusion error: {np.around(val, decimals=2)}")
         ```
 
     """
     if not states:
         raise ValueError("At least one state must be provided.")
-    if dims is None or len(dims) != 2:
-        raise ValueError("Argument `dims` must be a list `[dim_A, dim_B]` of the two subsystem dimensions.")
+    if dim is None or len(dim) != 2:
+        raise ValueError("Argument `dim` must be a list `[dim_A, dim_B]` of the two subsystem dimensions.")
 
     states = [to_density_matrix(state) for state in states]
     num_states = len(states)
@@ -112,9 +112,9 @@ def state_exclusion_locc(
     if not np.isclose(sum(probs), 1):
         raise ValueError("Probabilities must sum to 1.")
 
-    dim_a, dim_b = int(dims[0]), int(dims[1])
+    dim_a, dim_b = int(dim[0]), int(dim[1])
     if states[0].shape[0] != dim_a * dim_b:
-        raise ValueError("The product of `dims` must equal the dimension of the states.")
+        raise ValueError("The product of `dim` must equal the dimension of the states.")
     num_a = num_states if num_alice_outcomes is None else num_alice_outcomes
 
     best = float("inf")
--- a/toqito/state_opt/state_exclusion_locc.py
+++ b/toqito/state_opt/state_exclusion_locc.py
@@ -1,0 +1,189 @@
+"""Computes state exclusion under one-way LOCC via see-saw optimization."""
+
+import cvxpy
+import numpy as np
+
+from toqito.matrix_ops import partial_trace, to_density_matrix
+from toqito.rand import random_povm
+
+
+def state_exclusion_locc(
+    states: list[np.ndarray],
+    probs: list[float] | None = None,
+    dims: list[int] | None = None,
+    *,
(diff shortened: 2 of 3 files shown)


vprusso (commented):
Solid PR. The math is correct: induced operator `M_k = Σ_a A_a ⊗ B^a_k`, error `Σ_k p_k Tr[M_k ρ_k]`, each see-saw half is a clean SDP with the other party's variables fixed to numpy arrays so the per-step problem stays linear. The orderings `global ≤ PPT ≤ SEP ≤ LOCC` (for the *error*, since we're minimizing over a shrinking feasible set) are right, and using `PPT > global` as a certificate that the see-saw value isn't an artifact is exactly the right move — `PPT` lower-bounds the true LOCC value, so once it clears global, the gap is genuine no matter what the see-saw converged to.

### Domino-states catch

You're right and I was wrong in the issue. I asserted that domino states would show a LOCC exclusion gap; they don't, for exactly the reason you give — they're an orthonormal product basis, and after Alice's measurement Bob is left with a 3-state product ensemble on his side which he can identify exactly. I'll fix the issue body. The substitute test (perfect LOCC excludability of domino, error = 0) is the right test to keep.

### Minor

- **`dims` vs `dim`.** The rest of `toqito` uses `dim` as the keyword (cf. `partial_trace`, `partial_transpose`, `state_distinguishability`). Worth renaming for consistency, especially since you also accept `dim_a`, `dim_b` internally.
- **Convergence check.** `abs(prev - current) < tol` is fine but mathematically the see-saw is monotone non-increasing (each SDP minimizes, the other party held fixed), so `prev - current < tol` is the natural form. Not blocking.
- **`bob` as a dict keyed by `(a, k)` tuples** works but is unusual in toqito. A `list[list[cvxpy.Variable]]` indexed `bob[a][k]` would read more naturally. Style preference, not blocking.
- **Docstring list literal indentation** — same off-indented `vecs = [...]` body as #1567, easy fix.

### Acceptance criterion: "a case from the NLWE literature where the LOCC gap is known"

The issue asked for an NLWE example with a known LOCC gap. The custom `_GAP_VECS` ensemble shows a clean gap and is a reasonable substitute, but it's not in the literature. A canonical NLWE example that *does* show a LOCC-exclusion gap would be stronger — but as your domino analysis shows, finding such an example is harder than I implied in the issue (most NLWE examples are about distinguishability, not exclusion). I'd accept the `_GAP_VECS` ensemble as-is and treat the "canonical NLWE example" criterion as a follow-up if one is ever published.

---

Once `dims → dim` and the docstring indentation are tidied, I'm happy to merge. Nice work catching the domino mis-claim.

pandey-tushar:
Thanks! Addressed:

- Renamed the public `dims` argument to `dim`, matching `partial_trace`, `partial_transpose`, and
  `state_distinguishability`.
- Indented the docstring example list.

Glad the domino analysis was useful, and thanks for offering to update the issue. I left the `bob`
dict keyed by `(a, k)` and the `abs(prev - current)` convergence check as-is since they are not
blocking, but happy to switch `bob` to a `list[list[...]]` and tighten the convergence form if you
prefer.

vprusso (approved):
Looks good — the `dims → dim` rename is comprehensive (signature, docstring, error messages, all five test sites including the regex matches and the renamed `test_requires_dim`), and the cleaner `float(np.linalg.norm(v) ** 2)` normalization reads better.

You're right to leave the `bob` dict and `abs(prev - current)` as-is — both were tagged "not blocking" and they're idiomatic enough. No change needed.

Approving — happy to merge once the pytest matrix finishes (code-style is already green, just waiting on the runners).

pandey-tushar:
@vprusso I have added tests to improve codecov.

vprusso:
Coverage additions look clean — the dim-product mismatch, the `seed=None` path, and the explicit `num_alice_outcomes` parameter were all genuinely untested branches. Approval still stands; I'll merge once the new CI run finishes.

pandey-tushar:
@vprusso Thanks. I think you would need to trigger the workflow/pipeline again.

vprusso:
> @vprusso Thanks. I think you would need to trigger the workflow/pipeline again. 

Running now!

pandey-tushar:
> > @vprusso Thanks. I think you would need to trigger the workflow/pipeline again.
> 
> Running now!

Thank you. Seems like it's all green now!

PR #1576, opened 2026-06-14
feat: add PPT state exclusion support

### LINKED ISSUE #1520: Add PPT measurement support to state exclusion SDP

kris70lesgo:
Closes #1520.

Adds `measurement="ppt"` support to `state_exclusion`, allowing minimum-error and unambiguous quantum state exclusion SDPs to be solved with PPT-constrained measurement operators.

For minimum-error exclusion, the primal now supports:

```math
\begin{aligned}
\min \quad & \sum_i p_i \langle M_i, \rho_i \rangle \\
\text{s.t.} \quad & \sum_i M_i = I, \\
& M_i \succeq 0, \\
& \Gamma_S(M_i) \succeq 0 \quad \forall i.
\end{aligned}
```

Here, `\Gamma_S` denotes partial transpose over the subsystem indices passed through `subsystems`, with local dimensions given by `dimensions`.

The dual is implemented using the dual of the PPT cone. Since:

```math
\mathrm{PPT} = \mathrm{Pos} \cap \Gamma_S(\mathrm{Pos}),
```

the dual cone is:

```math
\mathrm{PPT}^* = \mathrm{Pos} + \Gamma_S(\mathrm{Pos}).
```

Thus the minimum-error dual constraints are modeled as:

```math
p_i \rho_i - Y \in \mathrm{Pos} + \Gamma_S(\mathrm{Pos}).
```

This matters for exclusion because the SDP minimizes overlap, unlike distinguishability, so the PPT dual cannot be obtained by simply copying the distinguishability implementation with the objective flipped.

The same PPT measurement constraint is also added for the unambiguous exclusion primal, including the inconclusive measurement operator. The unambiguous dual is implemented directly using the same dual PPT cone structure.

Tests added:

- PPT exclusion for Bell states, covering both `min_error` and `unambiguous`, primal and dual.
- Agreement between PPT exclusion and level-1 symmetric-extension hierarchy on a regression ensemble.
- A case where PPT exclusion is strictly worse than global exclusion.
- Bell-state-with-resource regression against the symmetric-extension hierarchy.
- Input validation for missing `subsystems` / `dimensions`.

Docs updated with the PPT SDP formulation, notation mapping, and a worked `measurement="ppt"` example.

Verification:

- `ruff check toqito/state_opt/state_exclusion.py toqito/state_opt/tests/test_state_exclusion.py`
- `ruff format --check toqito/state_opt/state_exclusion.py toqito/state_opt/tests/test_state_exclusion.py`
- `pytest toqito/state_opt/tests/test_state_exclusion.py`
- `pytest toqito/state_opt/tests`
- full `pytest`
- `mkdocs build -f mkdocs.yml` from `docs/`

All passed locally.

### DIFF
--- a/toqito/state_opt/state_exclusion.py
+++ b/toqito/state_opt/state_exclusion.py
@@ -171,7 +171,7 @@ def state_exclusion(
 
         states = [np.array([[1.], [0.]]), np.array([[1.],[1.]]) / np.sqrt(2)]
 
-        res, _ = state_exclusion(states, primal_dual="primal", strategy="unambiguous", abs_ipm_opt_tol=1e-5)
+        res, _ = state_exclusion(states, primal_dual="primal", strategy="unambiguous", abs_ipm_opt_tol=1e-7)
 
         print(np.around(res, decimals=2))
         ```
@@ -221,6 +221,9 @@ def state_exclusion(
             See https://gitlab.com/picos-api/picos/-/issues/341
 
     """
+    if measurement not in {"positive", "ppt"}:
+        raise ValueError("Argument `measurement` must be either 'positive' or 'ppt'.")
+
     if not has_same_dimension(vectors):
         raise ValueError("Vectors for state distinguishability must all have the same dimension.")
 
@@ -271,9 +274,6 @@ def state_exclusion(
             **kwargs,
         )
 
-    if measurement != "positive":
-        raise ValueError("Argument `measurement` must be either 'positive' or 'ppt'.")
-
     if strategy == "min_error":
         if primal_dual == "primal":
             return _min_error_primal(vectors=vectors, dim=dim, probs=probs, solver=solver, **kwargs)
@@ -303,7 +303,7 @@ def _ppt_dual_cone_constraint(
     name: str,
     subsystems: list[int],
     dimensions: list[int],
-) -> None:
+) -> Any:
     r"""Constrain `expr` to lie in the dual PPT cone.
 
     The PPT cone is \(\{M : M \succeq 0, \Gamma(M) \succeq 0\}\). Its dual cone is
@@ -314,7 +314,9 @@ def _ppt_dual_cone_constraint(
     dim = expr.shape[0]
     q_var = picos.HermitianVariable(name, (dim, dim))
     problem.add_constraint(q_var >> 0)
-    problem.add_constraint(expr - picos.partial_transpose(q_var, subsystems=subsystems, dimensions=dimensions) >> 0)
+    return problem.add_constraint(
+        expr - picos.partial_transpose(q_var, subsystems=subsystems, dimensions=dimensions) >> 0
+    )
 
 
 def _min_error_primal(
@@ -402,19 +404,22 @@ def _ppt_min_error_dual(
     """Find the dual problem for minimum-error quantum state exclusion SDP with PPT constraints."""
     problem = picos.Problem()
     y_var = picos.HermitianVariable("Y", (dim, dim))
+    dual_cone_constraints = []
 
     for i, vector in enumerate(vectors):
-        _ppt_dual_cone_constraint(
-            problem=problem,
-            expr=probs[i] * to_density_matrix(vector) - y_var,
-            name=f"Q[{i}]",
-            subsystems=subsystems,
-            dimensions=dimensions,
+        dual_cone_constraints.append(
+            _ppt_dual_cone_constraint(
+                problem=problem,
+                expr=probs[i] * to_density_matrix(vector) - y_var,
+                name=f"Q[{i}]",
+                subsystems=subsystems,
+                dimensions=dimensions,
+            )
         )
 
     problem.set_objective("max", picos.trace(y_var))
     solution = problem.solve(solver=solver, primals=None, **kwargs)
-    measurements = [problem.get_constraint(2 * i + 1).dual for i in range(len(vectors))]
+    measurements = [constraint.dual for constraint in dual_cone_constraints]
 
     return solution.value, measurements
 
@@ -463,6 +468,7 @@ def _ppt_unambiguous_primal(
     n = len(vectors)
     problem = picos.Problem()
     measurements = [picos.HermitianVariable(f"M[{i}]", (dim, dim)) for i in range(n)]
+    # The inconclusive POVM element is determined by completeness once the conclusive elements are chosen.
     inconclusive_measurement = picos.I(dim) - picos.sum(measurements)
 
     problem.add_list_of_constraints([meas >> 0 for meas in measurements])
--- a/toqito/state_opt/state_exclusion.py
+++ b/toqito/state_opt/state_exclusion.py
@@ -13,6 +13,9 @@ def state_exclusion(
     vectors: list[np.ndarray],
     probs: list[float] | None = None,
     strategy: str = "min_error",
+    measurement: str = "positive",
+    subsystems: list[int] | None = None,
+    dimensions: list[int] | None = None,
     solver: str = "cvxopt",
     primal_dual: str = "dual",
     **kwargs: Any,
@@ -64,6 +67,21 @@ def state_exclusion(
         \end{equation}
     \]
 
+    If `measurement = "ppt"`, the measurement operators are also constrained to be positive under partial transpose on
+    the requested subsystem(s). In the primal minimum-error problem, this adds the constraints
+
+    \[
+        \Gamma_j(M_i) \succeq 0
+    \]
+
+    for every measurement operator \(M_i\) and for the subsystem(s) \(j\) listed in `subsystems`, where \(\Gamma_j\)
+    is the partial transpose map with subsystem dimensions `dimensions`. The corresponding dual uses the dual PPT cone,
+    \(\text{Pos} + \Gamma_j(\text{Pos})\), so the constraints become
+
+    \[
+        p_i\rho_i - Y \in \text{Pos} + \Gamma_j(\text{Pos}).
+    \]
+
     For `strategy = "unambiguous"`, Bob never provides an incorrect answer, although it is
     possible that his answer is inconclusive. This function then yields the probability of an inconclusive outcome.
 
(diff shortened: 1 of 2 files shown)


vprusso (commented):
Thanks for picking this up — the math is sound and the test selection is the right shape. I verified the cone-duality framing numerically (PPT min-error primal/dual gap ~6×10⁻¹¹ on 4 Bell states, both values ≈ 0 as expected since they're antidistinguishable), and the formulation `expr - Γ(Q) ⪰ 0` with `Q ⪰ 0` is correctly modeling `expr ∈ Pos + Γ(Pos)`. The PR description's `PPT = Pos ∩ Γ(Pos)` → `PPT* = Pos + Γ(Pos)` argument is right (both `Pos` and `Γ(Pos)` are self-dual since `Γ` is its own inverse).

The test selection covers the right cases — Bell states ⇒ 0 (sanity), gap ensemble matches `symmetric_extension_hierarchy(level=1, exclude)` (regression against #1567), Bell+resource matches level 1 (the canonical asymmetry test), gap ensemble shows PPT > global strictly. Together these would catch any future regression in the PPT machinery.

A few items before merging.

### Must fix

1. **Unrelated change to an existing docstring example.** Line ~174:

   ```diff
   -        res, _ = state_exclusion(states, primal_dual="primal", strategy="unambiguous", abs_ipm_opt_tol=1e-7)
   +        res, _ = state_exclusion(states, primal_dual="primal", strategy="unambiguous", abs_ipm_opt_tol=1e-5)
   ```

   This relaxes the solver tolerance on an unrelated example that this PR doesn't touch functionally. Either explain why (did the existing example start failing under your environment?) or revert. If it's actually a numerical fix that benefits master, it should ride in a separate PR with its own justification.

2. **Brittle constraint indexing in `_ppt_min_error_dual`:**

   ```python
   measurements = [problem.get_constraint(2 * i + 1).dual for i in range(len(vectors))]
   ```

   This hardcodes the fact that `_ppt_dual_cone_constraint` adds *exactly two* constraints per call (the `Q ⪰ 0` and the `expr - Γ(Q) ⪰ 0`), in that order. If anyone ever refactors the helper to add a different number of constraints, this silently breaks. Two safer options:
   - Have `_ppt_dual_cone_constraint` return the relevant constraint object (the `expr - Γ(Q) ⪰ 0` one), and let the caller collect them into a list.
   - Use named constraints and look them up by name.

   This is the kind of bug that bites months later when someone "just" tweaks the helper.

### Worth addressing

3. **Missing coverage for validation branches.** `_validate_ppt_params` has three error paths (missing params, dim mismatch, out-of-range subsystems) but only one is tested. The dim-mismatch branch (`np.prod(dimensions) != dim`) and the out-of-range subsystems branch are entirely uncovered. Two cheap tests would fix this:

   ```python
   def test_state_exclusion_ppt_wrong_dimensions():
       with pytest.raises(ValueError, match="product of `dimensions`"):
           state_exclusion(vectors=[bell(0), bell(1)], measurement="ppt",
                           subsystems=[0], dimensions=[2, 3])  # 6 != 4

   def test_state_exclusion_ppt_subsystems_out_of_range():
       with pytest.raises(ValueError, match="index into `dimensions`"):
           state_exclusion(vectors=[bell(0), bell(1)], measurement="ppt",
                           subsystems=[5], dimensions=[2, 2])
   ```

4. **Missing coverage for the invalid-`measurement` branch.** Currently:

   ```python
   if measurement == "ppt":
       ...
       return ...
   if measurement != "positive":
       raise ValueError("Argument `measurement` must be either 'positive' or 'ppt'.")
   ```

   The `ValueError` line is uncovered. A one-liner with `measurement="random_string"` would close it. While you're there: I'd move the `measurement` validation to the top of the function alongside the existing parameter validation, so all "what did you pass me?" errors fire before the work starts.

### Nice-to-have

5. **`_ppt_unambiguous_dual` returns `(y_var, lagrangian_variables_a)` — no measurements.** Consistent with the existing positive-cone `_unambiguous_dual` (which also doesn't recover measurements), so I'm not asking you to change this. Just flagging the asymmetry with `_ppt_min_error_dual`, which *does* recover measurements via the brittle indexing in #2.

6. **The `inconclusive_measurement = picos.I(dim) - picos.sum(measurements)` shortcut** in `_ppt_unambiguous_primal` is correct (and arguably cleaner than declaring a new variable). Worth a one-line comment so future readers don't wonder why it differs from `_unambiguous_primal`'s explicit-variable pattern.

---

Once 1 and 2 are addressed (the docstring tolerance change explained-or-reverted, and the dual-constraint indexing made non-brittle), I'm happy to merge. 3–4 would push coverage up but aren't blockers; 5–6 are pure polish.

kris70lesgo:
Thanks for the review, Addressed all requested changes
- Reverted the unrelated docstring tolerance change back to `abs_ipm_opt_tol=1e-7`.
- Removed the brittle constraint indexing in `_ppt_min_error_dual`.
  - `_ppt_dual_cone_constraint` now returns the actual dual-cone constraint object.
  - `_ppt_min_error_dual` collects those constraints directly and reads their duals without relying on `problem.get_constraint(2 * i + 1)`.
- Moved `measurement` validation earlier in `state_exclusion`, alongside the other input checks.
- Added tests for the remaining PPT validation branches:
  - missing `subsystems` / `dimensions`
  - dimension product mismatch
  - out-of-range subsystem index
  - invalid `measurement`
- Added the requested comment explaining why the PPT unambiguous primal uses `I - sum(M_i)` for the inconclusive POVM element.

Verification run locally:

- `ruff check toqito/state_opt/state_exclusion.py toqito/state_opt/tests/test_state_exclusion.py`
- `ruff format --check toqito/state_opt/state_exclusion.py toqito/state_opt/tests/test_state_exclusion.py`
- `pytest toqito/state_opt/tests/test_state_exclusion.py`
- `pytest toqito/state_opt/tests`
All passed

One note on the docstring tolerance: after reverting it, `mkdocs build` still exits successfully locally, but the executable doc block logs the CVXOPT `ArithmeticError: 3` that originally motivated the tolerance change in my environment, I kept the revert in this PR as requested, since it is unrelated to the PPT functionality

vprusso (approved):
All five items addressed cleanly:

- Tolerance reverted to `1e-7` ✓
- `_ppt_dual_cone_constraint` returns the constraint object; `_ppt_min_error_dual` collects them in a list and reads `.dual` off each — no more positional indexing ✓
- `measurement` validation moved to the top, alongside the existing input checks ✓
- Three new validation tests cover all the previously-uncovered branches (product mismatch, out-of-range subsystem, invalid `measurement` string) ✓
- One-line comment on the `I - sum(M_i)` shortcut in `_ppt_unambiguous_primal` ✓

On the `ArithmeticError: 3` you saw on `abs_ipm_opt_tol=1e-7`: that's a known intermittent CVXOPT KKT-solver issue and we've seen it on other PRs. Not blocking — the docstring example block isn't a test, and the build still passes. If it starts failing in CI on some matrix run, we can address it separately.

Tiny nit (non-blocking): the `_ppt_dual_cone_constraint` return type became `Any` — could be tightened to `picos.constraints.Constraint`, but matches the rest of the file's style, so leave as you prefer.

Approving. Will merge once the pytest matrix finishes.

PR #1644, opened 2026-06-28
Add a space-optimized quantum relative entropy SDP.

simsaidan:
Extracts some shared logic from the previous channel relative entropy and uses it to find the qre.

## Changes

  -  [x] Add quantum relative entropy from https://arxiv.org/pdf/2404.17016

## Checklist

### DIFF
--- a/docs/content/refs.bib
+++ b/docs/content/refs.bib
@@ -994,6 +994,18 @@
   issn = {0036-1445},
 }
 
+@article{kossmann2024optimisingrelativeentropy,
+  title = {Optimising the relative entropy under semidefinite constraints},
+  author = {{Ko{\ss}mann}, Gereon and Schwonnek, Ren{\'e}},
+  journal = {npj Quantum Information},
+  year = {2026},
+  doi = {10.1038/s41534-026-01184-4},
+  eprint = {2404.17016},
+  archivePrefix = {arXiv},
+  primaryClass = {quant-ph},
+  url = {https://arxiv.org/abs/2404.17016},
+}
+
 @article{kossmann2025channelrelativeentropy,
   title = {Semidefinite optimization of the quantum relative entropy of channels},
   author = {{Ko{\ss}mann}, Gereon and Wilde, Mark M.},
--- a/toqito/channel_metrics/channel_relative_entropy.py
+++ b/toqito/channel_metrics/channel_relative_entropy.py
@@ -6,158 +6,22 @@ import cvxpy as cvx
 import numpy as np
 
 from toqito.channel_props import is_completely_positive, is_quantum_channel
+from toqito.cones._integral_relative_entropy import (
+    find_lambda as _find_lambda,
+)
+from toqito.cones._integral_relative_entropy import (
+    find_mu as _find_mu,
+)
+from toqito.cones._integral_relative_entropy import (
+    make_delta as _make_delta,
+)
+from toqito.cones._integral_relative_entropy import (
+    make_gamma as _make_gamma,
+)
+from toqito.cones._integral_relative_entropy import (
+    make_grid as _make_grid,
+)
 from toqito.matrix_ops.partial_trace import partial_trace
-
-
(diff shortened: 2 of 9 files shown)


simsaidan:
Hey @vprusso ! Hope you had an enjoyable weekend. This PR is a follow up to the one merged last week. The SDP in channel relative entropy can be applied directly on density matrices to get a SDP for calculating quantum relative entropy with no variable lifting (as seen in https://arxiv.org/pdf/2404.17016).

I factored out the shared logic and then added a new branch to the quantum relative entropy file that lets the user pick between techniques. Looking forward to your review!

vprusso (changes requested):
Thanks for this, the integral-representation branch is a useful addition and the test coverage across dim/apx/hermitian is good. A few things I'd like to see before merging.

Module placement and visibility

The new code belongs in toqito/cones (it's an entropy functional / SDP cone, and it's shared by both channel_relative_entropy and quantum_relative_entropy, so cones is the only neutral home that doesn't create a channel_metrics to state_props dependency). But it should be a normal public module there, not a private one:

- Rename toqito/cones/_integral_relative_entropy.py to toqito/cones/integral_relative_entropy.py, matching the other modules in that package.
- Export evaluate_relative_entropy_integral from toqito/cones/__init__.py alongside the existing cone functions.
- Keep make_grid, find_mu, find_lambda, make_delta, make_gamma as module-private _-prefixed helpers. They were private (_make_grid etc.) before this PR; promoting them to public names while putting them in a private module is backwards. Only the one entry point needs to be public.

That also lets the five separate `from toqito.cones._integral_relative_entropy import x as _x` lines in channel_relative_entropy.py collapse into a single normal import.

De-duplication

channel_relative_entropy.py still defines its own _make_grid / _find_mu / _find_lambda / _make_delta / _make_gamma on master. Please confirm the channel module switches to the shared helpers and the inline copies are removed, otherwise the extraction only de-duplicates half of it.

Accuracy / tolerances

The space_optimized branch loosens the test tolerances substantially (joint-affine from rtol=5e-4 to rtol=5e-2, and the parametrized accuracy check allows abs(err12) <= 1.5e-1). That reflects the method being coarse at the default epsilon_dec=1e-2 rather than a bug, but the docstring should state the accuracy-vs-space tradeoff explicitly and note that epsilon_dec can be decreased for more accuracy.

Minor

find_mu and find_lambda solve an SDP to obtain what are just the extreme generalized eigenvalues of the pencil (X, Y). scipy.linalg.eigh(X, Y) returns those directly (mu = min, lam = max) without a solver call, which also removes the OPTIMAL_INACCURATE path for these two. Optional, but it drops two SDP solves per evaluation.

simsaidan:
@vprusso i added a new commit which should address all feedback points. Let me know if I should make any other changes!

vprusso (approved):
Thanks, this addresses everything from the earlier review:

- The module is now the public toqito/cones/integral_relative_entropy.py, exported from the package, with the grid/coefficient builders kept private (_make_grid, _make_delta, _make_gamma).
- The duplicated helpers are gone from channel_relative_entropy.py, which now shares the ones in the cones module.
- The accuracy/space tradeoff is documented, including the note to lower epsilon_dec for more accuracy.
- mu/lambda are now generalized eigenvalues via scipy (eigh with an eig fallback for singular inputs) instead of auxiliary SDPs, which is cleaner and removes the OPTIMAL_INACCURATE path. The bounds are correct: lam = max eig(X, Y), mu = min eig(Y, X).

CI is green across the matrix and the new test file covers the error paths and the eig fallback. Looks good to merge.

One minor, non-blocking note: channel_relative_entropy imports the underscore-prefixed helpers (_make_grid, _sandwich_parameters, ...) across package boundaries. That is fine since they are genuinely shared internals and stay out of the public API, just flagging it as the one slightly unusual bit.

simsaidan:
@vprusso If you think it's non-blocking, can we merge as is?

PR #1711, opened 2026-07-02
Implement the rel_entr_quad, tsallis entropy, and tsallis relative en…

simsaidan:
…tropy

## Changes

  -  [ ] Change 1

## Checklist


simsaidan:
Hey @vprusso . I have this PR, which implements the final piece of cvxquad (rel_entr_quad) as well as tsallis entropy and tsallis relative entropy which were in Fawzi's paper but not the package. Looking forward to your review!

simsaidan:
@vprusso I added a new commit to this PR which has examples in the docstrings for each new function.

vprusso (changes requested):
Hi @simsaidan — a heads-up on some maintainer changes I pushed, and one thing that still needs your attention before this can merge.

**What I pushed to your branch** (via maintainer-edit):

- Merged the latest `master` — your branch was ~64 commits behind, which is what GitHub was flagging as a conflict.
- Fixed the failing `code-style` check: reformatted the new `cones` files and silenced two `PLC0415` warnings on the `quantum_relative_entropy` imports in `tsallis_relative_entropy.py`. Those imports are intentionally deferred to avoid a circular import (`cones.tsallis_relative_entropy` → `state_props.quantum_relative_entropy` → back into `cones`), so I kept them function-local with a `# noqa: PLC0415` and a comment rather than hoisting them. `code-style` is green again and the pytest matrix is re-running.

**What still needs fixing before merge** — this is separate from CI. The test suite passes, but it doesn't cover this case, so green CI here does not mean the PR is correct:

The affine-CVXPY path returns a wrong value. When the inputs are affine expressions that contain free variables, the function builds and *solves* its own `Minimize(tau)` problem and returns that scalar. So it optimizes over the free variables instead of evaluating at their `.value` — the result neither reflects the input nor composes into a larger CVXPY problem. Concretely:

```python
import numpy as np, cvxpy
from toqito.cones import tsallis_relative_entropy

B = np.diag([0.6, 0.4]); A = np.diag([0.7, 0.3])
print(tsallis_relative_entropy(A, B, 0.5))     # 0.011032   (correct)

X = cvxpy.Variable((2, 2), symmetric=True); X.value = A
print(tsallis_relative_entropy(X, B, 0.5))     # -0.5       (wrong, and negative is impossible here)
```

Same data, two very different answers: the numeric path gives ~0.011, the affine path gives -0.5. The same "build-a-problem, solve it, return a float" shape appears in `relative_entropy_quadrature` and `tsallis_entropy`, so they likely share the issue — worth checking all three.

The fix is really a design decision about what the affine path is *for*:

- If it is meant to be **embedded in a larger SDP** (which is what the epigraph description in the docstring suggests), it should return the epigraph *constraints* plus the auxiliary `tau`/`s` handles, not solve and return a float.
- If it is meant as a **convenience to evaluate at a point**, it should read `.value` and route through the numeric path.

Right now it does neither, which is where the -0.5 comes from.

Happy to talk through the approach if that's useful — the geometric-mean / Lieb-Ando machinery underneath looks solid, it's this affine entry point that needs rethinking. Thanks for the work on this.

simsaidan:
@vprusso Thanks for catching that. I think the first option is better since it is truer to the original package. I'm guessing this would affect most of the cones package right? I don't see why it would be limited to these three.

vprusso:
Good instinct, and you're right that it isn't limited to these three. The cones package splits into two groups:

- **Composable cone constructors** (`geometric_mean_hypo_cone`, `geometric_mean_epi_cone`, `operator_relative_entropy_epi_cone`, `geometric_mean`): these return CVXPY *constraints* and are fine.
- **Solve-internally, return-a-float** functions: your three, plus the pre-existing `lieb_ando`, `trace_matrix_power`, `ln_quantum_entropy`, `trace_matrix_log`, and `integral_relative_entropy`. These all share the pattern, so they all mishandle affine inputs the same way. I confirmed two: `tsallis_relative_entropy` returns `-0.5`, and `lieb_ando` comes back `unbounded`.

Since the first option (returning epigraph constraints, cvxquad-style) is the right long-term design but is a breaking change that touches callers like `quantum_relative_entropy` and really needs to be done consistently across all of these at once, I don't think it belongs in this PR. Rewriting the whole cones package would be a lot to bundle with "add three functions," and the pre-existing five are prior debt this PR shouldn't be on the hook for.

So let's scope this PR down:

- Keep the numeric path (both inputs are `numpy.ndarray`), which is correct and is the common use.
- On the affine path, rather than silently returning a wrong value, raise a clear error when the input contains free variables (e.g. "affine/variable inputs are not yet supported; pass numeric matrices"). Constant expressions that carry a concrete `.value` can still be routed through the numeric path.

That lands the three functions correctly for how they'll actually be used, without shipping the `-0.5` trap.

I've opened #1846 to track the proper composable redesign (option 1) across the whole package, with the reproductions and the caveats about callers and tests. Happy to review once you've scoped this one down. Thanks for thinking it through.

simsaidan:
@vprusso I like the two PR plan. I added a new commit which removes the affine support for these three functions.

vprusso (approved):
Re-reviewed after the scope-down. This addresses the blocker cleanly:

- The broken affine SDP path is removed; affine/free-variable CVXPY inputs now raise a clear error (`Affine or variable CVXPY inputs are not yet supported; pass numeric matrices`) instead of silently returning a wrong value like -0.5. Constant CVXPY expressions still route through the numeric path.
- I verified the numeric path against the definition S_t(A||B) = (1/t) tr(A - A^{1-t} B^t) over a battery of random density matrices at t = 0, 0.25, 0.5, 0.75, 1.0 — agreement to ~1e-15, including both boundaries (t->0 recovers the quantum relative entropy; t=1 gives 0, both continuous). No t=1 discontinuity on this path.
- The citation resolves correctly to Fawzi & Saunderson, matrix geometric means / semidefinite optimization, which is the right source for the SDP construction.
- code-style is fixed (formatting).

The broader composable redesign for the whole cones package is tracked in #1846. Approving; this is good to merge once CI is green. Thanks for scoping it down.

PR #1738, opened 2026-07-06
Improve Matsumoto fidelity singular handling

### LINKED ISSUE #1724: Clarify and improve singular-state handling in Matsumoto fidelity
## Problem
`toqito/state_metrics/matsumoto_fidelity.py` has two issues around singular inputs.

1. The displayed formula for the singular-state limit appears to omit `sigma` in the second operand.
2. The implementation uses a fixed one-sided `1e-7` regularization. That is scale-dependent and does not match the stated two-sided limiting expression.

## Suggested fix
- Correct the documentation formula.
- Revisit the singular-case implementation so it matches the documented mathematical definition.
- Add tests covering singular positive semidefinite inputs, including scale-sensitive cases.

vprusso:
## Summary
- correct the singular-state limit formula in the Matsumoto fidelity docstring
- replace fixed epsilon regularization with a pseudoinverse square-root path for singular states
- add regression tests for identical and orthogonal singular pure states

## Validation
- `uv run pytest toqito/state_metrics/tests/test_matsumoto_fidelity.py -q`

Closes #1724

### DIFF
--- a/toqito/state_metrics/matsumoto_fidelity.py
+++ b/toqito/state_metrics/matsumoto_fidelity.py
@@ -27,7 +27,7 @@ def matsumoto_fidelity(rho: np.ndarray, sigma: np.ndarray) -> float | np.floatin
     For singular states it is defined by the limit
 
     \[
-        \rho\#\sigma = \lim_{\epsilon\to0}(\rho+\epsilon\mathbb{I})\#(+\epsilon\mathbb{I}).
+        \rho\#\sigma = \lim_{\epsilon\to0}(\rho+\epsilon\mathbb{I})\#(\sigma+\epsilon\mathbb{I}).
     \]
 
     The return is a value between \(0\) and \(1\), with \(0\) corresponding to matrices `rho` and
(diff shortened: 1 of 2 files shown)


vprusso:
Pushed a follow-up commit addressing a correctness bug in the singular path. The pseudoinverse square-root expression is only exact when the inverted (larger-determinant) state is invertible; when **both** states are singular it drops the part of `sigma` outside `supp(rho)` and diverges from the limit definition (e.g. `matsumoto_fidelity(|0><0|, |+><+|)` returned `1/sqrt(2)` where the two-sided limit, and the SDP branch already in this function, give 0).

The fix falls back to the exact geometric-mean characterization `tr(rho # sigma) = max{ tr(W) : [[rho, W], [W, sigma]] >> 0 }` (solved with CLARABEL) whenever the inverted state is also singular, and keeps the fast closed form otherwise. Verified against the eps-regularized limit on the non-commuting pair (~0), a rank-2 pair (0.5), plus the existing identical (1) and orthogonal (0) cases; added regression tests for the first two. Full `state_metrics` suite passes.

PR #1746, opened 2026-07-06
Fix README logo rendering

vprusso:
## Summary
- switch README logo images from external raw URLs to repository-relative SVG paths
- add alt text for the logo image

## Validation
- verified README still points to existing logo assets under `docs/content/figures/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center">
  <picture>
-   <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/vprusso/toqito/master/docs/content/figures/logo-dark.svg" width="60%">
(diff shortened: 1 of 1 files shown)


vprusso:
Closing: the premise doesn't hold and the change would cause a regression.

- The existing logo URLs are not broken: `curl -sI` on both the light and dark `raw.githubusercontent.com` SVGs returns HTTP 200 with `content-type: image/svg+xml`.
- Switching to repo-relative image paths breaks the logo on the PyPI project page. `README.md` is the package long_description (`readme = "README.md"` in pyproject.toml), and PyPI does not resolve repository-relative paths, so they 404 there while the absolute raw URLs render on both GitHub and PyPI.

The only genuine improvement here is the added `alt` text, which can be applied on its own without changing the image URLs.

Issue #1799, opened 2026-07-06
docs: add an Examples section to quantum_conditional_entropy


vprusso:
`toqito/state_props/quantum_conditional_entropy.py` documents theory, Args, and Returns but has no **Examples** section. Add an `Examples` subsection with a runnable ```python exec``` block computing H(A|B) for a Bell state (should be negative) and a product state (non-negative), printing both.

**How to verify:** `uv run mkdocs build -f docs/mkdocs.yml` after `uv sync --group docs`.

---
*This is a **good first issue** — small and self-contained. See [CONTRIBUTING](https://github.com/vprusso/toqito/blob/master/CONTRIBUTING.md) and the [contributing guide](https://vprusso.github.io/toqito/contributing-guide/) to get set up (`uv sync`, `uv run pytest`, `uv run ruff check`). Comment here if you'd like to take it or have questions.*

vprusso:
Already resolved: `quantum_conditional_entropy` gained the requested Examples section (Bell state H(A|B) negative, product state non-negative, both printed) in #1804, which has merged. Closing as done.

PR #1804, opened 2026-07-07
docs: add quantum conditional entropy example

KSHITIZ6341:
Solved: vprusso/toqito#1799.
I added a runnable documentation example for quantum_conditional_entropy showing conditional entropy for a Bell state and a product state. The clean branch is pushed to your fork as docs-quantum-conditional-entropy-example-pr; no PR has been opened yet.
Validation completed: focused ruff, direct example execution, target pytest, and docs build from docs/.

### DIFF
--- a/toqito/state_props/quantum_conditional_entropy.py
+++ b/toqito/state_props/quantum_conditional_entropy.py
@@ -55,8 +55,6 @@ def quantum_conditional_entropy(
         Compute \(H(A|B)\) for a Bell state and for a product state:
 
         ```python exec="1" source="above" result="text"
-        import warnings
-
         import numpy as np
 
         from toqito.state_props import quantum_conditional_entropy
@@ -65,10 +63,8 @@ def quantum_conditional_entropy(
         bell_state = np.outer(psi, psi.conj())
         product_state = np.diag([1 / 2, 0, 1 / 2, 0])
 
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", message="The logm input matrix is exactly singular.")
-            bell_entropy = quantum_conditional_entropy(bell_state, [2, 2], sys=0)
-            product_entropy = quantum_conditional_entropy(product_state, [2, 2], sys=0)
+        bell_entropy = quantum_conditional_entropy(bell_state, [2, 2], sys=0)
+        product_entropy = quantum_conditional_entropy(product_state, [2, 2], sys=0)
 
         print(f"Bell state H(A|B): {bell_entropy:.6f}")
         print(f"Product state H(A|B): {product_entropy:.6f}")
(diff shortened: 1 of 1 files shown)


vprusso:
Thanks for this, and for picking up the good-first-issue! I ran the example and the math is spot on: `H(A|B) = -ln 2` for the Bell state (correctly negative) and `+ln 2` for the product state. 👍

One thing to sort out first: the `warnings.catch_warnings()` block is a workaround for a cosmetic `scipy` "logm input matrix is exactly singular" warning. That warning actually comes from deeper down (`quantum_relative_entropy`'s numeric `logm` path), not from your example, so rather than have the boilerplate live in the docstring we're fixing it at the source in #1805. Once that merges, `quantum_conditional_entropy` won't emit the warning at all.

So the ask here is small: after #1805 lands, could you simplify the example to drop the `import warnings` / `with warnings.catch_warnings(): ...` wrapper and just call `quantum_conditional_entropy` directly? Something like:

```python exec="1" source="above" result="text"
import numpy as np

from toqito.state_props import quantum_conditional_entropy

psi = np.array([1, 0, 0, 1]) / np.sqrt(2)
bell_state = np.outer(psi, psi.conj())
product_state = np.diag([1 / 2, 0, 1 / 2, 0])

bell_entropy = quantum_conditional_entropy(bell_state, [2, 2], sys=0)
product_entropy = quantum_conditional_entropy(product_state, [2, 2], sys=0)

print(f"Bell state H(A|B): {bell_entropy:.6f}")
print(f"Product state H(A|B): {product_entropy:.6f}")
```

No rush at all, since it depends on #1805 merging first. Really nice contribution!

vprusso (changes requested):
Good news: #1805 has merged, which fixes the singular-`logm` warning at its source in `quantum_relative_entropy`. That means `quantum_conditional_entropy` no longer emits the warning, so the `import warnings` / `with warnings.catch_warnings(): ...` wrapper in this example is now unnecessary. Could you simplify the block to call `quantum_conditional_entropy` directly (I left the exact simplified version in an earlier comment)? The math is correct; just dropping the boilerplate and it's ready.

KSHITIZ6341:
Thanks, this is updated now. I removed the `warnings` import and `catch_warnings` wrapper so the example calls `quantum_conditional_entropy` directly.

Validation run:
- `uv run pytest toqito/state_props/tests/test_quantum_conditional_entropy.py`
- `uv run --with ruff ruff check toqito/state_props/quantum_conditional_entropy.py`
- `uv run --with ruff ruff format --check toqito/state_props/quantum_conditional_entropy.py`
- `git diff --check`

vprusso (approved):
Thanks for simplifying. Now that #1805 fixed the singular-logm warning at its source, dropping the warnings-suppression boilerplate is exactly right. Verified the example runs and prints the correct conditional entropies (Bell H(A|B) = -ln(2), product H(A|B) = +ln(2)). Docstring-only. Approving.

PR #1822, opened 2026-07-07
polish: include the offending t value in geometric_mean's range error

### LINKED ISSUE #1814: polish: include the offending t value in geometric_mean's range error
`toqito/cones/geometric_mean.py` raises `ValueError("The weight must be in the range [-1, 2].")` without reporting the actual value passed. Include it, e.g. `f"The weight t must be in the range [-1, 2]; got {t}."`, and pin the message with a `match=` in the test.

---

dchaudhari7177:
## Description
Closes #1814

The out-of-range weight error now reports the value that was passed: `The weight t must be in the range [-1, 2]; got 2.01.` Both out-of-range test rows pin the exact message (the test already matches with `re.escape`).

## Changes
  -  [x] Include the offending `t` in the `ValueError` message
  -  [x] Pin the new message in the two out-of-range test cases

## Checklist
  -  [x] Verify all previous and newly added unit tests pass in `pytest` (20/20 in `test_geometric_mean.py`).

### DIFF
--- a/toqito/cones/geometric_mean.py
+++ b/toqito/cones/geometric_mean.py
@@ -42,7 +42,7 @@ def geometric_mean(mat_a: np.ndarray, mat_b: np.ndarray, t: float) -> np.ndarray
         raise ValueError("The matrices must be the same size.")
     _require_square_2d(mat_a, "The matrices")
     if t < -1 or t > 2:
-        raise ValueError("The weight must be in the range [-1, 2].")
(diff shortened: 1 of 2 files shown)


vprusso (approved):
Verified the message `The weight t must be in the range [-1, 2]; got -1.01.` / `got 2.01.` matches both updated test rows (float formatting checks out). Thanks for #1814!

PR #1824, opened 2026-07-07
test: pin exact messages for XORGame constructor validation

### LINKED ISSUE #1811: test: cover the XORGame constructor validation branches
`toqito/nonlocal_games/xor_game.py`'s `XORGame.__init__` has three `ValueError` validation branches that are currently untested (`test_xor_game.py` has no `pytest.raises`/`assert_raises` at all):
- `prob_mat` and `pred_mat` shape mismatch,
- negative `prob_mat` entries,
- `prob_mat` not summing to 1.

Add tests in `toqito/nonlocal_games/tests/test_xor_game.py` that trigger each with `pytest.raises(ValueError, match=...)`.

---

dchaudhari7177:
## Description
Closes #1811

One note on the issue premise: `test_xor_game.py` does already exercise these branches, but only through loose `self.assertRaises(ValueError)` blocks (`test_negative_prob_mat`, `test_invalid_prob_mat`, `test_non_square_prob_mat`, `test_zero_prob_mat`) that pass no matter which of the three errors fires — the earlier grep for `assert_raises` misses the camelCase `assertRaises`. What was genuinely missing is exact-message coverage, which this PR adds as requested: parametrized `pytest.raises(ValueError, match=re.escape(...))` tests for each branch.

## Changes
  -  [x] Parametrized test triggering the shape-mismatch, negative-entry, and sum-to-1 branches, each pinned to its exact error message

## Checklist
  -  [x] Verify all previous and newly added unit tests pass in `pytest` (19/19 in `test_xor_game.py`).

### DIFF
--- a/toqito/nonlocal_games/tests/test_xor_game.py
+++ b/toqito/nonlocal_games/tests/test_xor_game.py
@@ -1,8 +1,10 @@
 """Tests for XORGame class."""
 
+import re
 import unittest
 
 import numpy as np
+import pytest
 
 from toqito.nonlocal_games.xor_game import XORGame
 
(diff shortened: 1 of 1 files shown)


vprusso (approved):
Verified by execution: all three cases raise the exact pinned messages, and the inputs hit the intended branch given the shape → non-negative → sum-to-1 validation order (e.g. `1/4*ones((2,2))` is a valid probability matrix, so case 1 genuinely exercises the shape check). Good call on the issue premise too — the old `assertRaises(ValueError)` blocks passed regardless of which error fired; pinning the messages is exactly the intended improvement. Thanks for #1811!

PR #1823, opened 2026-07-07
polish: report the mismatched shapes in vectors_to_gram_matrix

dchaudhari7177:
## Description
Closes #1813

The shape-mismatch error now tells the caller which shape differs: `All vectors must have the same shape; expected (3,), got (2,).` The check walks the list and reports the first offending shape. The invalid-input tests switched from a bare `assert_raises` to `pytest.raises(..., match=re.escape(...))` asserting the exact message for both the vector and density-matrix cases.

## Changes
  -  [x] Report expected and offending shapes in the `ValueError`
  -  [x] Assert the exact message in `test_vectors_to_gram_matrix_invalid_input`

## Checklist
  -  [x] Use `ruff` for errors related to code style and formatting.
  -  [x] Verify all previous and newly added unit tests pass in `pytest` (6/6 in `test_vectors_to_gram_matrix.py`).
  -  [x] Check the documentation build does not lead to any failures.

PR #1829, opened 2026-07-07
Report the mismatched shapes in vectors_to_gram_matrix

### LINKED ISSUE #1813: polish: report the mismatched shapes in vectors_to_gram_matrix
`toqito/matrix_ops/vectors_to_gram_matrix.py` raises `ValueError("All vectors must be of the same shape.")` without telling the caller which shape differs. Include the expected and offending shapes, e.g. `f"All vectors must have the same shape; expected {vectors[0].shape}, got {v.shape}."`, and add/adjust a test asserting the clearer message.

---

delaidam:
Closes #1813

The `ValueError` raised in `vectors_to_gram_matrix` when input vectors or
density matrices have inconsistent shapes gave no indication of which
shapes actually differed. Now reports the expected shape (from the first
input) and the offending one, e.g.:

  All vectors must have the same shape; expected (3,), got (2,).

Also pinned the message in the existing invalid-input test with `match=`
(previously it only asserted `ValueError` was raised, not its content).

## Checklist
- [x] `uv run ruff check` and `uv run ruff format --check` pass
- [x] `uv run pytest` passes (full suite: 3442 passed, 59 skipped, 3 xfailed)
- [x] New/changed lines are covered by the updated test

### DIFF
--- a/toqito/matrix_ops/tests/test_vectors_to_gram_matrix.py
+++ b/toqito/matrix_ops/tests/test_vectors_to_gram_matrix.py
@@ -1,5 +1,7 @@
 """Test vectors_to_gram_matrix."""
 
+import re
+
 import numpy as np
 import pytest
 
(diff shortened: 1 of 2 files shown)


vprusso:
Thanks for this, @delaidam — your implementation is correct and does exactly what #1813 asked. Unfortunately it overlaps with #1823, which was opened a bit earlier for the same issue and has already been reviewed and approved, so I'm going to close this one as a duplicate to avoid merging the same change twice. This is no reflection on your work — it's just the luck of two people picking up the same issue at the same time.

If you'd like to keep contributing (please do!), there are plenty of other open tasks in the [good first issue queue](https://github.com/vprusso/toqito/labels/good%20first%20issue) that nobody has claimed yet — a quick comment on one to say you're taking it will help avoid another collision. Thanks again, and sorry for the overlap!

PR #1838, opened 2026-07-07
Consolidate redundant XORGame constructor validation tests

delaidam:
Follow-up to #1811, per maintainer feedback there.

test_negative_prob_mat, test_invalid_prob_mat, test_non_square_prob_mat,
and test_zero_prob_mat each covered one of the three ValueError branches
in XORGame.__init__ without asserting message content, and called
quantum_value() after the expected exception (dead code that never runs).

test_xor_game_invalid_input (#1824) already covers the same three
branches with pytest.raises(..., match=...), making these four redundant.
Removed them, no coverage loss (verified: same 3 branches still exercised,
full toqito/nonlocal_games/ suite still passes).

## Checklist
- [x] `uv run ruff check` and `uv run ruff format --check` pass
- [x] `uv run pytest toqito/nonlocal_games/tests/test_xor_game.py` - 15 passed
- [x] `uv run pytest toqito/nonlocal_games/` — 81 passed, 11 skipped, 1 xfailed

### DIFF
--- a/toqito/nonlocal_games/tests/test_xor_game.py
+++ b/toqito/nonlocal_games/tests/test_xor_game.py
@@ -136,42 +136,6 @@ class TestXORGame(unittest.TestCase):
         res = game.classical_value()
         self.assertEqual(res, 0.625)
 
-    def test_negative_prob_mat(self):
-        """Tests for invalid negative probability matrix."""
-        with self.assertRaises(ValueError):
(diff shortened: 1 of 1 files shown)


vprusso (approved):
Verified the parametrized `test_xor_game_invalid_input` still covers all three `__init__` validation branches (shape / negative / sum-to-1), so removing the four redundant unittest-style tests loses no coverage. 15 tests pass, ruff clean. Thanks for taking the consolidation.

PR #1839, opened 2026-07-07
fix(perms): handle row vectors in permute_systems with explicit dim (#1827)

### LINKED ISSUE #1827: bug: permute_systems rejects a (1, D) row vector when dim is given explicitly
`toqito/perms/permute_systems.py`

A row-vector input of shape `(1, D)` is rejected with `InvalidDim` when an explicit `dim` is passed, even though the exact same data as a 1-D array `(D,)` or a column vector `(D, 1)` is accepted and permuted correctly — and the `dim=None` path already handles the `(1, D)` row. Only the explicit-`dim` path is inconsistent.

**Reproduction:**
```python
import numpy as np
from toqito.perms import permute_systems

perm, dim = [1, 2, 0], [2, 2, 2]
print(permute_systems(np.arange(8), perm, dim))                         # 1-D  -> [0 4 1 5 2 6 3 7]
print(permute_systems(np.arange(8).reshape(8, 1), perm, dim).ravel())   # col  -> [0 4 1 5 2 6 3 7]
print(permute_systems(np.arange(8).reshape(1, 8), perm, dim))           # row  -> raises InvalidDim
```

**Observed:** the 1-D and column forms return `[0 4 1 5 2 6 3 7]`; the `(1, 8)` row raises `ValueError: InvalidDim: The dimensions specified in DIM do not agree with the size of X.`  **Expected:** the `(1, 8)` row (exactly 8 = 2*2*2 elements) should permute to `[0 4 1 5 2 6 3 7]` like the other orientations.

The input-orientation logic treats every 2-D array as a column vector; work out why the `(1, D)` row is misclassified on the explicit-`dim` path (but not on `dim=None`) and make the two paths consistent.

---
*The failing case and expected result are given, but the fix is intentionally left open — it requires understanding the relevant code. See the [contributing guide](https://vprusso.github.io/toqito/contributing-guide/) to get set up.*

Mahizhan-S:
Closes #1827

Fixes an inconsistency where `permute_systems` rejected row-vector inputs of shape `(1, D)` when `dim` was explicitly provided.

The fix makes row vectors behave consistently with 1-D arrays and column vectors.

A regression test was added to verify that `(1, D)` inputs are correctly permuted when `dim` is specified.

### DIFF
--- a/toqito/perms/permute_systems.py
+++ b/toqito/perms/permute_systems.py
@@ -152,12 +152,15 @@ def permute_systems(
         dim_arr = dim
 
     if is_vec:
-        # 1 if column vector
-        if len(input_mat.shape) > 1:
-            vec_orien = 0
-        # 2 if row vector
-        else:
+        if len(input_mat.shape) == 1:
+            # 1-D array: treat as row vector
             vec_orien = 1
+        elif input_mat.shape[0] == 1:
+            # (1, D) row vector
+            vec_orien = 1
+        else:
+            # (D, 1) column vector
+            vec_orien = 0
 
     if len(dim_arr.shape) == 1:
         # Force dim to be a row vector.
--- a/toqito/perms/tests/test_permute_systems.py
+++ b/toqito/perms/tests/test_permute_systems.py
@@ -166,4 +166,3 @@ def test_permute_systems_invalid_dim_mismatch_perm():
     with np.testing.assert_raises(ValueError):
         test_input_mat = np.array([1, 2, 3, 4])
         permute_systems(test_input_mat, [2, 1, 3, 4])
-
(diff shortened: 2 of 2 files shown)


vprusso (commented):
The fix itself is correct: `(1, D)` row vectors with an explicit `dim` now permute consistently with the 1-D and `(D, 1)` column forms, there's no regression on the other orientations (outputs are byte-identical to master), and 128 tests pass including the `swap` / `partial_trace` / `partial_transpose` consumers.

One thing before merge: the PR description says a regression test was added, but the net change to the test file is empty (a trailing blank line was added in one commit and removed in the next), so nothing actually guards this fix. Could you add a small case pinning it, e.g.

```python
assert np.array_equal(
    permute_systems(np.arange(8).reshape(1, 8), [1, 2, 0], [2, 2, 2]),
    np.array([0, 4, 1, 5, 2, 6, 3, 7]),
)
```

With that added it's good to go.

Mahizhan-S:
Done

vprusso (approved):
Thanks for adding the regression test — `test_permute_systems_row_vector_with_explicit_dim` pins exactly the fixed case and passes (full file: 8 passed). The fix is correct with no regression on the other orientations or the consumer modules. Merging.

PR #1840, opened 2026-07-07
polish: include the offending index in the gell_mann error message

### LINKED ISSUE #1795: polish: include the offending index in the gell_mann error message
`toqito/matrices/gell_mann.py` (around line 112) raises `ValueError("Gell-Mann index values can only be values from 0 to 8 (inclusive).")` without reporting the bad value. Include it, e.g. `f"Gell-Mann index must be an integer in [0, 8]; got {ind}."`. The existing `gell_mann(9)` test uses `assert_raises` with no `match`; add a `match=` to pin the improved message.

---

dchaudhari7177:
## Description
Closes #1795

The out-of-range Gell-Mann index error now reports the value that was passed: `Gell-Mann index must be an integer in [0, 8]; got 9.` The existing `gell_mann(9)` test was upgraded from a bare `assert_raises` to `pytest.raises(ValueError, match=re.escape(...))` pinning the exact message.

## Changes
  -  [x] Echo the offending `ind` in the `ValueError`
  -  [x] Pin the new message in `test_gell_mann_invalid_idx`

## Checklist

### DIFF
--- a/toqito/matrices/gell_mann.py
+++ b/toqito/matrices/gell_mann.py
@@ -109,7 +109,7 @@ def gell_mann(ind: int, is_sparse: bool = False) -> np.ndarray | csr_array:
     elif ind == 8:
         gm_op = np.array([[1, 0, 0], [0, 1, 0], [0, 0, -2]]) / np.sqrt(3)
     else:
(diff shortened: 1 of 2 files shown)


vprusso (approved):
Verified the generated message (with the offending index) exactly matches the updated test; test file passes.

PR #1841, opened 2026-07-07
polish: report the bad subsystem index and valid range in partial_trace

### LINKED ISSUE #1796: polish: report the bad subsystem index and valid range in partial_trace
`toqito/matrix_ops/partial_trace.py` raises the identical vague `ValueError("Subsystem indices in `sys` are out of bounds.")` in two branches (around lines 170 and 176) without saying which index was wrong or the valid range. Improve both to include the offending `sys` value and the valid range, e.g. `f"Subsystem index {s} is out of bounds; must be in [0, {num_sys})."`. Add/adjust a test to assert the clearer message.

---

dchaudhari7177:
## Description
Closes #1796

Both out-of-bounds branches in `partial_trace` raised the same vague message, `"Subsystem indices in `sys` are out of bounds."`, without saying which index was wrong. Each now reports the offending index and the valid range, e.g. `"Subsystem index 2 is out of bounds; must be in [0, 2)."` The list branch was changed from `any(...)` to an explicit loop so it can name the specific bad index. `test_sys_out_of_bounds` now asserts the specific index for the int, list, and negative cases.

## Changes
  -  [x] Report the offending index + valid range in both branches
  -  [x] Assert the specific message per case in the test

## Checklist
  -  [x] Verify all previous and newly added unit tests pass in `pytest` (72/72 in `test_partial_trace.py`).

### DIFF
--- a/toqito/matrix_ops/partial_trace.py
+++ b/toqito/matrix_ops/partial_trace.py
@@ -167,13 +167,14 @@ def partial_trace(
     # Validate sys indices and compute subsystem product.
     if isinstance(sys, int):
         if sys < 0 or sys >= num_sys:
-            raise ValueError("Subsystem indices in `sys` are out of bounds.")
+            raise ValueError(f"Subsystem index {sys} is out of bounds; must be in [0, {num_sys}).")
         prod_dim_sys = dim[sys]
         sys = np.array([sys])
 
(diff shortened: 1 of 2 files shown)


vprusso (approved):
Verified the message (offending index + valid range) matches the updated test for both the int and list/tuple branches; tests pass.

Issue #1826, opened 2026-07-07
bug: is_mutually_unbiased_basis accepts non-orthonormal sets (incomplete check)


vprusso:
`toqito/state_props/is_mutually_unbiased_basis.py`

The function is documented to check whether the vectors form *mutually unbiased **bases***, and its definition (in the docstring) requires each \(\mathcal{B}_x\) to be an **orthonormal basis**. But it only verifies the cross-basis condition \(|\langle u, v\rangle|^2 = 1/d\); it never checks that each group of vectors is actually an orthonormal basis. So a degenerate set with repeated (or otherwise non-orthonormal) vectors is wrongly accepted.

**Reproduction:**
```python
import numpy as np
from toqito.states import basis
from toqito.state_props import is_mutually_unbiased_basis

e0, e1 = basis(2, 0), basis(2, 1)
v = 1 / np.sqrt(2) * (e0 + e1)
# Second 'basis' is {v, v}: two identical vectors, so NOT an orthonormal basis of C^2.
print(is_mutually_unbiased_basis([e0, e1, v, v]))
```

**Observed:** `True`  **Expected:** `False` — `{v, v}` cannot be an orthonormal basis (the two vectors are identical and do not span \(\mathbb{C}^2\)); the cross inner products all equal 1/2, so the incomplete check passes.

Figure out where the per-basis orthonormality validation belongs (each consecutive block of `dim` vectors should be an orthonormal basis) and add it, with a sensible numerical tolerance. Note: this gap is independent of the Gram-matrix refactor in #1767 / #1776, which only touches the existing cross-basis check.

---
*The failing case and expected result are given, but the fix is intentionally left open — it requires understanding the relevant code. See the [contributing guide](https://vprusso.github.io/toqito/contributing-guide/) to get set up.*

vprusso:
Fixed by #1845, which validates that each basis block is orthonormal (the diagonal Gram blocks equal the identity) in addition to the cross-basis unbiasedness check. The reported case `[e0, e1, v, v]` now returns `False`.

PR #1845, opened 2026-07-08
fix: validate orthonormality in mutually unbiased basis check (#1826)

Mahizhan-S:
## Problem

`is_mutually_unbiased_basis` verified only the mutual unbiasedness condition between different bases:

$|\langle u, v\rangle|^2 = 1/dim$

However, it did not verify that each group of `dim` vectors was actually an orthonormal basis. This allowed invalid inputs containing repeated or non-orthonormal vectors to incorrectly pass the check.

For example, a set like `{v, v}` could satisfy the cross-basis inner product condition while not forming a valid basis.

## Changes

- Added an orthonormality validation step for each consecutive block of `dim` vectors.
- Reused the existing Gram matrix calculation to check whether each basis block matches the identity matrix.
- Added a numerical tolerance (`atol=1e-10`) for floating-point comparisons.
- Invalid non-orthonormal basis sets now correctly return `False`.

This ensures that the function now validates both requirements of mutually unbiased bases:
1. Each individual set of vectors is an orthonormal basis.
2. Different bases satisfy the mutual unbiasedness condition.

### DIFF
--- a/toqito/state_props/is_mutually_unbiased_basis.py
+++ b/toqito/state_props/is_mutually_unbiased_basis.py
@@ -95,8 +95,21 @@ def is_mutually_unbiased_basis(vectors: list[np.ndarray | list[float | Any]]) ->
     mat = np.column_stack([np.asarray(vector).reshape(-1) for vector in vectors])
     gram = np.abs(mat.conj().T @ mat) ** 2
 
+    # Check that each consecutive block of `dim` vectors forms an orthonormal basis.
+    identity = np.eye(dim)
+
+    for basis_start in range(0, num_vectors, dim):
+        basis_block = gram[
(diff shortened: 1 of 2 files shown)


vprusso (approved):
Verified this fixes #1826: each diagonal `dim x dim` block of the squared-Gram matrix equals the identity exactly when that basis is orthonormal (|<v_i,v_j>|^2 = delta_ij), so validating those blocks is the right check. Confirmed the reported case `[e0, e1, v, v]` now returns `False`, a valid complete set of MUBs still returns `True`, and ragged input (num_vectors not a multiple of dim) is handled by the existing guard rather than crashing on the non-square block. Regression test matches the issue's reproduction and CI is green. Nice fix.

PR #1848, opened 2026-07-09
Fix swap subsystem count for rectangular dimensions

### LINKED ISSUE #1828: bug: swap miscounts subsystems for the documented 2-row (rectangular) dim form
`toqito/perms/swap.py`

`swap`'s docstring documents a 2-row `dim` form for non-square operators (row dimensions in the first row, column dimensions in the second) over "possibly more than 2" subsystems. But `swap` determines the subsystem count from the number of **rows** of `dim` (always 2 in this form) instead of the number of **columns** (the true subsystem count), so its bounds check rejects valid `sys` indices. `permute_systems` accepts the same operator and `dim`.

**Reproduction:**
```python
import numpy as np
from toqito.perms import swap, permute_systems

M = np.arange(96).reshape(12, 8).astype(float)  # row dims [2,2,3]=12, col dims [2,2,2]=8
dim2 = [[2, 2, 3], [2, 2, 2]]
print(permute_systems(M, [2, 1, 0], dim2).shape)  # -> (12, 8), works
print(swap(M, [1, 3], dim2))                       # -> raises InvalidSys
```

**Observed:** `permute_systems(...)` returns shape `(12, 8)`; `swap(M, [1, 3], dim2)` raises `ValueError: InvalidSys: The subsystems in sys must be between 1 and len(dim). inclusive.`  **Expected:** `swap` should treat this as a 3-subsystem operator, accept `sys=[1, 3]`, and return the `(12, 8)` operator with subsystems 1 and 3 swapped.

Find where `swap` computes the number of subsystems and make it use the actual subsystem count for the documented 2-row `dim` form (compare with how `permute_systems` handles it).

---
*The failing case and expected result are given, but the fix is intentionally left open — it requires understanding the relevant code. See the [contributing guide](https://vprusso.github.io/toqito/contributing-guide/) to get set up.*

KSHITIZ6341:
## Description
Fixes #1828.

The documented two-row rectangular `dim` form stores one subsystem per column. `swap` was counting rows instead, so valid subsystem indices beyond 2 were rejected before reaching `permute_systems`.

## Changes
- Count subsystems from the column count when `dim` is a two-row rectangular dimension array.
- Add regression coverage for swapping systems 1 and 3 of a rectangular 3-subsystem operator.

## Checklist
- [ ] Check the documentation build does not lead to any failures.

### DIFF
--- a/toqito/perms/swap.py
+++ b/toqito/perms/swap.py
@@ -141,7 +141,7 @@ def swap(
         if not all(isinstance(d, (int, float, np.integer, np.floating)) for d in np.ravel(dim)):
             raise TypeError("dim entries must be int or float values.")
         dim = np.array(dim, dtype=int)
-        num_sys = len(dim)
+        num_sys = dim.shape[1] if dim.ndim == 2 and dim.shape[0] == 2 else len(dim)
 
     if len(sys) != 2:
         raise ValueError("InvalidSys: sys must be a vector with exactly two elements.")
--- a/toqito/perms/tests/test_swap.py
+++ b/toqito/perms/tests/test_swap.py
@@ -3,7 +3,7 @@
 import numpy as np
 import pytest
 
-from toqito.perms import swap
+from toqito.perms import permute_systems, swap
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


vprusso (approved):
Verified this fixes #1828: deriving the subsystem count from the number of columns of a two-row rectangular `dim` (rather than the two rows) makes `swap(M, [1, 3], [[2,2,3],[2,2,2]])` return the correct (12, 8) operator, matching `permute_systems`. All 31 test_swap.py cases pass, and the scalar / dim=None / 1-row / square / 2-subsystem forms are unaffected (they still use len(dim)). ruff clean.

Note on CI: the red X is **not** from this PR. The pytest steps all pass; the only failing step is 'Upload coverage information', where the Codecov CLI download failed GPG signature verification (a transient infra issue). I've re-triggered the job.

PR #1852, opened 2026-07-09
fix: guard cone value functions against free CVXPY variables

KSHITIZ6341:
## Summary
- add a shared guard for affine CVXPY expressions whose variables affect the expression value
- reject free CVXPY variables in cone value functions that solve their own SDP
- cover `lieb_ando`, `trace_matrix_power`, `ln_quantum_entropy`, `trace_matrix_log`, and `quantum_relative_entropy`
- preserve algebraically fixed affine expressions such as `Constant(A) + W - W`

Addresses the interim safety behavior discussed in #1846. This does not implement the larger composable-cone redesign.

## Validation
- `/Users/quant/.local/bin/uv run pytest toqito/cones/tests/test_lieb_ando.py toqito/cones/tests/test_trace_matrix_power.py toqito/cones/tests/test_ln_quantum_entropy.py toqito/cones/tests/test_trace_matrix_log.py`
- `/Users/quant/.local/bin/uv run pytest toqito/state_props/tests/test_quantum_relative_entropy.py`
- `/Users/quant/.local/bin/uv run ruff check toqito/cones/_utils.py toqito/cones/lieb_ando.py toqito/cones/trace_matrix_power.py toqito/cones/ln_quantum_entropy.py toqito/cones/trace_matrix_log.py toqito/state_props/quantum_relative_entropy.py toqito/cones/tests/test_lieb_ando.py toqito/cones/tests/test_trace_matrix_power.py toqito/cones/tests/test_ln_quantum_entropy.py toqito/cones/tests/test_trace_matrix_log.py toqito/state_props/tests/test_quantum_relative_entropy.py`
- `/Users/quant/.local/bin/uv run ruff format --check toqito/cones/_utils.py toqito/cones/lieb_ando.py toqito/cones/trace_matrix_power.py toqito/cones/ln_quantum_entropy.py toqito/cones/trace_matrix_log.py toqito/state_props/quantum_relative_entropy.py toqito/cones/tests/test_lieb_ando.py toqito/cones/tests/test_trace_matrix_power.py toqito/cones/tests/test_ln_quantum_entropy.py toqito/cones/tests/test_trace_matrix_log.py toqito/state_props/tests/test_quantum_relative_entropy.py`
- `git diff --check`

Issue #1846, opened 2026-07-09
cones: value functions solve internally and cannot compose into larger SDPs (affine inputs ignore .value)


vprusso:
## Summary

Several functions in `toqito/cones/` solve their own semidefinite program internally and return a scalar. When they are handed an affine CVXPY expression that contains free variables, they optimize over those free variables instead of evaluating at the value supplied, so the result ignores the input and does not compose into a larger SDP. The input's `.value` is used only for the positive-semidefinite prechecks and is then dropped during the solve.

This is a design-level issue that spans the package, so it is filed separately from any single function's PR.

## Two categories of cones functions

**Composable cone constructors (correct, not affected).** These return CVXPY *constraints* for the caller to embed:

- `geometric_mean_hypo_cone`
- `geometric_mean_epi_cone`
- `operator_relative_entropy_epi_cone`
- `geometric_mean`

**Solve-internally, return-a-float (affected).** These build a `cvxpy.Problem`, call `.solve()`, and return a float:

- `lieb_ando`
- `trace_matrix_power`
- `ln_quantum_entropy`
- `trace_matrix_log`
- `integral_relative_entropy`

(The functions added in #1711, `relative_entropy_quadrature`, `tsallis_entropy`, and `tsallis_relative_entropy`, share the same pattern. #1711 handles its own three functions defensively; this issue tracks the broader redesign.)

## Reproductions

`tsallis_relative_entropy` returns a wrong finite value on the affine path:

```python
import numpy as np, cvxpy
from toqito.cones import tsallis_relative_entropy

B = np.diag([0.6, 0.4]); A = np.diag([0.7, 0.3])
print(tsallis_relative_entropy(A, B, 0.5))          # 0.011032   (correct)

X = cvxpy.Variable((2, 2), symmetric=True); X.value = A
print(tsallis_relative_entropy(X, B, 0.5))          # -0.5       (wrong; negative is impossible here)
```

`lieb_ando` becomes unbounded on the affine path (free variables are unconstrained once the function solves in isolation):

```python
import numpy as np, cvxpy
from toqito.cones.lieb_ando import lieb_ando

A = np.diag([0.7, 0.3]); B = np.diag([0.6, 0.4]); K = np.eye(2)
X = cvxpy.Variable((2, 2), symmetric=True); X.value = A
Y = cvxpy.Variable((2, 2), symmetric=True); Y.value = B
[...]

PR #1853, opened 2026-07-09
fix: prevent unsafe CVXPY variable evaluation in cone value functions

Mahizhan-S:
## Summary

This PR implements the interim safety behavior discussed in #1846 for cone value functions that internally solve semidefinite programs.

Previously, several functions could silently optimize over free CVXPY variables instead of evaluating the supplied expression value. Because these functions constructed and solved their own SDPs internally, affine CVXPY expressions containing free variables could produce incorrect values or unbounded optimization problems.

This change detects unsupported free CVXPY variables and raises a clear error instead of silently returning incorrect results.

This PR intentionally does **not** implement the larger composable-cone redesign proposed in #1846, where CVXPY inputs would return epigraph constraints and auxiliary variables instead of solving internally.

---

### 1. Added shared CVXPY variable detection helper

Added `_contains_effective_variables()` in:

- `toqito/cones/_utils.py`

The helper:

- Detects whether CVXPY variables actually influence an expression's value.
- Avoids false positives for algebraically fixed expressions where variables cancel (for example, `Constant(A) + X - X`).
- Temporarily perturbs variable values during detection and restores their original values afterwards.

---

### 2. Added guards to internally-solving cone value functions

Added free-variable validation to:

- `toqito/cones/lieb_ando.py`
- `toqito/cones/trace_matrix_power.py`
- `toqito/cones/ln_quantum_entropy.py`
- `toqito/cones/trace_matrix_log.py`
- `toqito/cones/integral_relative_entropy.py`

These functions construct and solve their own SDPs and return scalar values.

**Before**

- Free CVXPY variables could become optimization variables inside the internally constructed SDP.
- Results could silently differ from the supplied expression value or become unbounded.

**After**

- Unsupported free CVXPY variables raise a clear `ValueError`.
- Numeric NumPy inputs continue to follow the existing evaluation path.
- Constant CVXPY expressions remain supported.

---

### 3. Preserved CVXPY composition behavior

Updated:

- `toqito/state_props/quantum_relative_entropy.py`

The initial guard approach was intentionally **not** applied here because `quantum_relative_entropy` participates in larger CVXPY optimization problems.

Instead, the implementation preserves existing CVXPY composition workflows while ensuring that internally-solving cone value functions reject unsupported free-variable inputs.

---

### 4. Expanded regression test coverage

Added and updated tests covering:

- Rejection of unsupported free CVXPY variables.
- Numeric NumPy inputs.
- Constant CVXPY expressions.
- Direct unit tests for `_contains_effective_variables()`.
- Algebraically cancelled expressions such as `Constant(A) + X - X`.
- Existing `quantum_relative_entropy` CVXPY workflows.
- Updated expected error messages where validation now occurs earlier.

---
## Scope

This PR implements the interim safety behavior proposed in #1846 by rejecting unsupported free CVXPY variables in cone value functions that solve their own SDPs.

It does **not** implement the larger API redesign in which CVXPY inputs return composable cone constraints instead of solving internally.


vprusso (changes requested):
This is the more complete of the two PRs for #1846: it guards all five functions (including `integral_relative_entropy`) and the free-variable detection is sound. I'm consolidating on this one and have closed #1852.

Two changes before merge:

1. **Consolidate onto a single guard.** #1711 lands first and adds a free-variable guard, `_reject_nonconstant_cvxpy`, to `cones/_utils.py` for the three new functions (`tsallis_entropy`, `tsallis_relative_entropy`, `relative_entropy_quadrature`). To avoid two overlapping guards in the same file, please rebase on `master` once #1711 merges and reuse `_reject_nonconstant_cvxpy` for the five existing functions here, dropping the custom `_contains_effective_variables` probe. That leaves one guard across all eight functions. The only input the simpler `is_constant()`-based guard rejects that yours accepts is a contrived effectively-constant expression like `A + X - X`, which no real caller passes, so nothing of value is lost.

2. **Guard placement (minor).** Call the guard after the existing affine / value-None / PSD checks rather than before them, so a free-variable input that is also, say, non-PSD still surfaces the more specific diagnostic.

Thanks for the thorough coverage and the `test_utils` additions.

Mahizhan-S:
## Summary

This PR implements the interim safety behavior discussed in #1846 for cone value functions that internally solve SDPs.

Previously, these functions could silently treat free CVXPY variables as optimization variables inside their internally constructed SDPs instead of evaluating the supplied values. Depending on the function, this could produce incorrect numerical results or unbounded optimization problems.

Rather than silently returning incorrect results, the affected functions now reject unsupported affine/variable CVXPY inputs with a clear error while preserving the existing numeric evaluation path.

This PR intentionally does **not** implement the larger composable-cone redesign proposed in #1846 (returning CVXPY constraints instead of solving internally).

---

## Changes

### 1. Consolidated onto the shared CVXPY guard

Following the review feedback, this PR now reuses the shared helper introduced in `toqito/cones/_utils.py`:

- `_reject_nonconstant_cvxpy`
- `_AFFINE_VARIABLE_NOT_SUPPORTED`

The previous `_contains_effective_variables()` helper has been removed entirely so there is now a single guard implementation shared across all internally-solving cone value functions.

---

### 2. Updated internally-solving cone functions

The following functions now consistently reject unsupported affine/variable CVXPY inputs using the shared helper:

- `toqito/cones/lieb_ando.py`
- `toqito/cones/trace_matrix_power.py`
- `toqito/cones/ln_quantum_entropy.py`
- `toqito/cones/trace_matrix_log.py`
- `toqito/cones/integral_relative_entropy.py`

Behavior is unchanged for NumPy inputs.

For `ln_quantum_entropy` and `trace_matrix_log`, affine expressions with valid numeric `.value` continue through the existing numeric evaluation path rather than constructing an internal SDP.

---

### 3. Guard placement updated

Following the maintainer's review, the shared guard is now evaluated **after** the existing affine, `.value`, and positive-semidefinite validation where appropriate.

This preserves more specific diagnostics when an input is invalid for reasons unrelated to unsupported CVXPY variables.

---

### 4. Test updates

Regression tests were updated to reflect the new shared guard behavior.

Changes include:

- removing tests for the deleted `_contains_effective_variables()` helper.
- updating expected error messages to `_AFFINE_VARIABLE_NOT_SUPPORTED`.
- preserving existing numeric evaluation behavior.
- verifying the updated validation order.

---

## Review feedback addressed

This revision addresses the requested review changes by:

- consolidating onto the shared `_reject_nonconstant_cvxpy` helper;
- removing the custom `_contains_effective_variables()` implementation;
- updating guard placement to follow the existing validation checks;
- updating the associated regression tests.

Mahizhan-S:
### Removal of `test_utils.py`

`test_utils.py` was removed because it only tested the previous internal helper implementation.

After rebasing onto `master` and consolidating onto the shared `_reject_nonconstant_cvxpy` guard, the dedicated helper-specific tests were no longer applicable. The validation behavior is now exercised through the existing public API regression tests for:

- `lieb_ando`
- `trace_matrix_power`
- `trace_matrix_log`
- `ln_quantum_entropy`
- `integral_relative_entropy`
- `quantum_relative_entropy`

This keeps the test suite focused on observable behavior through the public interfaces rather than the implementation details of an internal utility.

vprusso (approved):
Re-reviewed after the rebase. This is done well:

- Consolidated onto the shared `_reject_nonconstant_cvxpy` guard from #1711 (dropped the custom probe), so there's now a single guard across the package.
- Guard placement fixed: it runs after the affine / value-None / PSD checks, so those more specific diagnostics are preserved.
- The coverage is correct and principled: the guard is applied to the three functions that build and solve an SDP over their inputs (`lieb_ando`, `trace_matrix_power`, `integral_relative_entropy`), which is where the silent-wrong-value bug lives. `ln_quantum_entropy` and `trace_matrix_log` recurse on `.value` and evaluate at the point, so they have no such bug and correctly need no guard — I verified both return the correct value for a free-variable input.
- Full cones suite + quantum_relative_entropy tests pass (609 passed), ruff clean.

Minor, non-blocking: the interim contract is now slightly asymmetric (three functions reject affine inputs, two evaluate them at `.value`). Both are safe, and the option-1 redesign in #1846 will unify them, so no change needed here. Thanks for the thorough job.

PR #1863, opened 2026-07-13
Two cone split examples

simsaidan:
## Changes

  -  [x] Split trace matrix log
  -  [x] Split ln quantum entropy
  -  [x] Fix callers in quantum relative entropy

## Checklist

### DIFF
--- a/toqito/cones/__init__.py
+++ b/toqito/cones/__init__.py
@@ -1,10 +1,12 @@
 """CVXPY cone constraints for operator-monotone matrix functions, ported from CVXQUAD (Fawzi--Saunderson) [@cvxquadlink].
 
 These builders return lists of CVXPY constraints describing the epigraph or hypograph of
-a matrix function (matrix geometric mean, operator relative entropy) for use inside a
-semidefinite program.
+a matrix function (matrix geometric mean, operator relative entropy, quantum entropy,
+matrix-logarithm trace) for use inside a semidefinite program.
 """
 
 from toqito.cones.geometric_mean_epi_cone import geometric_mean_epi_cone
 from toqito.cones.geometric_mean_hypo_cone import geometric_mean_hypo_cone
+from toqito.cones.ln_quantum_entropy_hypo_cone import ln_quantum_entropy_hypo_cone
 from toqito.cones.operator_relative_entropy_epi_cone import operator_relative_entropy_epi_cone
+from toqito.cones.trace_matrix_log_hypo_cone import trace_matrix_log_hypo_cone
--- a/toqito/cones/ln_quantum_entropy_hypo_cone.py
+++ b/toqito/cones/ln_quantum_entropy_hypo_cone.py
@@ -1,0 +1,106 @@
+"""CVXPY constraints for the von Neumann entropy hypograph cone."""
+
+# Adapted from CVXQUAD (https://github.com/hfawzi/cvxquad), BSD-2-Clause.
+# Original implementation by Fawzi, Saunderson, et al.
+
+import cvxpy
+import numpy as np
+
+from toqito.cones._utils import _require_square_2d, _symmetric_like_variable
+from toqito.cones.operator_relative_entropy_epi_cone import (
+    operator_relative_entropy_epi_cone,
+)
+
+
+def ln_quantum_entropy_hypo_cone(
+    mat_x: cvxpy.Expression,
+    t: cvxpy.Expression,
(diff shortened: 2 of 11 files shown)


vprusso (changes requested):
This is a strong first example. The cone builders are correct — the tests confirm `Maximize(t)` at `Constant(A)` matches the numeric `tr(C log A)` / entropy, and the free-variable composition tests genuinely recover e.g. `log(3) = 1.0986` with `X → I/3`, so it's real SDP composition rather than point evaluation. The numeric forms are preserved, and the `*_hypo_cone` naming with the `(mat_x, t, [mat_c], *, m, k, apx, hermitian) -> list[Constraint]` signature matches the existing `geometric_mean_hypo_cone` / `operator_relative_entropy_epi_cone` convention. Nicely done, and thanks for the CVXQUAD attribution.

Three things before merge — the second matters most, since this file is the template we'll replicate for the remaining functions:

**1. CI (blocker, trivial).** `ruff format --check` fails on `test_trace_matrix_log.py` and `test_ln_quantum_entropy.py`: a `_NOT_SUPPORTED = re.escape(...)` is hand-wrapped across three lines but fits on one under the project line length. Just run `ruff format` on those two files.

**2. Import back-edge (worth fixing now, before it propagates ~6x).** `trace_matrix_log_hypo_cone.py` imports `is_positive_semidefinite` from `matrix_props`. But post-dispersal `matrix_props` already imports from `cones`, so this makes `cones <-> matrix_props` a two-way package dependency. It doesn't deadlock today (only because `is_positive_semidefinite` and `cones._utils` are leaf modules), but the existing cone builders deliberately import *nothing* from `matrix_props`/`state_props` — keeping those edges one-way was the point of the dispersal. Since this is the template, replicating the PSD-validation import across the remaining builders would entrench the two-way edge everywhere. Could you validate `mat_c` without importing `matrix_props` — e.g. a small inline `eigvalsh(mat_c).min() >= -tol` — or drop the check and let the cvxpy PSD constraint / caller handle it?

**3. Caller consumption (please clarify the intent).** The PR title says "fix callers in quantum_relative_entropy," but `quantum_relative_entropy` now calls the *numeric* `trace_matrix_log` / `ln_quantum_entropy` in its constant branches and still uses `operator_relative_entropy_epi_cone` directly in its SDP branch — neither new `*_hypo_cone` builder is imported by any non-test module. So the composable value is currently exercised only by the new tests, not wired into a real caller. That's fine if the goal is to *expose* these as public composable building blocks — but if the intent was to have `quantum_relative_entropy` compose from `ln_quantum_entropy_hypo_cone`, that part didn't land. Worth stating which, so the pattern is unambiguous for the rollout.

Minor: `test_trace_matrix_log.py:73` casts a complex Hermitian `mat_c` to `float64` for the PSD check (emits a `ComplexWarning`, discards the imaginary part); the cone-test twin correctly uses `complex128` — worth matching.

Overall the cone-split mechanics are right and the template is clean. Mainly want to settle the back-edge before it propagates to the rest.

simsaidan:
@vprusso Added a commit which addresses the CI blocker, the back-edge, and the minor testing type comment. The point about the caller consumption is definitely true, but if you are fine with it, I'd like to leave out of scope for this PR (unless you think I should finish the other cones on this PR). 

To put it here though, I think I would split quantum_relative_entropy into two files like these. if either arg is affine, route to the cone file, otherwise, the numeric file. I haven't thought through the space-optimized version yet since the inputs to that right now don't support cvxpy variables.

vprusso:
Fixes look good: back-edge is gone (inline eigvalsh check), CI is green, and the type nit is handled. Tests pass locally.

Deferring the caller wiring is fine by me, keep this PR at its current scope rather than pulling in the rest of the cones here. This file is the template and I'd rather replicate it in small follow-ups. One ask before I merge: open a tracking issue for the `quantum_relative_entropy` split so it doesn't get lost. Your plan (affine arg to the cone file, otherwise numeric, mirroring `trace_matrix_power`/`lieb_ando`) is the right shape.

One thing to call out: this drops `m, k, apx` from `ln_quantum_entropy` and `trace_matrix_log`. I'm fine with that, the exact `logm` path is cleaner and the quadrature knobs belong on the cone builders, but I'll note it as a breaking change in the next release.

simsaidan:
@vprusso Added a issue so we can merge this ( I can make the issue more robust later today)

Issue #1867, opened 2026-07-16
Add a composable cone for quantum_conditional_entropy (SDP composition)


vprusso:
### Background

PR #1866 split `lieb_ando`, `trace_matrix_power`, and `quantum_relative_entropy` into a numeric path and a composable cone path (following #1863 / #1865). As part of that, `quantum_conditional_entropy` was changed to accept numeric (and constant CVXPY) inputs only; affine or variable CVXPY inputs are now rejected. The `test_sdp_maximize_over_density_matrices` test that optimized conditional entropy over a free `cvxpy` variable was removed and replaced by a "free variable is rejected" test.

The result is that `quantum_conditional_entropy` can no longer be composed directly into a parent SDP. This was intentional and deferred to a follow-up (this issue).

### Task

Add a composable cone for the conditional entropy so it can be used inside a semidefinite program again, and route affine/variable inputs to it (numeric inputs stay on the numeric path).

- Conditional entropy is defined via the quantum relative entropy: `H(A|B) = -D(rho_AB || I_A ⊗ rho_B)` and `H(B|A) = -D(rho_AB || rho_A ⊗ I_B)`. The natural implementation composes `quantum_relative_entropy_epi_cone` (added in #1866) with the appropriate Kronecker lift on the traced subsystem, as the current `quantum_conditional_entropy` docstring already notes.
- Follow the numeric/cone split pattern from #1863 and #1866: numeric/constant inputs use the numeric path; affine or variable inputs return cone constraints rather than being rejected.
- Restore SDP-composition test coverage (a free-`cvxpy`-variable optimization recovering a known conditional-entropy value), analogous to the `*_epi_cone_composition` tests added in #1866.

### Notes

Depends on `quantum_relative_entropy_epi_cone` from #1866 (merged). Keep `cones` a leaf: the new builder must not import from `matrix_props`/`state_props`.

vprusso:
Reframing the scope. The pattern the project settled on (in #1866, and now #1868) is that the numeric value functions reject affine/variable input and composition happens through a separate cone builder, not through in-function routing. The original ask here for in-function routing plus a function-level composition test was over-specified against that pattern.

#1868 delivers the right thing: quantum_conditional_entropy_hypo_cone, composing quantum_relative_entropy_epi_cone with the Kronecker lift, plus a genuine free-variable composition test that recovers H(A|B). So the deliverable for this issue is the builder, which #1868 provides. This will close when #1868 merges.

PR #1866, opened 2026-07-15
Second cone splitting PR covering lieb, trace matrix power, and qre

simsaidan:
space optimized relative entropy is punted to a future pr, as is quantum conditional entropy

## Changes

  -  [ ] Change 1

## Checklist
Before marking your PR ready for review, make sure you checked the following locally. If this is your first PR, you might be notified of some workflow failures after a maintainer has approved the workflow jobs to be run on your PR. 

Additional information is available in the [documentation](https://vprusso.github.io/toqito/contributing-guide/).

  -  [x] Use `ruff` for errors related to code style and formatting.
  -  [x] Verify all previous and newly added unit tests pass in `pytest`.
  -  [x] Check the documentation build does not lead to any failures.

PR #1868, opened 2026-07-17
Cone splits for tsallis, qce, integral re, req

simsaidan:
## Changes

  -  [ ] Change 1

## Checklist

### DIFF
--- a/toqito/channel_metrics/tests/test_channel_relative_entropy.py
+++ b/toqito/channel_metrics/tests/test_channel_relative_entropy.py
@@ -38,7 +38,7 @@ def test_sandwich_parameters_raises_on_eigh_failure(monkeypatch):
         raise LinAlgError("singular pencil")
 
     monkeypatch.setattr(
-        "toqito.state_props.integral_relative_entropy._generalized_eigenvalues",
+        "toqito.cones._integral_relative_entropy_helpers._generalized_eigenvalues",
         failing_eigh,
     )
 
--- a/toqito/cones/__init__.py
+++ b/toqito/cones/__init__.py
@@ -7,11 +7,27 @@ matrix-logarithm trace) for use inside a semidefinite program.
 
 from toqito.cones.geometric_mean_epi_cone import geometric_mean_epi_cone
 from toqito.cones.geometric_mean_hypo_cone import geometric_mean_hypo_cone
+from toqito.cones.integral_relative_entropy_lower_cone import (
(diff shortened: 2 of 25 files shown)


simsaidan:
@vprusso This PR is along the same lines as the other two, but for tsallis entropies, quantum conditional entropy, rel_entr_quad, and integral relative entropy. Looking forward to your review!

vprusso:
Reviewed this closely, including running the builders numerically. Strong work, especially the integral split: the extracted lower/upper cone constraints reproduce the old internal solve bit for bit, and all seven builders check out for epi/hypo direction, sign, and the Kronecker lift. No coverage lost, the tests are genuine free-variable composition tests, and the critical rule holds: no cone module imports matrix_props or state_props.

One change I want before merge:

`quantum_conditional_entropy_hypo_cone.py` imports `toqito.matrix_ops.partial_trace`. Since `matrix_ops` already imports `cones` (`geometric_mean` -> `cones._utils`), that closes a two-way `cones <-> matrix_ops` loop. It is not fatal today, but it is the same leaf violation we removed in #1863, just pointed at `matrix_ops` instead of `matrix_props`. Please keep `cones` a strict leaf: inline a small partial trace in the builder (or add a leaf helper in `cones._utils`) rather than importing from `matrix_ops`.

Two smaller items:

- `quantum_conditional_entropy_hypo_cone` validates `dim` with `isinstance(dim[0], int)`, which rejects a numpy int array even though the signature advertises `np.ndarray` (lists work, so no test catches it). Accept `np.integer` / `numbers.Integral`.
- One test builds a cvxpy `Constant` from a nested Python list, which triggers a "nested list Constant is undefined behavior" UserWarning. Wrap it in `np.array`.

Two notes, not blockers:

- `relative_entropy_quadrature` drops its `m` and `k` parameters (moved to the cone builder). Fine and consistent with the pattern, but it is a public signature break, so I will note it in the release.
- On routing: keeping the numeric functions rejecting affine input and exposing composition only through the separate builders is the right call and matches how `quantum_relative_entropy` went in #1866. That means #1867's ask for in-function routing was over-specified on my end; I will relax it to "provide a composable builder," which this delivers. So do not add in-function routing.

Fix the partial_trace edge and the two minor items and this is good to merge.

simsaidan:
@vprusso I added a new commit addressing those three feedback points. Let me know if I should make any other changes.

PR #1890, opened 2026-07-25
perf: vectorize standard basis construction

### LINKED ISSUE #1887: perf: vectorize a loop-based matrix constructor
A few small matrix constructors build their result with a Python `for` loop over a preallocated `np.zeros`/`np.empty`, where numpy vectorization would be faster and clearer.

### What to do
Pick ONE of the following and replace its loop with a vectorized construction:
- `toqito/matrices/standard_basis.py`
- `toqito/matrices/gen_pauli_z.py`
- `toqito/matrices/entangled_subspace.py`
- `toqito/perms/perfect_matchings.py`

### Requirements
- The output must be bit-for-bit identical to the current implementation. Keep the existing test, and add a case comparing the new output to the previous one if helpful.
- Include a measured speedup on a representative size (a short `timeit` snippet in the PR description is enough).
- Do not change the function's signature or behavior.

This one needs some numpy comfort, so it is a good pick if you want a slightly meatier first PR.

### Verifying
```
uv run pytest toqito/matrices/tests toqito/perms/tests
```

09Catho:
﻿## Summary

- replace the Python-level standard-basis construction loop with `numpy.eye`
- preserve flattened and column-vector output shapes
- broaden exact-output coverage across dimensions 1, 2, and 5

## Performance

Measured locally with Python 3.13 and NumPy through the project `uv` environment:

| Dimension | Previous | Vectorized | Speedup |
| --- | ---: | ---: | ---: |
| 32 | 1313.3 us/call | 16.8 us/call | 78.0x |
| 128 | 18848.0 us/call | 58.8 us/call | 320.4x |

The benchmark also asserted `np.array_equal` between the previous and new outputs.

## Validation

- `uv run pytest toqito/matrices/tests -q` (143 passed)
- `uv run ruff check toqito/matrices/standard_basis.py toqito/matrices/tests/test_standard_basis.py`
- `uv run ruff format --check toqito/matrices/standard_basis.py toqito/matrices/tests/test_standard_basis.py`
- `git diff --check`

Closes #1887

### DIFF
--- a/toqito/matrices/standard_basis.py
+++ b/toqito/matrices/standard_basis.py
@@ -35,9 +35,7 @@ def standard_basis(dim: int, flatten: bool = False) -> list[np.ndarray]:
         ```
 
     """
-    first_basis_vector = np.zeros(dim) if flatten else np.zeros((dim, 1))
-    first_basis_vector[0] = 1.0
-
-    # The standard_basis is obtained by cyclic permutations of the first basis
-    # vector
-    return [np.array([first_basis_vector[i - j] for i in range(dim)]) for j in range(dim)]
+    basis = np.eye(dim)
+    if not flatten:
+        basis = basis[..., np.newaxis]
+    return list(basis)
--- a/toqito/matrices/tests/test_standard_basis.py
+++ b/toqito/matrices/tests/test_standard_basis.py
@@ -1,23 +1,24 @@
 """Test standard_basis."""
 
 import numpy as np
+import pytest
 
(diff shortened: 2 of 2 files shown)


vprusso:
Thanks, and welcome to toqito! Confirmed the np.eye version matches the previous output exactly in both flatten modes across dimensions, and the parametrized exact-equality tests are a nice improvement over the originals. Big speedup too. Merging.

Issue #1896, opened 2026-07-25
test: broaden narrow single-case tests to parametrized tests


vprusso:
Some tests exercise a function at only a single hardcoded dimension/case. Broadening them into `@pytest.mark.parametrize` tests over several dimensions catches more edge cases with little extra code. #1890 did exactly this for `standard_basis` (going from two dim-2 tests to parametrized tests over dims 1, 2, 5 with exact-equality checks), which is a good model to follow.

### What to do
Find tests that only cover one hardcoded case and parametrize them over a few representative sizes, tightening `assert np.allclose(...)` to exact `np.array_equal(...)` where the output is exact. A starting candidate:
- `toqito/states/tests/test_horodecki.py` (`test_horodecki_state_3_3_default`)

Look for others with:
```
git grep -nE "^def test_.*_(2|3|4)_(d|dim|qubit)" toqito
```
and scan test files for single-value assertions that could cover more dimensions.

### Requirements
- Keep the tests meaningful (don't assert the implementation against itself in a way that can't fail). Compare against an independent expected value where possible.

### Verifying
```
uv run pytest toqito/<the test file you changed>
```

Jethin10:
Hi! I'd like to work on this. I'll start with toqito/states/tests/test_horodecki.py, add representative parametrized cases with independent expected values, and run the targeted pytest and Ruff checks. I'll keep the scope focused and reference PR #1890's testing approach.

PR #1897, opened 2026-07-25
test: parametrize Horodecki state dimensions

Jethin10:
## Summary
- consolidate the default, 3-by-3, and 2-by-4 Horodecki state checks into one parametrized test
- use full-precision expected matrices instead of rounded fixtures
- tighten the 2-by-4 tolerance from 0.2 to 1e-12 so incorrect matrix entries cannot pass unnoticed

## Validation
- `uv run pytest toqito/states/tests/test_horodecki.py -q` (6 passed)
- `uv run ruff check toqito/states/tests/test_horodecki.py`
- `uv run ruff format --check toqito/states/tests/test_horodecki.py`
- `uv run ty check toqito/states/tests/test_horodecki.py`
- full pytest was attempted locally but exceeded 15 minutes without surfacing a failure; CI will run the authoritative full suite
- repository-wide `ty check toqito` currently reports existing diagnostics outside this changed test file

Addresses #1896.


vprusso:
Thanks, and welcome! This is a nice improvement, folding the three cases into one parametrized test and tightening the assertions (the old 2x4 case used atol=0.2, which barely tested anything; exact values at atol=1e-12 is much better). Tests pass. Merging.

PR #1898, opened 2026-07-26
perf(states): vectorize ghz and w_state constructors

### LINKED ISSUE #1895: perf: vectorize the ghz and w_state constructors
Split out from #1887. Two state constructors set a handful of entries in a preallocated zero vector with a small Python loop; both vectorize cleanly.

- `toqito/states/ghz.py`: the loop sets `ghz_state[index] = coeff[i]` where `index = i * step` for a constant `step`. Vectorizes to something like `ghz_state[np.arange(dim) * step, 0] = coeff`.
- `toqito/states/w_state.py`: the loop sets `ret_w_state[2**i] = coeff_arr[num_qubits - i - 1]`. Vectorizes to `ret_w_state[2**np.arange(num_qubits), 0] = coeff_arr[::-1]`.

Both are small, and I've confirmed the vectorized form gives identical output.

### What to do
Replace the loops in `ghz` and `w_state` with the vectorized index assignments (one PR can do both, or one each).

### Requirements
- Output bit-for-bit identical (keep the existing tests).
- Do not change signatures or behavior.

### Verifying
```
uv run pytest toqito/states/tests/test_ghz.py toqito/states/tests/test_w_state.py
```

smoran02:
## Description

Closes #1895. Replaces the small per-coefficient Python loops in the `ghz` and `w_state` constructors with single vectorized index assignments, as proposed in the issue. Output is bit-for-bit identical to the loop versions and no signatures change.

## Changes

  -  [x] `toqito/states/ghz.py`: the fill loop becomes `ghz_state[np.arange(dim) * step, 0] = coeff` with `step = sum(dim**k for k in range(num_qubits))`.
  -  [x] `toqito/states/w_state.py`: the fill loop becomes `ret_w_state[2 ** np.arange(num_qubits), 0] = coeff_arr[::-1]`.
  -  [x] Both functions flatten the coefficient array after coercion (`np.array(coeff).ravel()`) so column-shaped input keeps working exactly as it did with the old element-wise assignment.
  -  [x] Tests: added `dim=1` and `num_qubits=1` boundary cases and a column-coefficient case for `ghz`, a column-coefficient test for `w_state`, and exact-equality tests for both (using coefficients like `[3, 4]` that normalize to exactly `[0.6, 0.8]` in float64, plus unit-norm integer coefficients that pin dtype preservation).

Equivalence was verified against the old loop implementations with `np.array_equal` over `dim` 1-7 x `num_qubits` 1-5 and `w_state` sizes 2-13, with default, integer, complex, unnormalized, and column-shaped coefficients: zero mismatches in value, shape, or dtype, and all error paths unchanged.

Performance: the fill step is a wash at tiny sizes (microsecond scale) and wins grow with size: about 5.8x at `dim=16`, 29.7x at `dim=100`, 142x at `dim=1000` for `ghz`, and 2-4x for `w_state` at 10-20 qubits.

One note for reviewers: flattening makes the coefficient length check size-based rather than shape-based, so a malformed matrix input such as `ghz(4, 2, np.eye(2))` is now accepted (flattened to length 4) where it previously raised. Row-vector input `(1, dim)`, which is what the docstrings literally describe, now works as documented. Happy to add a stricter shape guard if preferred.

## Checklist

  -  [x] Use `ruff` for errors related to code style and formatting. (`uv run ruff check`: all checks passed; `uv run ruff format --check`: 502 files already formatted)
  -  [x] Verify all previous and newly added unit tests pass in `pytest`. (full suite: 3660 passed, 63 skipped, 3 xfailed in 15:55; targeted: 45 passed)
  -  [x] Check the documentation build does not lead to any failures. (`mkdocs build` from `docs/`: built in 325.63 seconds with no errors)

### DIFF
--- a/toqito/states/ghz.py
+++ b/toqito/states/ghz.py
@@ -66,7 +66,8 @@ def ghz(dim: int, num_qubits: int, coeff: list[int] | None = None) -> np.ndarray
     if coeff is None:
         coeff = np.ones(dim)
     else:
-        coeff = np.array(coeff)
+        # Flatten so column-vector input keeps working with the vectorized fill below.
+        coeff = np.array(coeff).ravel()
     if len(coeff) != dim:
         raise ValueError("InvalidCoeff: The variable `coeff` must be a vector of length equal to `dim`.")
 
@@ -79,10 +80,9 @@ def ghz(dim: int, num_qubits: int, coeff: list[int] | None = None) -> np.ndarray
 
     # Initialize the GHZ state vector, matching the coefficient dtype so complex coefficients are preserved.
     ghz_state = np.zeros((dim**num_qubits, 1), dtype=coeff.dtype)
-    # Fill the state vector with the corresponding coefficients.
-    for i in range(dim):
-        # Calculate the index for the tensor product state |i, i, ..., i>.
-        index = sum(i * (dim**k) for k in range(num_qubits))
-        ghz_state[index] = coeff[i]
+    # Fill the state vector with the corresponding coefficients. The tensor product state
+    # |i, i, ..., i> lives at index i * (1 + dim + dim^2 + ... + dim^(num_qubits - 1)).
+    step = sum(dim**k for k in range(num_qubits))
+    ghz_state[np.arange(dim) * step, 0] = coeff
 
     return ghz_state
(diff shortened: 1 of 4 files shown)


vprusso:
Thanks, and welcome! Confirmed the vectorized `ghz` and `w_state` outputs are identical (shape, dtype, values) to the loops across dimensions, and CI is green. Merging.

Issue #1886, opened 2026-07-25
docs: add runnable Examples sections to public functions that lack them


vprusso:
The documentation build executes the `Examples` code blocks in docstrings (the ` ```python exec="1" ` blocks), so a runnable example doubles as documentation and a lightweight smoke test. Several public functions have no `Examples` section.

### What to do
Add an `Examples` section with a short, runnable snippet, matching the existing style:
```` 
```python exec="1" source="above" result="text"
import numpy as np
from toqito.states import dicke

print(dicke(...))
```
````

Good targets (pick one or a few per PR):
- `toqito/measurement_ops/measure.py`
- `toqito/states/dicke.py`
- `toqito/state_props/has_symmetric_inner_extension.py`
- `toqito/state_props/integral_relative_entropy.py`
- the cone builders in `toqito/cones/` (e.g. `geometric_mean_epi_cone.py`)

### Verifying
The example runs during the docs build:
```
uv sync --group docs
cd docs && uv run mkdocs build
```
Confirm the snippet runs without error and the printed output is sensible. Looking at an existing function's `Examples` section (e.g. `toqito/channels/dephasing.py`) is the easiest way to see the expected format.

Jethin10:
I’ll add a focused runnable Examples section to toqito/state_props/has_symmetric_inner_extension.py, using a small two-qubit maximally mixed state so the docs build exercises the public API without making the example expensive. I’ll validate the snippet directly and with the MkDocs build.

PR #1903, opened 2026-07-27
docs: add runnable integral relative entropy example

Artamta:
## Description

  Adds a runnable example for `evaluate_relative_entropy_integral` as part of #1886.

  ## Changes

  - [x] Demonstrate the function using two diagonal density matrices.
  - [x] Print a concise, reproducible relative-entropy estimate.

  ## Checklist

  - [x] Check the documentation build does not lead to any failures.

  ## Testing performed

  - New example runs successfully and prints `0.1194`.
  - 19 targeted tests passed.
  - `ruff check` passed.
  - `ruff format --check` passed.
  - `git diff --check` passed.
  - Full docs build progressed through generation but was stopped during an unrelated, long-running existing CVXPY example.

### DIFF
--- a/toqito/state_props/integral_relative_entropy.py
+++ b/toqito/state_props/integral_relative_entropy.py
@@ -75,6 +75,21 @@ def evaluate_relative_entropy_integral(
             degenerate.
         RuntimeError: If a bound SDP fails to solve.
 
+    Examples:
+        Estimate the relative entropy between two diagonal density matrices.
+
+        ```python exec="1" source="above" result="text"
+        import numpy as np
+
+        from toqito.state_props import evaluate_relative_entropy_integral
+
+        mat_x = np.diag([0.75, 0.25])
+        mat_y = np.diag([0.5, 0.5])
+
+        estimate = evaluate_relative_entropy_integral(mat_x, mat_y)
+        print(f"{estimate:.4f}")
+        ```
+
     """
     if isinstance(mat_x, cvxpy.Expression) and mat_x.value is not None:
         mat_x_eval = np.asarray(mat_x.value)


vprusso:
Thanks @Artamta. The example is well-formed and runs, but reviewing it surfaced an accuracy problem in the function itself that I don't want to put front-and-center in the rendered docs.

For `X = diag(0.75, 0.25)`, `Y = diag(0.5, 0.5)`:

- The exact relative entropy is `0.13081` nats.
- Evaluating the function's *own* stated integral formula, $D(X\|Y) = \int_\mu^\lambda \frac{ds}{s}\,\mathrm{tr}^+[Ys - X] + \log\lambda + 1 - \lambda$, with fine quadrature also gives `0.13081`.
- But `evaluate_relative_entropy_integral` returns `0.11940`, and the lower and upper bounds both collapse to `0.11940` (zero gap, so they don't bracket the true value). Refining the grid (`epsilon_dec` from 1e-1 down to 1e-4) doesn't move it.

So this isn't a coarse-approximation artifact; the estimate converges to the wrong number on a trivial diagonal case. A docstring example is the first thing a user runs, so I'd rather not ship one whose printed value is visibly ~9% off from a result the reader can compute by hand.

Holding this until the underlying accuracy issue is resolved (that's a separate fix, not something to paper over here). Once the function returns a faithful value, this example is a good addition. Nothing wrong with your PR mechanically.

simsaidan:
@vprusso I came across this PR, and I took a look into the integral relative entropy implementation. I think the issue is in the mu/lambda generator. It was setting the lower bound mu to 1/lambda, which for small lambda was overestimating the grid lower bound, leading to too small values. Opened a fix in #1904, let me know what you think

vprusso:
The underlying accuracy bug is now fixed on master (#1904, thanks to @simsaidan): `_sandwich_parameters` was truncating the integration range at `1/lam`. On current master your example prints `0.1308`, matching the exact relative entropy, so it's a faithful demonstration now. Thanks for the patience, @Artamta. Merging.

PR #1904, opened 2026-07-30
Fix sandwich bug for integral relative entropy

simsaidan:
the previous sandwich function for finding mu was buggy and set mu to 1/lam. fix the function and all a fallback for negative mu. add tests to catch original bug.

## Changes

  -  [ ] Change 1

## Checklist

### DIFF
--- a/toqito/channel_metrics/channel_relative_entropy.py
+++ b/toqito/channel_metrics/channel_relative_entropy.py
@@ -126,7 +126,7 @@ def channel_relative_entropy(
     choi_2 = channel_2
     solve_kwargs = {"eps": 1e-8, "verbose": False, **kwargs}
 
-    mu, lam = _sandwich_parameters(choi_1, choi_2)
+    mu, lam = _sandwich_parameters(choi_1, choi_2, epsilon_dec)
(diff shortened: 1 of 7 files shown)


vprusso:
Thanks @simsaidan, this is a clean root-cause. Confirmed: `_sandwich_parameters` was returning `mu = min` of the generalized eigenvalues of `(sigma, rho)`, which equals `1/lam`, so the grid started at `1/lam` instead of the true lower endpoint `min` of the eigenvalues of `(rho, sigma)`, truncating the integral and biasing the estimate down.

I verified the fix across diagonal, multi-level, near-singular, and random non-diagonal Hermitian pairs: the qubit cases now match the closed form to ~1e-11, and the general cases converge to the Umegaki relative entropy within `epsilon_dec` (worst 1.3e-4 at `epsilon_dec=1e-3`). The `mu <= 0` fallback to `epsilon_dec` for singular `X` matches the remark following Corollary 1 of the cited paper, and the closed-form regression tests would have caught the original bug. Threading `epsilon_dec` through every call site is the right call. Merging.

PR #1906, opened 2026-07-30
docs: add runnable dicke and measure examples

happykawayigt:
## Description

Adds proper runnable `Examples` sections to `dicke` and `measure` as part of #1886.

## Changes

- [x] Promote the existing Dicke state snippets into a rendered `Examples` section.
- [x] Add a rendered `Examples` section to `measure` and correct the state/operator argument order in its projector example.

## Validation

- [x] Ran every documented snippet directly against this branch and verified the printed results
- [x] Targeted tests (22 passed)
- [x] `ruff check`
- [x] `ruff format --check`
- [x] `git diff --check`
- [x] GitHub Actions test and style checks passed

A full local MkDocs build was not run because `mkdocs` is unavailable in the local environment. The repository's Docs Preview check was skipped by its workflow configuration.

### DIFF
--- a/toqito/measurement_ops/measure.py
+++ b/toqito/measurement_ops/measure.py
@@ -42,17 +42,20 @@ def measure(
 
     and, if \(p_i > tol\), the post‐measurement state is updated via
 
-    \[
-        u = \frac{1}{\sqrt{3}} e_0 + \sqrt{\frac{2}{3}} e_1
-    \]
+    Examples:
+        Consider the state
 
-    where we define \(u u^* = \rho \in \text{D}(\mathcal{X})\).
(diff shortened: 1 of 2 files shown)


vprusso:
Thanks @happykawayigt. Verified against the branch: the corrected `measure(rho, proj_0)` prints 1/3 and `measure(rho, proj_1)` prints 2/3, matching the documented probabilities. The argument-order swap is a real fix, the signature is `measure(state, measurement)`, so state comes first (the value coincides here only because of the symmetry of this particular example). Moving the example-specific math into the `Examples:` block is the right structure. Merging.

Issue #1888, opened 2026-07-25
docs: add missing Raises: sections to function docstrings


vprusso:
Many functions raise `ValueError`/`TypeError` on bad input but do not document it in a `Raises:` docstring section. Adding these sections makes the API self-documenting and surfaces the conditions in the rendered docs.

### What to do
For a function that raises but has no `Raises:` section, add one listing each exception and the condition that triggers it, in the Google style used elsewhere in the codebase:
```
    Raises:
        ValueError: If `dim` does not evenly divide the input.
```

Good targets (pick a few per PR):
- `toqito/channel_ops/kraus_to_choi.py`
- `toqito/channel_props/channel_dim.py`
- `toqito/channels/amplitude_damping.py`, `bitflip.py`, `phase_damping.py`, `ldot_channel.py`
- `toqito/matrices/cyclic_permutation.py`

List each `raise` and its condition with:
```
git grep -nE "raise (Value|Type)Error" toqito/<file>
```

### Verifying
```
uv sync --group docs
cd docs && uv run mkdocs build
```
Confirm the new `Raises:` section renders on the function's docs page.

dikshakbobade:
Hi, I'd like to work on this — I'll start with a few of the listed files and submit a PR. Thanks!

delaidam:
@dikshakbobade are you still working on this? If you haven't gotten to
kraus_to_choi.py or channel_dim.py yet (the two files from the list
not covered by #1905), I'd be happy to take those — just let me know
either way so we don't duplicate effort.

vprusso:
@delaidam thanks for checking before starting, and sorry for the slow reply - that courtesy is genuinely appreciated and it is the right instinct.

Answering the specific question: `kraus_to_choi.py` and `channel_dim.py` were both picked up in #1908, so those two are done. I have since swept the remaining public functions in #1911, which added `Raises:` sections to 24 more across `channel_ops`, `channel_props`, `matrices`, `matrix_ops`, `matrix_props`, `measurement_props`, `perms`, `rand`, `state_opt`, `state_props`, and `states`.

That closes this one out. An AST sweep now finds no public function that raises `ValueError`/`TypeError` without documenting it, so there is nothing left here to claim. Closing.

[...]

PR #1908, opened 2026-08-01
docs: document channel input errors

happykawayigt:
## Description

Documents explicit input-validation exceptions in two channel APIs as part of #1888.

## Changes

- [x] Document the invalid `sys` and empty Kraus-list cases in `kraus_to_choi`.
- [x] Document rectangular-space, dimension-mismatch, inconsistent Kraus-size, and invalid `dim` cases in `channel_dim`.

## Checklist

- [ ] Check the documentation build does not lead to any failures.

## Validation

- Targeted tests: 28 passed
- `ruff check`
- `ruff format --check`
- `git diff --check`

A full local MkDocs build was not run because the documentation dependency group is not installed in this worktree.

## AI assistance

Implemented and reviewed with OpenAI Codex; all listed validation commands were run locally.

### DIFF
--- a/toqito/channel_ops/kraus_to_choi.py
+++ b/toqito/channel_ops/kraus_to_choi.py
@@ -28,6 +28,9 @@ def kraus_to_choi(kraus_ops: list[np.ndarray] | list[list[np.ndarray]], sys: int
     Returns:
         The corresponding Choi matrix of the provided Kraus operators.
 
+    Raises:
+        ValueError: If ``sys`` is negative or ``kraus_ops`` is empty.
+
     Examples:
         The transpose map:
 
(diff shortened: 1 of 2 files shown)


vprusso (approved):
Checked each documented exception against the code. `kraus_to_choi` raises exactly on negative `sys` and an empty operator list. For `channel_dim`, the four described cases cover all six raise sites (both allow_rect branches, the dim-mismatch, the inconsistent Kraus sizes, and the `_expand_dim` shape check). Both docstrings are already raw strings, so the added \(2 \times 2\) is safe.

Minor nit, not blocking: this file's existing prose uses single backticks (`phi`, `dim`, `allow_rect`) while the new text uses double. Both render identically as inline code and the codebase is already mixed, so nothing to fix here.

PR #1920, opened 2026-08-04
perf: vectorize channel Gram matrix construction

### LINKED ISSUE #1913: perf: vectorize the Gram matrix construction in channel_gram_matrix
`channel_gram_matrix` builds its Gram matrix with a nested Python loop, calling `np.trace` once per entry:

```python
n = len(isometries)
gram = np.zeros((n, n), dtype=complex)
for i in range(n):
    for j in range(n):
        gram[i, j] = np.trace(isometries[j].conj().T @ isometries[i] @ sigma)
```

That is `n^2` separate matrix products, and each one materializes a full matrix only to take its trace and throw the rest away. The whole thing can be a single contraction.

### What to do

Stack the isometries into one array and compute all entries at once. One formulation that works:

```python
mats = np.stack(isometries)  # shape (n, rows, cols)
gram = np.einsum("jab,iac,cb->ij", mats.conj(), mats, sigma, optimize=True)
```

Keep the signature, the return dtype, and the `sigma` validation exactly as they are. Only the loop should change.

### Why it is worth doing

Measured against the current implementation, with random complex isometries:

| isometries | local dim | before | after | speedup |
| ---: | ---: | ---: | ---: | ---: |
| 8 | 4 | 4.5 ms | 0.9 ms | 5.2x |
| 20 | 8 | 29.9 ms | 1.1 ms | 26.8x |

Agreement with the current output was `max|difference|` of about `1e-14`, which is ordinary floating-point reordering noise, not a behavior change.

### Verifying

The result must match the current implementation. Before changing anything, capture a reference:

```python
import numpy as np
from toqito.channel_ops import channel_gram_matrix

rng = np.random.default_rng(0)
isometries = [rng.normal(size=(4, 4)) + 1j * rng.normal(size=(4, 4)) for _ in range(8)]
sigma = np.eye(4) / 4
reference = channel_gram_matrix(isometries, sigma)
```

then confirm `np.allclose(reference, new_result)` after your change. Also run:

```
uv run pytest toqito/channel_ops/tests/ -q
uv run ruff check toqito/channel_ops/channel_gram_matrix.py
```

Please include a short before/after timing in the PR description.

BortnikMaxim:
## Description

Vectorize the O(n²) Python loop in channel_gram_matrix as a single optimized NumPy contraction while preserving the public signature, validation behavior, output values, and complex return dtype.

Closes #1913

## Changes

- stack the validated isometries once and compute all Gram entries with np.einsum
- explicitly retain the existing complex output dtype
- add a regression test against the entry-by-entry definition using complex rectangular isometries and a nontrivial density matrix

## Benchmark

Median of 7 repeats with 10 calls per repeat on arm64 macOS, Python 3.12.13, and NumPy 2.4.1. Timings include the existing validation.

| isometries | local dim | before | after | speedup |
| ---: | ---: | ---: | ---: | ---: |
| 8 | 4 | 0.375 ms | 0.174 ms | 2.2x |
| 20 | 8 | 2.919 ms | 0.461 ms | 6.3x |
| 40 | 16 | 11.740 ms | 0.954 ms | 12.3x |

## Checklist

- [x] ruff check and ruff format --check pass for channel_ops
- [x] channel_ops test suite passes: 71 passed
- [ ] documentation build not run locally because this change does not touch docs

### DIFF
--- a/toqito/channel_ops/channel_gram_matrix.py
+++ b/toqito/channel_ops/channel_gram_matrix.py
@@ -79,9 +79,5 @@ def channel_gram_matrix(
     elif sigma.shape != (dim_in, dim_in):
         raise ValueError("sigma must be a (dim_in, dim_in) density operator.")
 
-    n = len(isometries)
-    gram = np.zeros((n, n), dtype=complex)
-    for i in range(n):
-        for j in range(n):
-            gram[i, j] = np.trace(isometries[j].conj().T @ isometries[i] @ sigma)
-    return gram
+    mats = np.stack(isometries)
+    return np.einsum("jab,iac,cb->ij", mats.conj(), mats, sigma, dtype=complex, optimize=True)
--- a/toqito/channel_ops/tests/test_channel_gram_matrix.py
+++ b/toqito/channel_ops/tests/test_channel_gram_matrix.py
@@ -40,6 +40,24 @@ def test_custom_sigma():
     assert np.isclose(gram[0, 1], 0.5)
 
 
+def test_vectorized_result_matches_reference_for_rectangular_isometries():
+    """The vectorized contraction agrees with the entry-by-entry definition."""
+    rng = np.random.default_rng(0)
+    isometries = [np.linalg.qr(rng.normal(size=(5, 3)) + 1j * rng.normal(size=(5, 3)))[0] for _ in range(4)]
+    sigma_factor = rng.normal(size=(3, 3)) + 1j * rng.normal(size=(3, 3))
+    sigma = sigma_factor @ sigma_factor.conj().T
+    sigma /= np.trace(sigma)
+    expected = np.array(
+        [[np.trace(right.conj().T @ left @ sigma) for right in isometries] for left in isometries],
+        dtype=complex,
+    )
+
+    result = channel_gram_matrix(isometries, sigma)
+
+    assert result.dtype == np.dtype(complex)
+    assert np.allclose(result, expected)
+
+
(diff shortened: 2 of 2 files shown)


vprusso (approved):
Thanks @BortnikMaxim. Verified rather than taken on faith, and it holds up.

**Correctness.** Compared against the previous entry-by-entry implementation, max elementwise difference ~1e-16 across square complex, **rectangular** (5x3, 7x2), and real-valued isometries. The einsum index string is right for the rectangular case, which is the one that would most easily have been broken: `jab,iac,cb->ij` contracts to `tr(V_j* V_i sigma)` with `b, c` both ranging over the input dimension. Good call keeping `dtype=complex` explicit, since the old version got that from the pre-allocated array and dropping it would have silently changed the return dtype for real inputs.

**Benchmark reproduces.** Comparing the full function (including validation) on master vs this branch:

| case | your table | my measurement |
| --- | ---: | ---: |
| n=8, d=4 | 2.2x | 2.0x |
| n=20, d=8 | 6.3x | 5.2x |
| n=40, d=16 | 12.3x | 8.5x |

Same shape, minor differences attributable to hardware. Note for anyone re-measuring: you have to compare full function against full function. Timing the bare loop against the full new function makes the vectorized version look *slower* at small n, because input validation dominates there.

The added regression test builds the expected matrix from the definition with rectangular complex isometries and a non-trivial sigma, which is exactly the right test to have.

Approving; will merge once the pytest matrix finishes.

PR #1921, opened 2026-08-04
perf(channels): build depolarizing Kraus operators without a nested loop

### LINKED ISSUE #1914: perf: build the depolarizing Kraus operators without a nested loop
When `depolarizing` is asked for Kraus operators, it builds the `d^2` single-entry operators one at a time, allocating a fresh zero matrix for each:

```python
coeff = np.sqrt((1 - param_p) / dim)
if coeff > 0:
    for i in range(dim):
        for j in range(dim):
            op = np.zeros((dim, dim))
            op[i, j] = coeff
            kraus_ops.append(op)
```

The set of all single-entry matrices is just the identity of size `d^2` reshaped into `d^2` matrices of size `d` by `d`, so the loop can go away.

### What to do

Replace the nested loop with a single construction, for example:

```python
kraus_ops.extend(coeff * np.eye(dim * dim).reshape(dim * dim, dim, dim))
```

Two things to preserve:

- The ordering of the operators must stay the same (row-major over `i` then `j`), since callers may rely on it.
- The existing `param_p > 0` and `coeff > 0` guards must stay, so zero-weight operators are still omitted.

### Why it is worth doing

Measured against the current implementation:

| dim | before | after | speedup |
| ---: | ---: | ---: | ---: |
| 4 | 0.4 ms | 0.3 ms | 1.7x |
| 16 | 13.1 ms | 3.9 ms | 3.3x |

The generated operators were identical (not merely close) to the current ones at both sizes.

### Verifying

Compare against the current output element by element, including order:

```python
import numpy as np
from toqito.channels import depolarizing

before = depolarizing(4, 0.3, return_kraus_ops=True)   # capture on master first
after = depolarizing(4, 0.3, return_kraus_ops=True)    # after your change
assert len(before) == len(after)
assert all(np.allclose(x, y) for x, y in zip(before, after))
```

Then run:

```
uv run pytest toqito/channels/tests/test_depolarizing.py -q
uv run ruff check toqito/channels/depolarizing.py
```

WAHIB-EL-KHADIRI:
Closes #1914.

Builds the depolarizing Kraus operators without the `d^2` nested-loop allocations.

The single-entry operators `sqrt((1-p)/d) |i><j|` are exactly the `d^2` rows of the `(d^2)`-identity reshaped into `d`-by-`d` matrices, so the loop is replaced by a single construction:

```python
kraus_ops.extend(coeff * np.eye(dim * dim).reshape(dim * dim, dim, dim))
```

**Preserved**, as requested in the issue:
- **Ordering** — row-major over `i` then `j` (the `reshape` of the identity yields exactly that order).
- **Guards** — the `param_p > 0` (keeps `sqrt(p)·I`) and `coeff > 0` (drops zero-weight ops) branches are untouched.

**Verification** — for `dim ∈ {1..5}` and `p ∈ {0, 0.3, 0.5, 1.0}`, the new construction returns an identical list (count, ordering, and values) to the old nested loop:

```
IDENTICAL for all dims/p (count, ordering, values). vectorized fix correct.
```

No behavior change, only the internal construction.

---
🤖 Assisted by Claude (disclosed).

### DIFF
--- a/toqito/channels/depolarizing.py
+++ b/toqito/channels/depolarizing.py
@@ -113,11 +113,9 @@ def depolarizing(dim: int, param_p: float = 0, return_kraus_ops: bool = False) -
             kraus_ops.append(np.sqrt(param_p) * np.eye(dim))
         coeff = np.sqrt((1 - param_p) / dim)
         if coeff > 0:
-            for i in range(dim):
-                for j in range(dim):
-                    op = np.zeros((dim, dim))
-                    op[i, j] = coeff
-                    kraus_ops.append(op)
+            # The single-entry operators |i><j| for all i, j are the d^2 rows of the
+            # (d^2)-identity reshaped into d-by-d matrices, in row-major (i then j) order.
+            kraus_ops.extend(coeff * np.eye(dim * dim).reshape(dim * dim, dim, dim))
         return kraus_ops
 
     result = np.zeros((dim**2, dim**2), dtype=np.float64)


vprusso (approved):
Thanks @WAHIB-EL-KHADIRI. Verified by execution rather than inspection: the new construction produces a list identical to the previous nested loop (exact `np.array_equal`, not just `allclose`) in count, ordering, and values for every `dim` in 1..6 crossed with `p` in {0, 0.3, 0.5, 1.0}. The row-major ordering claim is the one that mattered and it holds. The Choi (non-Kraus) path is untouched. I measure about 3.2x at dim 8 and 16.

The comment explaining why the identity reshape gives exactly the single-entry operators is a good addition; that is the non-obvious step. Appreciate the explicit disclosure of AI assistance too.

PR #1922, opened 2026-08-04
fix: zero-pad start_vec with Schmidt rank below k in sk-norm lower bound

### LINKED ISSUE #1918: bug: start_vec with Schmidt rank below k crashes in __lower_bound_sk_norm_randomized
Found while closing coverage gaps in #1917. This is a latent bug, not a regression, and no released behavior changes until someone passes this combination of arguments.

In `__lower_bound_sk_norm_randomized` (`toqito/matrix_props/sk_norm.py`), a caller-supplied `start_vec` is accepted when its Schmidt rank is **less than or equal to** `k`:

```python
singular_vals, vt_mat, u_mat = schmidt_decomposition(start_vec, dim)
if (s_rank := len(singular_vals)) > k:
    warnings.warn(...)   # too large: fall back to a random vector
else:
    opt_vec = start_vec
    opt_schmidt = np.zeros((max(dim) * k, max(dim) * k))
    opt_schmidt[: k * dim_a, 0] = np.ravel(vt_mat @ np.diag(singular_vals), order="F")
    opt_schmidt[: k * dim_b, 1] = np.ravel(u_mat, order="F")
```

The two assignments assume the right-hand side has `k * dim_a` (and `k * dim_b`) entries. But `vt_mat` has shape `(dim_a, s_rank)`, so the ravel produces `dim_a * s_rank` entries. Those agree only when `s_rank == k`. When the Schmidt rank is strictly less than `k`, which the guard explicitly allows, the assignment raises an opaque broadcast error.

### Reproduction

```python
import numpy as np
from toqito.matrix_props.sk_norm import __lower_bound_sk_norm_randomized as f

# s_rank == k: fine
f(np.eye(4), k=1, dim=[2, 2], tol=1e-4, start_vec=np.kron([1., 0.], [1., 0.]))
# -> 1.0

# s_rank < k: crashes
f(np.eye(4), k=2, dim=[2, 2], tol=1e-4, start_vec=np.kron([1., 0.], [1., 0.]))
# ValueError: could not broadcast input array from shape (2,) into shape (4,)

f(np.eye(9), k=2, dim=[3, 3], tol=1e-4, start_vec=np.kron([1., 0., 0.], [1., 0., 0.]))
# ValueError: could not broadcast input array from shape (3,) into shape (6,)
```

### Why it has not been noticed

The only existing test that exercises this path, `test_sk_norm_randomized_start_vec_low_schmidt_rank_used_as_init`, uses `k=1` with a product vector. There `s_rank == k == 1`, so the shapes happen to line up and the bug is invisible. Its name suggests it covers the "rank below k" case, but no test actually does.

### Suggested fix

Write the Schmidt data into the leading `s_rank` columns rather than assuming `k`, so a lower-rank seed is zero-padded into the `k`-sized workspace instead of overflowing. Whoever fixes it should decide deliberately whether a lower-rank `start_vec` should be padded and used, or rejected up front with a clear error. Padding looks closer to the intent, since the guard was clearly written to accept it.

### Related

There is a second, independent consequence. Because a lower-rank `start_vec` cannot be supplied, the branch

```python
if (s_rank := schmidt_rank(opt_vec, dim)) < k:
    return __lower_bound_sk_norm_randomized(mat, s_rank, dim, tol, opt_vec)
```

is only reachable when the rank happens to collapse mid-iteration, which did not occur across 12 random seeds. That line is the sole remaining uncovered line in `sk_norm.py` after #1917. Fixing the bug above would also make that branch testable directly.

arnavk23:
## Description

Fixes #1918.

The randomized S(k)-norm lower bound crashed when a caller-supplied `start_vec` had Schmidt rank strictly below `k`. The Schmidt data was written assuming `k` blocks, but `vt_mat`/`u_mat` only have `s_rank` columns, so the raveled length (`dim_a * s_rank`) did not match the assignment slice (`k * dim_a`), raising an opaque broadcast error. The seed is now zero-padded into the `k`-sized workspace so a lower-rank `start_vec` is used as intended.

## Changes

  -  [x] Write the Schmidt components into the leading `s_rank` blocks, zero-padding the remaining `k - s_rank` blocks, matching the intent of the existing `s_rank <= k` guard.
  -  [x] Build the diagonal scaling matrix from `singular_vals.ravel()`. `schmidt_decomposition` returns `singular_vals` as an `(s_rank, 1)` column, so `np.diag` previously extracted the diagonal instead of forming a matrix; that only worked by accident for `s_rank == 1` and raised a matmul mismatch for `s_rank > 1`.
  -  [x] Allocate `opt_schmidt` as `(max(dim) * k, 2)` with `complex` dtype to match the random-init path. The old square shape left all-zero columns that produced a divide-by-zero for `k > 1`, and the real dtype discarded the imaginary part of complex `start_vec`.
  -  [x] Add tests covering `s_rank < k`, `s_rank == k > 1`, and complex `start_vec`. These also make the mid-iteration rank-collapse branch (`schmidt_rank(opt_vec) < k`) reachable from a public call.

## Checklist

### DIFF
--- a/toqito/matrix_props/sk_norm.py
+++ b/toqito/matrix_props/sk_norm.py
@@ -326,9 +326,9 @@ def __lower_bound_sk_norm_randomized(
             )
         else:
             opt_vec = start_vec
-            opt_schmidt = np.zeros((max(dim) * k, max(dim) * k))
-            opt_schmidt[: k * dim_a, 0] = np.ravel(vt_mat @ np.diag(singular_vals), order="F")
-            opt_schmidt[: k * dim_b, 1] = np.ravel(u_mat, order="F")
+            opt_schmidt = np.zeros((max(dim) * k, 2), dtype=complex)
+            opt_schmidt[: s_rank * dim_a, 0] = np.ravel(vt_mat @ np.diag(singular_vals.ravel()), order="F")
+            opt_schmidt[: s_rank * dim_b, 1] = np.ravel(u_mat, order="F")
 
     if opt_vec is None:
         opt_schmidt = np.random.randn(max(dim) * k, 2) + 1j * np.random.randn(max(dim) * k, 2)
--- a/toqito/matrix_props/tests/test_sk_norm.py
+++ b/toqito/matrix_props/tests/test_sk_norm.py
@@ -189,3 +189,27 @@ def test_sk_norm_randomized_start_vec_low_schmidt_rank_used_as_init():
     product = np.kron(np.array([1.0, 0.0]), np.array([1.0, 0.0]))
     result = __lower_bound_sk_norm_randomized(np.eye(4), k=1, dim=dim, tol=1e-4, start_vec=product)
     assert np.isfinite(result)
+
+
+def test_sk_norm_randomized_start_vec_schmidt_rank_below_k_zero_padded():
+    """Private helper: `start_vec` with Schmidt rank < k is zero-padded into the k-sized workspace."""
+    dim = [2, 2]
+    product = np.kron(np.array([1.0, 0.0]), np.array([1.0, 0.0]))
+    result = __lower_bound_sk_norm_randomized(np.eye(4), k=2, dim=dim, tol=1e-4, start_vec=product)
+    np.testing.assert_allclose(result, 1.0)
+
+
+def test_sk_norm_randomized_start_vec_full_schmidt_rank_gt_one():
+    """Private helper: `start_vec` with Schmidt rank == k > 1 is used directly as the iteration seed."""
+    dim = [2, 2]
+    max_ent = max_entangled(2, is_normalized=True).reshape(-1)
+    result = __lower_bound_sk_norm_randomized(np.eye(4), k=2, dim=dim, tol=1e-4, start_vec=max_ent)
+    np.testing.assert_allclose(result, 1.0)
+
+
+def test_sk_norm_randomized_start_vec_complex_below_k():
+    """Private helper: complex `start_vec` with Schmidt rank < k is handled without a casting error."""
+    dim = [3, 3]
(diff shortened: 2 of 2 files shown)


vprusso (approved):
Thanks @arnavk23. This is a better fix than the issue asked for, and I verified all three claims by running them against master.

| case | master | this PR |
| --- | --- | --- |
| `s_rank < k` (k=2, rank-1 seed) | `ValueError: could not broadcast (2,) into (4,)` | 1.000000 |
| `s_rank == k > 1` (k=2, max-entangled seed) | `ValueError: matmul core dimension mismatch` | 1.000000 |
| complex seed, `s_rank < k` | `ValueError: could not broadcast (3,) into (6,)` | 1.000000 |
| control: `k=1`, rank-1 seed | 1.000000 | 1.000000 |

The second row is a bug I did not find when filing #1918, and it is the more consequential one. `np.diag` on the `(s_rank, 1)` column *extracts* a diagonal instead of building one, so it silently returned the right thing only for `s_rank == 1`. Since the recursive call passes `k = s_rank`, that means the mid-iteration rank-collapse branch on line 354 was not merely untested, it would have raised for any collapse to rank above 1. Your diagnosis of the `opt_schmidt` shape is right too: the old square allocation left all-zero columns that the later column-norm division would have hit, which is consistent with only `k = 1` ever working, since there the square shape coincidentally has exactly two columns.

No regression: the full `test_sk_norm.py` passes (23 tests), and the public `sk_operator_norm` returns bit-identical values on master and this branch under a fixed seed for Werner, maximally-entangled, and random-PSD inputs at `k=1` and `k=2`. The random-init path is untouched.

Bonus worth noting: your tests reach line 354, which was the single line I could not cover in #1917. Nice.

PR #1923, opened 2026-08-05
fix(state_opt): forward solver kwargs on the PPT path

idhantsingh027:
`_ppt_primal` and `_ppt_dual` accepted no `**kwargs` and called `problem.solve(solver=solver)`, so the solver options that `state_distinguishability` documents ("Additional arguments to pass to picos' solve method") were silently discarded whenever `measurement="ppt"`. Every non-PPT helper in this module already forwards them, as does every PPT helper in `state_exclusion.py`.

The practical effect is that the PPT path could not be rescued from a solver breakdown. cvxopt exhausts its default iteration budget on PPT problems whose optimum lies on the boundary of the feasible set and fails inside LAPACK with a bare `ArithmeticError: 7`. Raising `max_iterations` is the documented remedy, and it had no effect. `measurement="ppt"` with `strategy="unambiguous"` failed this way for every input tried, which is why that branch had no coverage.

Tests cover the forwarding itself -- picos rejects an unknown option, so the rejection is proof the option arrived -- and the previously unexercised PPT/unambiguous branch, which now returns the exact value 1 for two orthogonal product states along with the expected n+1 POVM elements.

`state_distinguishability.py` goes from 94.50% to 100% coverage.

## Changes

  -  [ ] Change 1

## Checklist

### DIFF
--- a/toqito/state_opt/state_distinguishability.py
+++ b/toqito/state_opt/state_distinguishability.py
@@ -202,9 +202,16 @@ def state_distinguishability(
                 probs=probs,
                 solver=solver,
                 strategy=strategy,
+                **kwargs,
             )
         return _ppt_dual(
-            vectors=vectors, subsystems=subsystems, dimensions=dimensions, probs=probs, solver=solver, strategy=strategy
+            vectors=vectors,
+            subsystems=subsystems,
+            dimensions=dimensions,
+            probs=probs,
+            solver=solver,
+            strategy=strategy,
+            **kwargs,
         )
 
     if strategy == "min_error":
@@ -467,6 +474,7 @@ def _ppt_primal(
     probs: list[float],
     solver: str = "cvxopt",
     strategy: str = "min_error",
+    **kwargs,
 ):
     """Primal problem for the SDP with PPT constraints."""
     n = len(vectors)
@@ -500,7 +508,7 @@ def _ppt_primal(
                     problem.add_constraint(probs[j] * dms[j] | measurements[i] == 0)
 
     problem.set_objective("max", np.real(picos.sum([probs[i] * (dms[i] | measurements[i]) for i in range(n)])))
-    solution = problem.solve(solver=solver)
+    solution = problem.solve(solver=solver, **kwargs)
 
     return solution.value, measurements
 
@@ -512,6 +520,7 @@ def _ppt_dual(
     probs: list[float],
     solver: str = "cvxopt",
     strategy: str = "min_error",
+    **kwargs,
 ):
     """Semidefinite program with PPT constraints (dual problem)."""
     d = vectors[0].shape[0]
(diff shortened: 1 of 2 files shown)


vprusso (changes requested):
Thanks @idhantsingh027 - the diagnosis and the fix are both right, and this is a real bug worth fixing. `_ppt_primal`/`_ppt_dual` genuinely did drop the solver options the docstring promises to forward, while every other helper in the module forwards them. I confirmed the fix works: with this branch, an unknown option now reaches picos on the PPT path and is rejected (`LookupError: Unknown option 'bogus_option'`), where on master it was silently swallowed.

One change needed before this can merge. CI is red on all three Ubuntu jobs, and I reproduced it locally on Linux:

```
FAILED test_state_distinguishability_ppt_unambiguous - ArithmeticError: 7
cvxopt/misc.py:612: ArithmeticError
```

38 pass, that one fails. Your other two new tests pass.

The nice part is that this is exactly the failure your own PR description diagnoses: cvxopt exhausting its iteration budget on a PPT problem whose optimum sits on the boundary. The remedy you identified works, it just is not applied in the test:

| invocation | result |
| --- | --- |
| as written in the test (defaults) | `ArithmeticError: 7` |
| `max_iterations=500` | value `1.000000`, 3 POVM elements |

So `test_state_distinguishability_ppt_unambiguous` should pass `max_iterations=500` (or whatever margin you prefer). That both fixes the failure and makes the test demonstrate the capability this PR adds, which is a stronger regression test than the current form: it would fail if the forwarding regressed, since the solve would break down again.

It passing on macOS and failing on Linux is just a difference in the cvxopt/LAPACK build, so please do not rely on a local macOS run for this one.

Everything else looks good to me. Re-request review once that is in and I will merge.

idhantsingh027:
Done - ..._ppt_unambiguous now passes max_iterations=500, and I get your numbers locally (value 1.000000, 3 POVM elements). Agreed it's the stronger form; noted in the docstring that a forwarding regression breaks that solve too.

One correction on the mechanism, since it affects which direction the knob goes. It isn't cvxopt exhausting its budget, picos leaves max_iterations unset, and the cvxopt backend turns that into maxiters = 1e6. cvxopt never meets its convergence tolerance on these PPT problems, so with a budget that large it keeps iterating while its scaling degrades until LAPACK fails. Capping at 500 reduces the budget and stops it before the breakdown. So raising the cap makes it more fragile, not less. I also tried cvxopt_kktreg (1e-7 and 1e-9) and it doesn't help; only the cap does.

That points at a separate question this PR doesn't touch: whether the PPT helpers should set a default iteration cap themselves, so measurement="ppt" doesn't hand users an ArithmeticError: 7 out of the box. Happy to open an issue or a follow-up PR if you think it's worth it.

vprusso (approved):
Fixed exactly as discussed, thanks. The full matrix is green and I get 39/39 locally on Linux, where the previous revision failed.

Nice touch adding the docstring note that `max_iterations` only reaches the solver because of this change's forwarding. That is what turns the test from a workaround into a real regression test: if the forwarding is ever dropped again, the option stops arriving, cvxopt breaks down, and the test fails loudly instead of silently passing.

Good find on the underlying bug, and the reasoning in the description (every other helper in the module forwards, the PPT ones did not, and the consequence was that the documented remedy for a solver breakdown had no effect) was exactly right.

Issue #1924, opened 2026-08-07
docs: BB84 parallel-repetition gallery uses wrong predicate axis order


Devansh-ops:
## What is incorrect?

`docs/content/examples/extended_nonlocal_games/parallel_repetition.py` constructs the BB84 predicate tensor with the axes ordered as `(alice_out, bob_out, alice_in, bob_in, referee_row, referee_col)`. `ExtendedNonlocalGame` expects `(referee_row, referee_col, alice_out, bob_out, alice_in, bob_in)`.

Because every dimension in this example is 2, the invalid ordering is shape-compatible and is accepted silently, but it represents a different game.

## The issue

Running the gallery example from current `master` prints:

```text
Unentangled value:    0.75000
Non-signaling value:  0.75000
Exact (cos²(π/8)):    0.85355
ω(G²) unentangled:   0.56250
ω(G)² expected:      0.72855
Strong parallel rep holds: False
ω_ns(G²):            0.56250
ω_ns(G)²:            0.72855
ω_ns(G²) > ω_ns(G)²: False
```

This contradicts both the surrounding narrative and the cited results. It also differs from the correctly ordered BB84 construction in `enlg_bb84.py` and the existing `ExtendedNonlocalGame` tests.

This is the remaining numerical defect in the tutorial introduced by #1486 for #21; #1714 corrected its references but did not change this construction.

## Correct statement

The gallery should construct `pred_mat[:, :, a, b, x, y]`. With only that construction corrected, current `ExtendedNonlocalGame` methods return approximately:

- single-game unentangled value: `0.85355339 = cos²(π/8)`;
- single-game non-signaling value: `0.85355`;
- two-repetition unentangled value: `0.72855339 = cos⁴(π/8)`; and
- two-repetition non-signaling value: `0.7382`, which is strictly greater than `cos⁴(π/8)`.

The single-game and unentangled parallel-repetition results are supported by Johnston, Mittal, Russo, and Watrous, *Extended non-local games and monogamy-of-entanglement games* (2016), DOI 10.1098/rspa.2016.0003. Russo's thesis, *Extended Nonlocal Games* (2017), Section 6.2, equations (6.47)–(6.49), reports `ω_ns(G_BB84²) ≈ 0.73826` and `ω_ns(G_BB84) = cos²(π/8)`.

## Impact

This appears limited to the executable gallery construction. The public `ExtendedNonlocalGame` methods already produce the cited values when given the correctly ordered predicate tensor, and their existing BB84 tests pass.

PR #1925, opened 2026-08-07
test(docs): assert BB84 parallel repetition values

Devansh-ops:
Follow-up to #1924 and #1926.

## Description

#1926 corrected the BB84 predicate-axis ordering in the parallel-repetition gallery. This follow-up adds executable regression assertions so the gallery fails loudly if its computed values ever drift from the cited results instead of silently rendering contradictory output.

## Changes

- [x] Assert the single-game unentangled and non-signaling values against `cos²(π/8)`.
- [x] Assert the two-repetition unentangled value against `cos⁴(π/8)`.
- [x] Assert the two-repetition non-signaling value against the `0.73826` numerical value reported in Russo's thesis.

No library code, predicate construction, or narrative claims are changed.

## Validation

- Rebased onto current `master` after #1926.
- Executed the parallel-repetition gallery successfully with all assertions passing.
- Passed the four focused `ExtendedNonlocalGame` BB84 tests.
- Passed forced Ruff check and format on the gallery file.
- Passed `git diff --check`.

## Checklist

- [x] The title is concise and informative.
- [x] The description explains the regression protection.
- [x] Linked the diagnosing issue and merged mechanical fix.
- [x] Added executable regression assertions.
- [ ] Full upstream CI refresh.

## AI assistance

Codex assisted with collision checking, diagnosis, rebase conflict resolution, and drafting. The contribution was reviewed against the cited papers, executable outputs, repository conventions, maintainer feedback, and existing regression tests.

### DIFF
--- a/docs/content/examples/extended_nonlocal_games/parallel_repetition.py
+++ b/docs/content/examples/extended_nonlocal_games/parallel_repetition.py
@@ -93,6 +93,8 @@ omega_ue = bb84.unentangled_value()
 omega_ns = bb84.nonsignaling_value()
 exact = np.cos(np.pi / 8) ** 2
 
+np.testing.assert_allclose([omega_ue, omega_ns], exact, rtol=1e-3)
+
 print(f"Unentangled value:    {omega_ue:.5f}")
 print(f"Non-signaling value:  {omega_ns:.5f}")
 print(f"Exact (cos²(π/8)):    {exact:.5f}")
@@ -114,6 +116,8 @@ bb84_2_reps = ExtendedNonlocalGame(bb84_prob_mat, bb84_pred_mat, reps=2)
 
 omega_ue_2 = bb84_2_reps.unentangled_value()
 expected_ue_2 = exact**2
+
+np.testing.assert_allclose(omega_ue_2, expected_ue_2, rtol=1e-3)
 
 print(f"ω(G²) unentangled:   {omega_ue_2:.5f}")
 print(f"ω(G)² expected:      {expected_ue_2:.5f}")
@@ -138,6 +142,9 @@ print(f"Strong parallel rep holds: {np.isclose(omega_ue_2, expected_ue_2, atol=1
 # %%
 omega_ns_2 = bb84_2_reps.nonsignaling_value()
 expected_ns_2 = exact**2
+known_ns_2_value = 0.73826
+
+np.testing.assert_allclose(omega_ns_2, known_ns_2_value, rtol=1e-3)
 
 print(f"ω_ns(G²):            {omega_ns_2:.5f}")
 print(f"ω_ns(G)²:            {expected_ns_2:.5f}")


vprusso:
Thanks @Devansh-ops, and thanks especially for #1924 - that issue did the real work here. It named the exact axis permutation, explained why the wrong order is silently shape-compatible (every dimension is 2), and predicted the corrected values with citations. I confirmed all four by running the corrected construction:

| quantity | computed | your prediction |
| --- | ---: | --- |
| single-game unentangled | 0.85355339 | $\cos^2(\pi/8)$ ✓ |
| single-game non-signaling | 0.85355834 | 0.85355 ✓ |
| two-repetition unentangled | 0.72855339 | $\cos^4(\pi/8)$ ✓ |
| two-repetition non-signaling | 0.73819036 | 0.7382, $> \cos^4(\pi/8)$ ✓ |

Two notes.

**This is still a draft.** Meanwhile #1926 landed the identical core change as a non-draft. If you intend this one to be the fix, mark it ready for review and say so; I would rather merge the version from the person who diagnosed the problem. The assertions you add are a genuine improvement over the minimal fix, since they make the gallery fail loudly if the construction ever drifts back rather than silently printing wrong numbers again.

**One item needs the maintainer's sign-off, not mine.** You rewrite implication 3 from "the quantum value under parallel repetition remains an open question ... including BB84" to "the standard quantum value obeys strong parallel repetition for this BB84 game." My reading agrees with you: for the BB84 monogamy-of-entanglement game, strong parallel repetition for the standard quantum value is a known result. But the gallery itself only computes the unentangled and non-signaling values, so that sentence is asserting something the page does not demonstrate, and @vprusso is a co-author of both cited works. I have flagged it for his call rather than approving it myself.

If you would rather not carry the narrative change, splitting it into a separate PR would let the mechanical fix land immediately.

vprusso:
@Devansh-ops following up to make this concrete: please mark this ready for review and I will merge it as the fix for #1924, since you diagnosed the problem.

Two things to fold in first:

1. Please carry over the explanatory comment from #1926, which this branch does not have. Something like the following just above the construction, so the layout requirement is stated where the mistake was made and does not silently drift back:

   ```python
   # `ExtendedNonlocalGame` expects the referee's row and column to be the
   # leading axes: `pred_mat[:, :, a, b, x, y]` holds V(a,b|x,y).
   ```

2. On implication 3: please drop the rewritten sentence from this PR. The claim itself is defensible, but the gallery only computes the unentangled and non-signaling values, so as written the page asserts something about the standard quantum value that it never demonstrates. Either restore the original wording, or scope the sentence to what the page actually shows, namely that the unentangled value obeys strong parallel repetition here while the non-signaling value does not, which your own numbers establish. If you want to make the stronger statement about the quantum value, a separate PR citing the relevant result can be reviewed on its own merits.

Everything else, including the assertions, is good and I would like to keep it.

vprusso:
@Devansh-ops update, and an apology for the reversal.

I have reopened and queued #1926 (the equivalent mechanical fix) rather than continue waiting on this draft. The reason is only timing: the gallery currently renders values that contradict its own narrative on every claim it makes, and that is a live defect on the published docs. Holding it open for a draft was the wrong call on my part, and reversing myself twice is on me, not on you.

To be clear about credit: #1924 did the actual diagnostic work here. It identified the exact axis permutation, explained why the wrong order is silently shape-compatible, and predicted all four corrected values with citations, every one of which I confirmed. I have said so in the merge commit message, and the issue is what #1926 closes.

This PR is still worth landing for the part that is genuinely yours and is not in #1926: the assertions. Once #1926 is in, please rebase and keep only

- `np.testing.assert_allclose` on the single-game values, the two-repetition unentangled value, and the two-repetition non-signaling value against 0.73826

so the gallery fails loudly rather than silently printing wrong numbers if the construction ever drifts again. That is the durable protection against this recurring, and the minimal fix does not provide it.

Please still drop the implication-3 rewrite from this PR, for the reason given earlier: the page only computes the unentangled and non-signaling values, so it should not assert a result about the standard quantum value that it never demonstrates. Happy to review that separately if you want to make the case with a citation.

vprusso:
@Devansh-ops status note: #1926 merged, so the axis-order fix is now on master and #1924 is closed. This branch has gone to conflicting as a result, since it carries the same change.

To land the part that is still uniquely yours, please rebase onto master and keep only the `np.testing.assert_allclose` checks on the single-game values, the two-repetition unentangled value, and the two-repetition non-signaling value against 0.73826. The axis fix itself is already in, so it should drop out in the rebase.

As noted before, please leave the implication-3 rewrite out of this one; it can be reviewed separately with a citation if you want to make that case.

If you would rather not carry this further, say so and I will close it. No obligation, and the diagnosis in #1924 was the valuable part regardless.

Devansh-ops:
Rebased onto current master after #1926 and marked ready for review. The PR now contains only the requested np.testing.assert_allclose regression checks; the duplicate axis-order change and implication-3 rewrite have been removed. The gallery ran with all assertions passing, the four focused BB84 tests passed, forced Ruff check/format passed, and git diff --check passed.

shinzoxD (approved):
Independently reviewed and validated the rebased assertions-only diff. The full parallel-repetition gallery completed with values 0.85355/0.85356, 0.72855, and 0.73819; all new assert_allclose checks passed. I also ran the four focused BB84 unentangled/non-signaling tests (4 passed), the repository-wide Ruff check and format check, and git diff --check. The tolerances accommodate the observed solver variance while remaining tight enough to catch a meaningful regression. No blocking findings.

vprusso (approved):
Thanks for rebasing this down, @Devansh-ops - this is exactly the scope I was hoping for. Assertions only, no axis change (already on master via #1926), no narrative rewrite.

These are the durable part of your contribution. The mechanical fix stops the gallery being wrong today; these stop it going wrong again silently, which is how the original defect survived from #1486 through #1714 unnoticed.

Checked that the tolerances are meaningful rather than vacuous: `rtol=1e-3` against `0.73826` passes on the actual value 0.73819 with room to spare, while the gap between that and $\cos^4(\pi/8) = 0.72855$ is about 1.3%, comfortably outside the tolerance. So the assertion would actually catch a regression back to the pre-#1926 behavior rather than silently accepting it.

And thank you again for #1924. Diagnosing this was the hard part, and it was a genuinely excellent issue.

Issue #1915, opened 2026-08-04
test: raise coverage of is_separable.py


vprusso:
`toqito/state_props/is_separable.py` is the least-covered module in the project, at about 88% with roughly 32 uncovered lines. Everything else is at or near 100%, and the repository overall sits just under 99%.

This module matters more than the number suggests: it dispatches across a series of separability criteria, and the uncovered lines are mostly individual criteria branches that never get exercised. A bug in one of those branches would return a wrong separability verdict rather than raising, which is the kind of failure that is easy to miss.

### What to do

Find the uncovered lines and add tests that reach them:

```
uv run pytest --cov-config=.coveragerc --cov=toqito --cov-report=term-missing \
  toqito/state_props/tests/test_is_separable.py
```

Use the `Missing` column to see which lines are unreached. Note the `--cov=toqito` part: running coverage against a single submodule triggers a numpy "cannot load module more than once" error, so keep it as written.

Then, for each uncovered branch, construct a state that actually exercises it. Good sources of states with known separability status:

- `toqito.states` (for example `bell`, `isotropic`, `werner`, `horodecki`, `max_mixed`, `tile`)
- states already used elsewhere in `toqito/state_props/tests/`

Prefer a few well-chosen states that each light up a specific criterion over many random ones. A test whose name says which criterion it covers is much more useful than a generic one.

### Scope

This does not need to be done in one pass. A PR that covers even a handful of branches is welcome, and it is easier to review that way. Say in the PR which criteria you targeted and what the coverage number moved from and to.

### Verifying

```
uv run pytest toqito/state_props/tests/test_is_separable.py -q
uv run ruff check toqito/state_props/tests/test_is_separable.py
```

Please do not change `is_separable.py` itself in the same PR. If a new test uncovers an actual bug, that is a valuable find, so open a separate issue or PR for the fix so the two can be reviewed independently.

Karthikvenugopal:
Hi, I’d like to work on this issue. I’ll identify a small set of currently uncovered separability-criterion branches and add focused regression tests using states with known separability results.

I’ll keep the changes limited to the test file and report the coverage improvement in the PR.

PR #1928, opened 2026-08-09
test: cover four unexercised criterion branches in is_separable

dchaudhari7177:
Towards #1915. A first pass, kept small so it is easy to review, as you suggested.

## Which criteria

`is_separable` runs its criteria in a chain and returns on the first verdict, so the later ones are unreachable end-to-end — a cheaper criterion always decides first. That is also why they were uncovered. A wrong verdict from one of those branches would surface as a wrong separability answer, not an exception, so I tested each helper directly on a state chosen to land on it:

| criterion | state | branch |
|---|---|---|
| Reduction (Horodecki 1999) violated | 3x3 maximally entangled | `_reduction_ppt_criterion` |
| Choi 1975 positive-map witness | 3x3 maximally entangled | `_positive_map_witness_criteria` |
| Breuer-Hall positive-map witness | 4x4 maximally entangled | `_positive_map_witness_criteria` |
| rank-4 3x3 range criterion abstaining below rank 4 | rank-1 product state | `_range_projector_product_overlap_3x3_rank4` |

Testing the private helpers rather than going through `is_separable` follows what the file already does for `_filter_normal_form`, `_iterative_product_state_subtraction` and `_range_projector_product_overlap_3x3_rank4`.

Every positive test is paired with a control on the maximally mixed state asserting the criterion **abstains** (returns `None`). Without those, a helper that fired on everything would pass the positive tests.

I did not touch `is_separable.py`, per the issue.

## Coverage

`is_separable.py`: **92.18% -> 93.28%** (32 -> 28 uncovered lines).

```
uv run pytest --cov-config=.coveragerc --cov=toqito --cov-report=term-missing \
  toqito/state_props/tests/test_is_separable.py
```

`ruff check` and `ruff format --check` are clean on the test file.

## Two branches that look unreachable

While mapping the uncovered lines I hit three I could not reach by any input, because an earlier guard subsumes them. Reporting rather than testing, since a test would have to fake the arguments to get there:

- **`is_separable.py:1118` and `:1124`** — `_horodecki_low_rank_ppt_criteria` returns early on `state_rank <= max_dim_val`. The next two checks are `state_rank <= rank(rho_A)` and `state_rank <= rank(rho_B)`, but `rank(rho_A) <= d_A <= max_dim_val`, so anything satisfying them already returned. (The placeholder `test_horodecki_rank_le_marginal_rank` in the test file targets exactly these, which is presumably why it was left as `pass`.)
- **`is_separable.py:1230`** — `_qubit_qudit_ppt_criteria` returns early unless `min_dim_val == 2`. Reaching line 1230 needs `min_dim_val == 2` together with `d_a != 2` and `d_b != 2`, and `min(d_a, d_b) == 2` makes one of them 2.

Happy to open a separate issue for these if you'd like them tracked — I left them alone here rather than mixing a source change into a test PR.

I have tests for a few more branches (`_filter_normal_form`, the iterative subtraction budget) that I'd rather send as a follow-up once you've had a look at the shape of this one.

## AI disclosure

Written with Claude Code (Claude Opus 5): it mapped the uncovered lines to criteria, searched for states that reach each branch, and wrote the tests and this description. The coverage numbers, the test run and the unreachability arguments above were checked by hand.

### DIFF
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -7,6 +7,7 @@ import numpy as np
 import pytest
 
 from toqito.channel_ops.partial_channel import partial_channel
+from toqito.channels import partial_trace
 from toqito.matrix_props import is_density, is_positive_semidefinite
 from toqito.matrix_props.trace_norm import trace_norm
 from toqito.rand import random_density_matrix
@@ -19,7 +20,10 @@ from toqito.state_props.is_separable import (
     _filter_cmc_xi_sum,
     _filter_normal_form,
     _iterative_product_state_subtraction,
+    _positive_map_witness_criteria,
+    _qubit_qudit_ppt_criteria,
     _range_projector_product_overlap_3x3_rank4,
+    _reduction_ppt_criterion,
 )
 from toqito.states import basis, bell, horodecki, isotropic, max_entangled, tile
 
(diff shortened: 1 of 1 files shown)


vprusso (approved):
This is exactly the right way to attack #1915. Verified end to end.

**Coverage.** Reproduced by running the file with and without the new tests:

| | coverage | missed | partial |
| --- | ---: | ---: | ---: |
| without | 91.63% | 34 | 27 |
| with | 93.55% | 27 | 20 |

Seven fewer uncovered lines and seven fewer partial branches.

**The controls are what make this worth merging.** Pairing every positive test with an abstention check on the maximally mixed state is the part people usually skip. Without it, a witness that fired on everything would pass all four positive tests and look like coverage while testing nothing. Testing the helpers directly rather than through `is_separable` is also right, and it matches what the file already does elsewhere; going end-to-end, a cheaper criterion decides first and these branches are never reached, which is the reason they were uncovered to begin with.

**Your two unreachability findings are correct.** I checked the argument and then the code:

- Lines 1118 and 1124: there is a single call site, and `min_dim_val, max_dim_val = min(dA, dB), max(dA, dB)` (line 475), so the parameters are always consistent. Since `rank(rho_A) <= d_A <= max_dim_val`, any state satisfying `state_rank <= rank(rho_A)` already satisfies `state_rank <= max_dim_val` and returned at line 1110. 320 random trials across eight dimension pairs never reached either line.
- Line 1230: past the guard, `min_dim_val == 2` means `min(d_a, d_b) == 2`, so one of them is 2; reaching 1230 needs both `d_a != 2` and `d_b != 2`. Not satisfiable.

Good catch that `test_horodecki_rank_le_marginal_rank` is a bare `pass` stub aimed at precisely those lines. That explains why it was never filled in: it cannot be.

I have opened #1930 to track the three dead branches, since removing them is a source change and you were right to keep it out of a test-only PR.

Merging. Please do keep going on #1915 if you are willing.

PR #1931, opened 2026-08-09
refactor: remove unreachable separability branches

### LINKED ISSUE #1930: Three unreachable branches in is_separable (1118, 1124, 1230)
Three branches in `toqito/state_props/is_separable.py` cannot be reached from any input. Found by @dchaudhari7177 while working on #1915 (see #1928), and verified independently.

Both helpers have exactly one call site, and at that site the dimension parameters are computed as

```python
min_dim_val, max_dim_val = min(dA, dB), max(dA, dB)   # is_separable.py:475
```

so they are always consistent with `d_a`/`d_b`. That consistency is what makes the branches dead.

### `_horodecki_low_rank_ppt_criteria`, lines 1118 and 1124

```python
if state_rank <= max_dim_val:
    return (True, "rank(rho) <= max(d_A, d_B) ..."), None, None   # line 1110
...
if state_rank <= rank_marg_a:      # line 1117 -> 1118 unreachable
if state_rank <= rank_marg_b:      # line 1123 -> 1124 unreachable
```

`rho_A` acts on a `d_A`-dimensional space, so `rank(rho_A) <= d_A <= max(d_A, d_B) == max_dim_val`. Any state with `state_rank <= rank(rho_A)` therefore also has `state_rank <= max_dim_val` and already returned at line 1110. The same argument applies to `rank(rho_B)`.

Empirically: 320 random density matrices across the dimension pairs (2,2), (2,3), (3,2), (2,4), (3,3), (3,4), (4,4), (2,5) never reached either line.

Note that `test_horodecki_rank_le_marginal_rank` in `toqito/state_props/tests/test_is_separable.py:373` is a bare `pass` stub aimed at exactly these two lines. It was presumably left unfinished because the case cannot be constructed.

### `_qubit_qudit_ppt_criteria`, line 1230

```python
if min_dim_val != 2 or prod_dim_val == 0:
    return None                     # line 1219
...
if d_a != 2 and d_b == 2:
    ...
elif d_a != 2:
    return None                     # line 1230 unreachable
```

Past the guard, `min_dim_val == 2` means `min(d_a, d_b) == 2`, so at least one of `d_a`, `d_b` equals 2. Reaching line 1230 requires `d_a != 2` and `d_b != 2` simultaneously, which contradicts that.

### What to do

Decide per branch whether the code is dead or the guard above it is too strong:

- If these are genuinely redundant, delete them. That is the reading the arguments above support, and it removes three permanently-uncovered branches plus the stub test.
- If instead any was meant to catch a case the earlier guard is wrongly swallowing, the guard is the thing to fix, and that would be a behavior change needing its own tests.

Either way this is a source change, so it does not belong in a test PR. Whoever picks it up should also delete the `test_horodecki_rank_le_marginal_rank` stub.

Removing the dead branches would also let `is_separable.py` reach full branch coverage, which #1915 otherwise cannot achieve.

fzlzjerry:
## Description

This is intended to be a behavior-preserving refactor. It removes three branches in `is_separable` that cannot be reached under the dimension and rank invariants established by their callers. The marginal ranks cannot exceed their subsystem dimensions, so the earlier `rank(rho) <= max(d_A, d_B)` check always wins; after `min(d_A, d_B) == 2`, the 2xN fallback cannot observe both dimensions as non-2.

Closes #1930

## Changes

- remove the redundant marginal-rank verdicts and their now-unused tolerance parameter
- remove the contradictory 2xN fallback
- delete the skipped placeholder test for an input that cannot be constructed

## Behavior preservation

The maintainer compared `is_separable` before and after this change across 112 inputs, covering random density matrices, random separable product mixtures over multiple bipartite dimensions, and representative entangled and bound-entangled states. There were zero differences in either verdicts or reason strings.

## Validation

- `uv run pytest toqito/state_props/tests/test_is_separable.py -q` — 139 passed, 5 skipped, 2 xfailed
- `uvx --from ruff==0.15.0 ruff check toqito/state_props/is_separable.py toqito/state_props/tests/test_is_separable.py`
- `uvx --from ruff==0.15.0 ruff format --check toqito/state_props/is_separable.py toqito/state_props/tests/test_is_separable.py`
- `git diff --check`
- GitHub CI passed the full test matrix on Python 3.12, 3.13, and 3.14 across Ubuntu and macOS

## Checklist

- [x] Ruff lint and formatting checks pass with the version pinned by CI.
- [x] The complete `is_separable` test module passes.
- [x] The full GitHub CI test matrix passes.

### DIFF
--- a/toqito/state_props/is_separable.py
+++ b/toqito/state_props/is_separable.py
@@ -548,7 +548,7 @@ def is_separable(
 
     # --- 7. Operational Criteria for Low-Rank PPT States (Horodecki et al. 2000) ---
     verdict, rho_A_marginal, rho_B_marginal = _horodecki_low_rank_ppt_criteria(
-        current_state, dims_list, state_rank, max_dim_val, tol
+        current_state, dims_list, state_rank, max_dim_val
     )
     if verdict is not None:
         return verdict
@@ -1104,7 +1104,6 @@ def _horodecki_low_rank_ppt_criteria(
     dims: list[int],
     state_rank: int,
     max_dim_val: int,
-    tol: float,
 ) -> tuple[tuple[bool, str] | None, np.ndarray | None, np.ndarray | None]:
     """Run the Horodecki low-rank PPT criteria and return cached marginals."""
     if state_rank <= max_dim_val:
@@ -1112,20 +1111,6 @@ def _horodecki_low_rank_ppt_criteria(
 
     rho_a_marginal = partial_trace(state, sys=[1], dim=dims)
     rho_b_marginal = partial_trace(state, sys=[0], dim=dims)
-    rank_marg_a = np.linalg.matrix_rank(rho_a_marginal, tol=tol)
-    rank_marg_b = np.linalg.matrix_rank(rho_b_marginal, tol=tol)
-    if state_rank <= rank_marg_a:
-        return (
-            (True, f"rank(rho)={state_rank} <= rank(rho_A)={rank_marg_a} (Horodecki et al. 2000)"),
-            rho_a_marginal,
-            rho_b_marginal,
-        )
-    if state_rank <= rank_marg_b:
-        return (
-            (True, f"rank(rho)={state_rank} <= rank(rho_B)={rank_marg_b} (Horodecki et al. 2000)"),
-            rho_a_marginal,
-            rho_b_marginal,
-        )
     return None, rho_a_marginal, rho_b_marginal
 
 
@@ -1226,8 +1211,6 @@ def _qubit_qudit_ppt_criteria(
     if d_a != 2 and d_b == 2:
         state_t_2xn = swap(state, sys=[1, 2], dim=dims)
         dim_for_hildebrand_map = [d_b, d_a]
-    elif d_a != 2:
-        return None
 
     if state_t_2xn is state:
         current_lam_2xn = sorted_eigs_desc
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -369,12 +369,6 @@ def test_skip_horodecki_if_not_applicable_proceeds_entangled_tiles(tiles_state_3
 # --- Specialized Tests (Edge Cases, Mocking, XFAIL for Incomplete Features) ---
 
 
-@pytest.mark.skip(reason="Requires specific state for Horodecki rank <= marginal rank test.")
-def test_horodecki_rank_le_marginal_rank():
-    """Placeholder for PPT state where rank <= marginal rank implies separability."""
-    pass
-
-
 def test_entangled_by_reduction_criterion_non_psd_choi_T():
     """is_separable expects PSD input; Choi of Transpose map is non-PSD for d>1."""
     d = 3


vprusso:
@fzlzjerry this is correct and I would like to merge it. It is still marked draft, so please mark it ready for review when you are happy with it.

I verified it rather than relying on the argument alone, since it is a source change that deletes branches:

**Behavior is unchanged.** Compared `is_separable` verdicts before and after across 112 inputs, spanning random density matrices and random separable product mixtures at (2,2), (2,3), (3,2), (2,4), (4,2), (3,3), (3,4), (4,4), (2,5), plus a Bell state, an isotropic state, a 3x3 Horodecki state, and a maximally entangled state. **Zero differing verdicts**, including the reason strings.

**The reasoning holds.** Both helpers have a single call site, and `min_dim_val, max_dim_val = min(dA, dB), max(dA, dB)`, so the invariants you rely on are guaranteed there. `rank(rho_A) <= d_A <= max_dim_val` makes the marginal-rank verdicts unreachable past the earlier `state_rank <= max_dim_val` return, and `min(d_a, d_b) == 2` makes `d_a != 2 and d_b != 2` unsatisfiable.

**Effect.** `is_separable.py` goes 93.55% -> 93.99%, statements 483 -> 475, and the test module runs 139 passed with one fewer skip now that the placeholder is gone.

Dropping the now-unused `tol` parameter is right, and I checked there are no other callers to break.

One request before merging: your checklist notes the full suite did not finish locally. CI is green across the matrix here, so that is covered, but please say explicitly in the description that the change is intended to be behavior-preserving and that verdicts were compared, since a future reader looking at a diff that deletes two `return (True, ...)` paths from a separability routine will want that stated up front. Feel free to cite the comparison above.

fzlzjerry:
Updated the description to call out the behavior-preserving intent and record the 112-input before/after comparison, including the zero differences in verdicts and reason strings. Thanks for validating the invariants and the full CI matrix.

shinzoxD (approved):
Verified at 01642e27. Both helpers have one production call site, where min/max dimensions are derived directly from d_A and d_B. Therefore rank(rho_A/B) <= d_A/B <= max(d_A, d_B) makes the deleted marginal-rank returns unreachable after the earlier rank guard, and min(d_A, d_B) == 2 makes the removed 2xN fallback unsatisfiable. Locally, the complete affected module passed with 139 passed, 5 skipped, and 2 xfailed; Ruff 0.15 lint and format checks passed; and git diff --check is clean.

vprusso (approved):
Thanks for marking this ready and rebasing, @fzlzjerry. Re-checked against current master, since master moved a fair bit while this sat (#1928, #1932, #1933, #1934, #1935 all landed).

**Still correct on the rebased branch.** The unreachability arguments depend on structure that is unchanged: line 475 still computes `min_dim_val, max_dim_val = min(dA, dB), max(dA, dB)`, so the parameters remain consistent at the single call site; line 1110 still guards the marginal-rank verdicts; line 1219 still guards the 2xN fallback. Since `rank(rho_A) <= d_A <= max_dim_val`, anything reaching the first pair already returned, and `min(d_a, d_b) == 2` makes `d_a != 2 and d_b != 2` unsatisfiable.

**Compatible with the newer tests.** `test_is_separable.py` runs 146 passed, 5 skipped, 2 xfailed on this branch, one fewer skip than master now that the placeholder is gone. That run includes the helper-level tests added in #1928 and #1933, several of which call `_qubit_qudit_ppt_criteria` directly, so the removal does not disturb them.

On behavior preservation: my earlier comparison across 112 inputs found zero differences in verdicts or reason strings. I tried to re-run it against current master and it exceeded my time budget, so I am relying on the earlier run. That is sound here because the deleted code and the guards above it are byte-identical to what was compared then, but I would rather say so than imply a fresh run I did not complete.

Dropping the now-unused `tol` parameter is right, and there are still no other callers.

Merging. Note this also unblocks #1915: lines 1118, 1124 and 1230 were the last permanently-uncoverable branches, so full branch coverage of `is_separable.py` becomes reachable now.

PR #1933, opened 2026-08-11
test: cover the DPS and eigenvalue helpers' error-handling branches

dchaudhari7177:
Refs #1915

Taking the issue up on "a PR that covers even a handful of branches is welcome, and it is easier to review that way."

## Coverage

**`toqito/state_props/is_separable.py`: 93.28% → 94.79%** (uncovered lines 28 → 17). Lines now covered: **1354-1359, 1368-1372.**

## Criteria targeted, and why these

Both are helpers that **degrade rather than propagate**, which makes them the worst place to have no coverage: a regression that turned an abstain into a confident verdict would return a *wrong separability answer* instead of raising — exactly the failure mode the issue describes as easy to miss.

**`_dps_hierarchy_criterion`** — neither `except` handler was reached:

- `ImportError` (no CVXPY/solver installed) → asserts `None`, i.e. "cannot answer", **not** `False`
- a generic `RuntimeError` (solver failed to converge) → same
- the `has_symmetric_inner_extension` call, since the second call in the `try` goes through the same handler
- the `level < 2` early return, for completeness

**`_sorted_real_eigs_desc`** — neither `LinAlgError` fallback was reached:

- `eigvalsh` fails → the `eigvals` fallback still returns correct descending eigenvalues
- both fail → returns `None`, which is what the caller checks before skipping the Vidal–Tarrach stage
- plus the happy path, so the fallbacks are known to *be* fallbacks

These patch `numpy.linalg` globally but call the helper directly, so they don't trip the `is_positive_semidefinite` / `eigvalsh` interaction the file's existing comment warns about.

## Deliberately not attempted

Worth saying explicitly, since they're the bulk of what's left:

- **1118, 1124, 1230** — the unreachable branches from #1930, being removed in #1931. Writing tests for code that's about to be deleted seemed like wasted review time.
- **469, 479, 481** — defensive `raise`s after the dimension arithmetic. They look unreachable for the same reason: a zero local dimension with a non-empty state is already rejected by the product-mismatch check above them. I'd rather flag that than assert it — probably belongs in #1930's sweep.
- **904, 916, 1307** — 904/916 are `_filter_normal_form` giving up after its iteration budget; 1307 guards `1 - t + t**2` against being zero, which it never is for real `t` (minimum 0.75). Reaching these needs a purpose-built pathological state, or dead-code removal.

Happy to take another pass at the genuine criteria branches (578, 741→752, 811→826, 1044, 1067, 1237-1238, 1258-1279) in a follow-up — those need states with known separability status rather than monkeypatching, so they're a different kind of work.

**`is_separable.py` itself is untouched**, per the issue.

## Verification

```
pytest toqito/state_props/tests/test_is_separable.py
→ 146 passed, 6 skipped, 2 xfailed

ruff check / ruff format --check → clean
```

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -16,6 +16,7 @@ from toqito.state_props.is_ppt import is_ppt
 from toqito.state_props.is_separable import (
     _TERHAL_2000_TILE_WITNESS_3X3,
     _choi_1975_choi_matrix,
+    _dps_hierarchy_criterion,
     _filter_cmc_bound,
     _filter_cmc_xi_sum,
     _filter_normal_form,
@@ -24,6 +25,7 @@ from toqito.state_props.is_separable import (
     _qubit_qudit_ppt_criteria,
     _range_projector_product_overlap_3x3_rank4,
     _reduction_ppt_criterion,
+    _sorted_real_eigs_desc,
 )
 from toqito.states import basis, bell, horodecki, isotropic, max_entangled, tile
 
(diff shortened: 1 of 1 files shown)


shinzoxD (approved):
Verified at c4a4473. The seven added tests map cleanly to the DPS level guard, ImportError/general-exception abstention paths (including the inner-cone call), and both eigenvalue fallback outcomes without overlapping #1931. Locally, the 7-test selection passed; the complete affected module passed with 146 passed, 6 skipped, and 2 xfailed; and the diff check is clean. All seven required CI checks are green.

vprusso (approved):
Reproduced the coverage claim exactly: **94.79%, 17 uncovered lines**, with 1354-1359 and 1368-1372 gone from the missing list.

The choice of target is the part I want to call out. Both helpers *degrade rather than propagate*, so an untested `except` branch there is the worst kind of gap: a regression turning an abstain into a confident verdict returns a wrong separability answer silently, with no exception to notice. Asserting `is None` rather than `is False` is exactly the right assertion, since it pins "cannot answer" as distinct from "not separable".

Mocking `has_symmetric_extension` to raise is legitimate here; these are error paths that cannot be reached by choosing an input, only by making the dependency fail.

PR #1934, opened 2026-08-11
docs(state_opt): document the cvxopt failure mode for PPT distinguishability

idhantsingh027:
`state_exclusion` warns that cvxopt can fail with a `ZeroDivisionError` or an
`ArithmeticError` on PPT problems, and points at `abs_ipm_opt_tol` or
`cvxopt_kktsolver="ldl"`, with a link to picos#341. `state_distinguishability`
has the same failure mode, carries no such note, and has no PPT example at all,
so a user who hits it has nothing in the docs to go on:

    state_distinguishability(
        [bell(0), bell(1), bell(2)],
        measurement="ppt", subsystems=[0], dimensions=[2, 2],
        primal_dual="primal",
    )
    # ArithmeticError: 7

Adds the same note, worded as in `state_exclusion`, and a PPT example passing
`cvxopt_kktsolver="ldl"`, which returns 0.67 on the same input. The example is a
static block rather than an executed one, matching the PPT example in
`state_exclusion` -- the only non-executed block in either module, presumably
for exactly this reason.

`ldl` clears every PPT instance tried here with the default iteration budget,
including ones a `max_iterations` cap does not rescue. Note the option only
reaches the solver at all because of #1923; before that the PPT helpers
discarded solver kwargs, so the documented remedy had no effect.

Docstring only; no behaviour change.

## Changes

  -  [ ] Change 1

## Checklist

### DIFF
--- a/toqito/state_opt/state_distinguishability.py
+++ b/toqito/state_opt/state_distinguishability.py
@@ -173,6 +173,34 @@ def state_distinguishability(
         print(np.around(res, decimals=2))
         ```
 
+        PPT-constrained minimum-error state distinguishability.
+
+        ```python
+        import numpy as np
+        from toqito.state_opt import state_distinguishability
+        from toqito.states import bell
+
+        states = [bell(0), bell(1), bell(2)]
+
+        res, _ = state_distinguishability(
+            states,
+            measurement="ppt",
+            subsystems=[0],
+            dimensions=[2, 2],
+            primal_dual="primal",
+            cvxopt_kktsolver="ldl",
+        )
+
+        print(np.around(res, decimals=2))
+        ```
+
+        !!! Note
+            If you encounter a `ZeroDivisionError` or an `ArithmeticError` when using cvxopt as a solver (which is the
+            default), you might want to set the `abs_ipm_opt_tol` option to a lower value (the default being `1e-8`) or
+            to set the `cvxopt_kktsolver` option to `ldl`.
+
+            See https://gitlab.com/picos-api/picos/-/issues/341
+
     """
     if measurement not in {"positive", "ppt"}:
         raise ValueError("Argument `measurement` must be either 'positive' or 'ppt'.")


toqito/state_opt/state_distinguishability.py
```diff
@@ -173,6 +173,34 @@ def state_distinguishability(
...
+            primal_dual="primal",
+            cvxopt_kktsolver="ldl",
+        )
+
+        print(np.around(res, decimals=2))
+        ```
+
+        !!! Note
+            If you encounter a `ZeroDivisionError` or an `ArithmeticError` when using cvxopt as a solver (which is the
+            default), you might want to set the `abs_ipm_opt_tol` option to a lower value (the default being `1e-8`) or
```
shinzoxD, line 199: `abs_ipm_opt_tol` is a tolerance, and the working change in the linked PICOS discussion is from the `1e-8` default to `1e-7`?a numerically **larger**, looser tolerance. Saying ?set ... to a lower value? directs a user toward `1e-9`, the opposite adjustment. Please say to relax/increase the tolerance (for example, `abs_ipm_opt_tol=1e-7`). I verified that exact option and the `cvxopt_kktsolver="ldl"` path both return `0.666667` with the locked PICOS 2.6.2/CVXOPT 1.3.3 environment. The existing `state_exclusion` note has the same wording, so ideally correct both copies while this guidance is being standardized.

shinzoxD (changes requested):
The example and both solver workarounds are functional, but the tolerance guidance points users in the wrong numerical direction.

vprusso (approved):
Verified both halves of the claim on this branch:

```
without workaround: ArithmeticError: 7
with kktsolver=ldl: 0.67
```

So the failure is real, and the documented remedy genuinely fixes it and returns exactly the 0.67 the example shows.

This is worth having for a reason beyond symmetry with `state_exclusion`: I hit this exact `ArithmeticError: 7` myself while reviewing #1923, and it cost time precisely because `state_distinguishability` carried no note about it. Someone will hit it again.

Keeping the block non-executed, matching the PPT example in `state_exclusion`, is also right; making the docs build depend on a solver path that is known to break on some platforms would be trading a documentation gap for a flaky build.

PR #1936, opened 2026-08-14
test: cover eig-failure and iteration fallback branches in is_separable

sdoygb:
## Summary
Continues #1915 by covering the remaining fallback / iteration branches in ``toqito/state_props/is_separable.py`` that no test reached:

- Vidal-Tarrach (rank-1 perturbation) check skipped when both eigen solvers fail (main-flow path).
- ``_filter_normal_form`` bailing when the post-filter B marginal is rank deficient, and returning the normal form when the iteration budget is exhausted.
- ``_iterative_product_state_subtraction`` stuck detection, backoff-loop exhaustion, and outer-budget exhaustion.
- ``_qubit_qudit_ppt_criteria`` eigvalsh -> eigvals fallback in the swapped 2xN branch, and the Johnston Lemma-1 abstain path when the block eigendecomposition fails.

## Coverage
Module line coverage for ``is_separable.py`` under the test module: 95.5% -> 97.5% (missed lines 14 -> 6, uncovered branches 18 -> 12).

The remaining misses are either defensive checks that are unreachable through the public API (e.g. the zero-dim guards, and the ``1 - t + t**2 < eps`` guard, whose minimum is 0.75) or branches that need bespoke PPT states to trigger (reduction violation, Johnston/Hildebrand boundary, Ha-Kye witness); those are out of scope here.

## Checks
- ``pytest toqito/state_props/tests/test_is_separable.py``: 154 passed, 5 skipped, 2 xfailed
- ``ruff check`` and ``ruff format --check`` (v0.15.0) clean

### DIFF
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -58,6 +58,25 @@ def _eigvalsh_fail_outside_psd(error_msg="mocked eigvalsh fail"):
     return _patched
 
 
+def _eigvalsh_fail_outside_psd_and_iter_sub(error_msg="mocked eigvalsh fail"):
+    """Fail ``eigvalsh`` everywhere except PSD checks and iterative subtraction.
+
+    Variant of ``_eigvalsh_fail_outside_psd`` that additionally passes through
+    ``_iterative_product_state_subtraction``'s direct ``eigvalsh`` calls (its
+    backoff loop requires genuine eigenvalues to run).
+    """
+
+    def _patched(*args, **kwargs):
+        if inspect.currentframe().f_back.f_code.co_name in {
+            "is_positive_semidefinite",
+            "_iterative_product_state_subtraction",
+        }:
+            return _REAL_EIGVALSH(*args, **kwargs)
+        raise np.linalg.LinAlgError(error_msg)
+
+    return _patched
+
+
 # --- Parameterized Tests for Invalid Inputs ---
 """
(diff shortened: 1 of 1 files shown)


vprusso:
Verified by execution rather than from the description.

Reviewed against current master rather than the branch base, since the branch is three commits behind and one of those (#1938) changed `is_positive_semidefinite` internals that the `eigvalsh` mocks here select on by frame name. No interaction: 154 passed, 5 skipped, 2 xfailed.

The coverage numbers are exactly right. I measured 95.52% to 97.48%, missed lines 14 to 6, partial branches 18 to 12, matching the claim.

Coverage alone only shows a line executes, so I also mutation-tested each of the eight new tests: broke the branch it targets in `is_separable.py` and checked the test fails. All eight were killed, so each one pins its branch rather than just walking through it. Concretely, the mutations were `if sorted_eigs_desc is not None` forced true (L591), the `t_b is None` bail (L904), the marginal-error acceptance test (L915), the stuck-detection and backoff-exhausted and budget-exhausted returns (L1044/L1065/L1067), the `eigvals` fallback (L1221), and the Lemma 1 abstain (L1260).

One note for future work, not a change request here: the frame-inspection mocking (`inspect.currentframe().f_back.f_code.co_name`) silently changes which calls get intercepted if a function is renamed or a call moves between frames. That pattern predates this PR, so it is not yours to fix, but it is the thing most likely to make these tests quietly stop testing what they name.

Thanks, this is careful work.

PR #1943, opened 2026-08-19
test: cover final inconclusive return of is_separable with a real state

sdoygb:
## Summary

Replaces the placeholder `test_final_return_false_for_unclassified_entangled` (currently a bare `pass` behind a skip marker) with a genuine end-to-end case that reaches the final inconclusive `(False, ...)` verdict of `is_separable`.

Closes the skipped portion of #1915.

## The state

\rho = 0.85 * rho_tiles + 0.15 * I/9, where rho_tiles = (I - sum|psi_i><psi_i|)/4 is the complement of the 3x3 Tile UPB (a rank-4 PPT entangled state, built with `toqito.states.tile` — no external fixtures).

## Why this state reaches the final verdict

- PSD and PPT (PT smallest eigenvalue ≈ +1.7e-2, far from the tolerance boundary);
- passes the realignment/CCNR criterion (||R(rho)||_1 ≈ 0.969 ≤ 1);
- not detected by any implemented positive-map witness (Choi/Terhal/Ha-Kye/Breuer-Hall);
- not decomposed by iterative product-state subtraction (the separable-decomposition boundary of this mixture lies between t=0.78 and t=0.80, so t=0.85 has a ~9% margin);
- the level-2 DPS hierarchy finds a 2-symmetric extension but fails the inner cone, so the finite hierarchy is inconclusive.

## Stability

- The state is deterministic (fixed construction from `tile()`), with margins: 3.1% on CCNR (deterministic SVD), 9% on iterative subtraction (fixed-seed random restarts);
- verified across 3 independent runs (different process seeds);
- full `test_is_separable.py` suite: 155 passed / 4 skipped / 2 xfailed (49.7s), unchanged baseline except this test now passes.

## Notes

- First real solver-backed (CVXPY/SCS) end-to-end coverage of the 12b/DPS fall-through path — previously only mocked.
- Test runtime ~9s (DPS SDP solve), acceptable for this file.

### DIFF
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -551,10 +551,7 @@ def test_final_return_false_for_unclassified_entangled():
 
     verdict, reason = is_separable(rho, dim=[3, 3], level=2)
     assert verdict is False
-    assert reason == (
-        "inconclusive: PPT but no implemented sufficient condition "
-        "proved separability"
-    )
+    assert reason == ("inconclusive: PPT but no implemented sufficient condition proved separability")
 
 
 def test_2xN_johnston_lemma1_eig_A_fails():
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -530,10 +530,31 @@ def test_2xN_johnston_spectrum_lam_too_short_skips_proceeds(mock_eig):
     assert is_separable(rho_2x4, dim=[2, 4])[0]
 
 
-@pytest.mark.skip(reason="Needs specific state that evades criteria but is known entangled for `final return False`.")
 def test_final_return_false_for_unclassified_entangled():
-    """Placeholder for final False return on unclassified entangled state."""
-    pass
+    """A PPT state passing every implemented criterion reaches the final False.
+
+    The 3x3 Tile UPB complement (a rank-4 PPT entangled state) mixed with the
+    maximally mixed state at ratio 0.85 is positive semidefinite and PPT, but
+    evades each implemented sufficient condition: the realignment/CCNR norm
+    stays below one, the positive-map witnesses (Choi, Terhal, Ha-Kye,
+    Breuer-Hall) do not fire, and iterative product-state subtraction stalls.
+    The level-2 DPS hierarchy finds a 2-symmetric extension yet fails the
+    inner cone, so the finite hierarchy is inconclusive and the function
+    falls through to the final ``(False, ...)`` verdict.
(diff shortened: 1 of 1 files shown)


sdoygb:
Pre-verified this test against the upcoming SCS→CLARABEL migration (#1942). Findings:

**Current master (inner cone hardcodes SCS):** test exercises the normal DPS path — outer `has_symmetric_extension` (AUTO→SCS) returns True, inner `has_symmetric_inner_extension` (SCS) returns False, final `(False, inconclusive)` fires. Stable across 3 independent runs.

**If the inner cone is switched to CLARABEL:**
1. Parameter mismatch: `has_symmetric_inner_extension` passes `eps_abs`/`eps_rel`/`max_iters`, which CLARABEL rejects (`unrecognized solver setting`) — would need `tol_gap_abs`/`tol_gap_rel`/`max_iter`.
2. Even with corrected parameters, CLARABEL terminates with **NumericalError** (solve time ~1.0s) on this SDP, raising SolverError.
3. The test stays green either way (the exception is caught in `_dps_hierarchy_criterion` → returns None → final False), but it would silently stop covering the *normal* DPS path and start covering the exception fallback instead.

**Outer check:** with explicit CLARABEL, `has_symmetric_extension` still returns True (consistent with SCS), so only the inner cone needs attention during migration.

Repro (toqito clone @ 5135ccb, cvxpy 1.9.2): swap `solver=cvxpy.SCS` → `solver=cvxpy.CLARABEL` in `has_symmetric_inner_extension` and fix the three parameter names above; run this test on `rho = 0.85*rho_tiles + 0.15*I/9`.

vprusso:
All six pytest jobs pass; the only red check is code-style, and the cause is one hunk. The formatter wants the implicit string concatenation in the new inconclusive-reason assert collapsed to a single line:

```diff
-    assert reason == (
-        "inconclusive: PPT but no implemented sufficient condition "
-        "proved separability"
-    )
+    assert reason == ("inconclusive: PPT but no implemented sufficient condition proved separability")
```

Running `ruff format toqito/state_props/tests/test_is_separable.py` and pushing clears it. I checked the file under both ruff 0.15.0 (what the CI action resolved) and 0.16.0 (the dev pin); they agree, so this is not a version-drift artifact, and current master is clean under both.

Since #1944 and #1945 are stacked on this branch, they inherit the same hunk and fail for the same single reason; fixing it here and rebasing the other two, or running ruff format on each branch, clears all three. I applied the stack tip to current master and ran the file: 158 passed, 3 skipped, 1 xfailed. Planned merge order once green is #1943, #1944, #1945.

PR #1944, opened 2026-08-19
test: cover Ha-Kye positive-map witness firing with a real state

sdoygb:
## Summary

Replaces the stale, skipped test `test_path_ha_kye_fallthrough_to_final_false_L534_to_L580` (skipped with \"mock chain outdated, needs rewrite\") with deterministic, real coverage of the Ha-Kye positive-map witness family in `_positive_map_witness_criteria`.

## Why the old test could never work

The old test mocked the Plücker-determinant branch and assumed the 3x3 Tile UPB complement passes through Choi/Terhal/Ha-Kye to reach the final `False`. That premise is wrong: the Tile state is actually caught by the **Terhal 2000 tile witness** (`tr(Hρ) ≈ -0.0036 < 0`) at step 12, before Ha-Kye is ever reached. The mock chain was stale not just because of the Plücker branch rename, but because the intended path was never real.

## What the new test does

A fixed local-unitary orbit `U × V` of the Tile UPB complement (unitaries hard-coded at 17-digit precision, no RNG dependence):

- keeps the state PPT and its entanglement class invariant (local unitaries commute with partial transpose),
- rotates the computational basis so that the Ha-Kye map with `t = 10` fires: `(phi_t ⊗ I)(rho)` fails to be PSD,
- while Choi 1975 (both subsystems) and Terhal 2000 still pass.

`_positive_map_witness_criteria` now has real coverage of the Ha-Kye loop's `False` branch (18-parameter sweep), returning `(False, "Ha-Kye positive-map witness (3x3, t=10)")`.

## Verification

- `is_ppt(rho, [3, 3])` passes (asserted inside the test)
- new test runtime ≈ 1.9 s (no DPS solver dependency)
- full test file: 156 passed / 3 skipped / 2 xfailed

## Honest note (search evidence)

The Ha-Kye family does **not** fire on standard parameter families in the computational basis: Horodecki 3x3 (full `a` sweep), chessboard (2000 random parameter sets), SHIFTS UPB complement, and 572 random PPT states all abstain. Out of 2400 random local-unitary orbits, exactly this one fires (`t=10`). The test's docstring documents this limitation (\"evades every member of the Ha-Kye family in the computational basis\").

Related: #1943 (real-state coverage of the final inconclusive return), same #1915.

### DIFF
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -530,10 +530,31 @@ def test_2xN_johnston_spectrum_lam_too_short_skips_proceeds(mock_eig):
     assert is_separable(rho_2x4, dim=[2, 4])[0]
 
 
-@pytest.mark.skip(reason="Needs specific state that evades criteria but is known entangled for `final return False`.")
 def test_final_return_false_for_unclassified_entangled():
-    """Placeholder for final False return on unclassified entangled state."""
-    pass
+    """A PPT state passing every implemented criterion reaches the final False.
+
+    The 3x3 Tile UPB complement (a rank-4 PPT entangled state) mixed with the
+    maximally mixed state at ratio 0.85 is positive semidefinite and PPT, but
+    evades each implemented sufficient condition: the realignment/CCNR norm
+    stays below one, the positive-map witnesses (Choi, Terhal, Ha-Kye,
+    Breuer-Hall) do not fire, and iterative product-state subtraction stalls.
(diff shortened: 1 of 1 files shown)


vprusso:
Same single code-style failure as #1943, inherited through the stacked branch; see the diagnosis there. The tests themselves pass when this branch is applied to current master. Running `ruff format toqito/state_props/tests/test_is_separable.py` on this branch (or rebasing after #1943 is fixed) clears it.

PR #1945, opened 2026-08-19
test: cover Breuer-Hall witness and 2x4 DPS path with real states

sdoygb:
## Summary

Replaces the stale `xfail` mock test `test_breuer_hall_on_dB_only_mocked_first_xfail` (which could never pass, see below) with two real, deterministic tests covering previously unexecuted branches of the 2x4/4x4 path in `is_separable`.

## Why the old test could never work

Two independent defects:

1. **Wrong premise**: the 2x4 Horodecki state (`a=0.5`) is *not* detected by the Breuer-Hall witness — `_positive_map_witness_criteria` returns `None` for it (all Breuer-Hall subsystem checks pass PSD). The test assumed Breuer-Hall would fire on it.
2. **Broken mock condition**: the side-effect checked `sys == 0`, but the implementation iterates `(1, 2)` (toqito's 1-indexed subsystem convention), so the mock could never be hit and the second assertion always failed.

## What the new tests do

**`test_2x4_horodecki_ppt_entangled_caught_by_dps`** (end-to-end): the 2x4 Horodecki state passes PPT, rank, CCNR/realignment and the full positive-map witness stage (including both Breuer-Hall subsystem directions — real execution of the Breuer-Hall pass-through path) and is first detected by the DPS hierarchy at level 2. This is also the first real (non-mocked) DPS coverage outside the 3x3 domain.

**`test_breuer_hall_witness_fires_on_breuer_state`** (helper-level): the Breuer state `rho = (P_phi + P_psi)/2` with `|psi> = (I x U)|phi>` and `U` the 4x4 anti-symmetric unitary fires the Breuer-Hall witness on subsystem 2 (`dim=4`) — the first real execution of the Breuer-Hall `False` branch. The docstring honestly notes the state is NPT (so the full flow catches it at the PPT criterion first); the test targets the witness branch directly.

**Werner test**: the skip reason was wrong ("unexpectedly not PPT"). The correct fact: 2x2 Werner states can never reach Breuer-Hall — `alpha > 0` is NPT (caught by the PPT criterion first) and `alpha < 0` is PPT but separable (2x2 PPT iff separable). Updated reason + docstring.

## Verification

- `158 passed, 3 skipped, 1 xfailed` on the full file (was 156/3/2; the stale xfail is gone)
- New tests are deterministic (no RNG, no mocks): Horodecki construction + fixed `U`

Related: #1943, #1944 (same #1915 coverage work).

### DIFF
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -753,7 +753,8 @@ def test_breuer_hall_on_dA_detects_entangled_2x2werner():
     states; this test documents that fact.
     """
     pytest.skip(
-        "2x2 Werner states cannot reach Breuer-Hall: NPT states are caught by the PPT criterion first, PPT states are separable."
+        "2x2 Werner states cannot reach Breuer-Hall: NPT states are caught by the "
+        "PPT criterion first, PPT states are separable."
     )
     # assert not is_separable(rho_w_ent_2x2, dim=[2, 2], tol=1e-8)[0]
 
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -709,37 +709,55 @@ def test_2xN_hildebrand_homothetic_true_v3():
     pass
 
 
-@pytest.mark.xfail(reason="Breuer-Hall mock test for 2x4 Horodecki state not fully supported.")
-@mock.patch("toqito.state_props.is_separable.partial_channel")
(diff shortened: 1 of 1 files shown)


vprusso:
Same single code-style failure as #1943, inherited through the stacked branch; see the diagnosis there. The tests themselves pass when this branch is applied to current master. Running `ruff format toqito/state_props/tests/test_is_separable.py` on this branch (or rebasing after #1943 is fixed) clears it.

PR #1948, opened 2026-08-31
Is separable  horodecki low rank ppt criteria

AVlyx:
## Description
Bug in is_separable where a state with a low rank would be incorrectly detected by _horodecki_low_rank_ppt_criteria as separable.

## Changes
  -  [ ] Brought back the branches where the method would compare the state's rank to the rank of its marginals.
  -  [ ] Removed the branch where the method would check if the state's rank is lower or equal to the dimension of Alice's and Bob's systems instead of the rank of their marginals.
  - [ ] Breur state is bound entangled from 0 to 1/(d+2)
  - [ ] Contribution guide documentation did not build with provided snippet 
  
  ```
uv run mkdocs serve -f docs/mkdocs.yml
INFO    -  DeprecationWarning: warning_filter doesn't do anything since MkDocs 1.2 and will be removed soon. All messages on the `mkdocs` logger get counted automatically.
             File "C:\Users\anton\OneDrive\Desktop\GitHub\toqito\.venv\Lib\site-packages\mkdocs_gallery\mkdocs_compatibility.py", line 12, in <module>
               from mkdocs.utils import warning_filter
             File "C:\Users\anton\OneDrive\Desktop\GitHub\toqito\.venv\Lib\site-packages\mkdocs\utils\__init__.py", line 403, in __getattr__
               warnings.warn(
ERROR   -  Config value 'plugins': Plugin 'gallery' option 'examples_dirs': Error validating config item #1: The path 'content/examples' isn't an existing directory.

Aborted with a configuration error!
```

## Checklist

### DIFF
--- a/docs/content/contributing-guide.md
+++ b/docs/content/contributing-guide.md
@@ -250,7 +250,8 @@ serve locally:
 
 ``` bash
 uv sync --group docs
-uv run mkdocs serve -f docs/mkdocs.yml
+cd docs
+uv run mkdocs serve -f mkdocs.yml
 ```
 
 This starts a live-reloading server at `http://127.0.0.1:8000`.
@@ -258,7 +259,8 @@ This starts a live-reloading server at `http://127.0.0.1:8000`.
 To produce a static build instead:
 
 ``` bash
-uv run mkdocs build -f docs/mkdocs.yml
+cd docs
+uv run mkdocs build -f mkdocs.yml
 ```
 
 For more information, visit the
--- a/toqito/state_props/is_separable.py
+++ b/toqito/state_props/is_separable.py
@@ -90,8 +90,8 @@ def is_separable(
 
         For PPT states (especially when \(d_A d_B > 6\)):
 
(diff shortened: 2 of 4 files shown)


vprusso:
Reviewed and verified this end to end: the marginal-rank fix is correct, and I confirmed it resolves the false-positive (a tiles bound-entangled state embedded in 4x4, which the old `rank <= max(d_A, d_B)` branch wrongly reported separable, now correctly falls through to the CCNR criterion and is reported entangled). Merges cleanly onto current master, and the full `is_separable` / `breuer` suites pass.

I pushed one small commit to your branch (maintainer edit) fixing two docstrings that otherwise contradicted the change: the `is_separable` module docstring still advertised the removed `rank <= max(d_A, d_B)` rule as "Theorem 1 of the paper", and the `breuer` Examples paragraph still said "any positive weight" against the corrected `0 < lambda <= 1/(d+2)` range on the line above. Merging once CI is green. Thanks for the fix.

PR #1961, opened 2026-09-04
perf: compute the spectrum once in renyi_entropy

### LINKED ISSUE #1958: perf: renyi_entropy computes the spectrum multiple times
### Summary
`state_props/renyi_entropy.py` factors `rho` more than once. Line 95 calls `is_density(rho)`, which factors `rho` via `is_positive_semidefinite` -> `eigvalsh`; line 104 then calls `np.linalg.eigvalsh(rho)` again. The `alpha == 1` branch (line 102) is worse: it delegates to `von_neumann_entropy`, which re-validates and re-factors a third time.

This is the same pattern recently fixed in `von_neumann_entropy`: compute the spectrum once and validate from it.

### Fix
Compute `eigs = np.linalg.eigvalsh(rho)` once, validate the density-matrix conditions from the spectrum plus `np.isclose(np.trace(rho), 1)`, and drop the `is_density` import. Mirror the `von_neumann_entropy` refactor.

Good first issue.

vprusso:
Closes: #1958

`renyi_entropy` factored `rho` more than once. The `alpha == 1` branch delegated to `von_neumann_entropy`, which re-runs `is_density` (already called at the top of `renyi_entropy`) and re-factors `rho` a second time via `eigvalsh`.

This computes the eigenvalues once and reads every spectrum-dependent order off them, computing the von Neumann case (`alpha == 1`) inline. The `is_density` validation and the `matrix_rank` path for `alpha == 0` are left unchanged, so behaviour is identical; the inline von Neumann value was checked bit-identical to `von_neumann_entropy` across random density matrices.

### DIFF
--- a/toqito/state_props/renyi_entropy.py
+++ b/toqito/state_props/renyi_entropy.py
@@ -3,7 +3,6 @@
 import numpy as np
 
 from toqito.matrix_props import is_density
-from toqito.state_props import von_neumann_entropy
 
 
 def renyi_entropy(rho: np.ndarray, alpha: float) -> float:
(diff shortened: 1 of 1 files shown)


vprusso:
Superseded by #1966 (thanks @Gambit-Checkmate), which implements the same single-spectrum fix for the same issue (#1958) and goes a step further by validating positive-semidefiniteness directly from the computed eigenvalues. Verified its behaviour is identical to the prior implementation across the numeric orders and all the density-matrix rejection cases, and its tests pass. Closing this one in its favor.

Issue #1957, opened 2026-09-04
robustness: consistent problem.status checks across SDP solve sites


vprusso:
### Summary
SDP solve sites are inconsistent about checking `problem.status` before returning the optimal value. Some check (`has_symmetric_extension.py:170`, `matrix_props/sk_norm.py:229,275`, `state_props/is_k_incoherent.py`), others return `.value` / `problem.solve()` directly (`state_opt/state_distinguishability.py`, `state_opt/state_exclusion.py`, `state_opt/optimal_clone.py`, `channel_metrics/completely_bounded_trace_norm.py`, both `fidelity_of_separability` functions).

### Assessment
No confirmed silent-wrong-number path exists: PICOS raises `SolutionFailure` on a non-optimal solve, and cvxpy returns `±inf` for infeasible/unbounded. So this is defensive consistency, not a live bug. But the inconsistency is real, and a status check makes a solver failure surface as a clear error rather than a downstream surprise.

### Fix
Add a `problem.status in {optimal, optimal_inaccurate}` (or PICOS equivalent) guard to the solve sites that lack one, matching the functions that already have it.

Enhancement / robustness.

PR #1967, opened 2026-09-04
cleanup: fix symmetric extension docs and dead branches

### LINKED ISSUE #1959: cleanup: fix has_symmetric_extension docstring and delete pragma-marked dead branches
### Summary
Two small cleanup items from an audit.

### 1. `has_symmetric_extension` docstring shows a wrong formula
The docstring example (`state_props/has_symmetric_extension.py:62-63`) writes `np.trace(rho**2)` (elementwise square) where the criterion needs the matrix square; the library code (line 139) correctly uses `np.linalg.matrix_power(rho, 2)`. The displayed inequality still happens to evaluate `True`, so it is not visibly broken, but the shown formula misrepresents the criterion. (The example's `rho` also has trace 4, i.e. is not a normalized density matrix.) Fix the docstring to match the code.

### 2. Delete `# pragma: no cover` dead branches
Several defensive branches are provably unreachable and already marked `# pragma: no cover`, so they can be deleted rather than carried:
- `states/mutually_unbiased_basis.py:99` — `if not _is_prime(p): continue` inside `_is_prime_power`, where `p` is the smallest divisor of `n` and the smallest divisor of any integer is always prime. Deleting it also removes a wasted primality test per factorization. (`_is_prime` stays in use at lines 56, 62.)
- `matrix_props/sk_norm.py:229,275,308` and `state_props/sandwiched_renyi_conditional_entropy.py:208,210,216,223,239` — guards for conditions an earlier validation already excludes.

Good first issue.

Gambit-Checkmate:
## Summary
- correct and normalize the two-qubit symmetric-extension docstring example
- remove pragma-marked unreachable branches in mutually unbiased bases and S(k)-norm calculations
- remove unreachable defensive branches in the uparrow sandwiched conditional Rényi entropy path

Closes #1959

## Validation
- `uv run pytest toqito/state_props/tests/test_has_symmetric_extension.py toqito/states/tests/test_mutually_unbiased_basis.py toqito/matrix_props/tests/test_sk_norm.py toqito/state_props/tests/test_sandwiched_renyi_conditional_entropy.py -q`
- `uv run --group lint ruff check` for the three changed files without existing baseline diagnostics
- `uv run --group lint ruff format --check` for all changed files

### DIFF
--- a/toqito/matrix_props/sk_norm.py
+++ b/toqito/matrix_props/sk_norm.py
@@ -226,8 +226,6 @@ def sk_operator_norm(
             warnings.filterwarnings("ignore", message="Constraint.*subexpressions")
             problem = cvxpy.Problem(objective, constraints)
             cvx_optval = problem.solve()
-        if problem.status != "optimal":  # pragma: no cover - defensive
-            raise ValueError("Numerical problems encountered.")
 
         upper_bound = min(upper_bound, np.real(cvx_optval))
 
@@ -272,8 +270,6 @@ def sk_operator_norm(
                     warnings.filterwarnings("ignore", message="Constraint.*subexpressions")
                     problem = cvxpy.Problem(objective, constraints)
                     cvx_optval = problem.solve()
-                if problem.status != "optimal":  # pragma: no cover - defensive
-                    raise ValueError("Numerical problems encountered.")
 
                 upper_bound = min(upper_bound, np.real(cvx_optval))
 
@@ -305,9 +301,6 @@ def __lower_bound_sk_norm_randomized(
     start_vec: np.ndarray | None = None,
 ) -> float:
     """Compute a lower bound of the S(k)-norm via a randomized method."""
-    if mat.shape[0] != mat.shape[1]:  # pragma: no cover - defensive; caller validates shape
-        raise ValueError("Input matrix must be square.")
-
     dim_a, dim_b = dim
 
     psi = max_entangled(k, is_normalized=False)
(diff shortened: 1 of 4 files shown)


vprusso:
Thanks for this. The two safe parts of it have now landed in #1963: the `has_symmetric_extension` docstring formula fix (`rho**2` -> `matrix_power(rho, 2)`) and the removal of the unreachable prime-power guard in `mutually_unbiased_basis`. So this branch will now conflict on those files.

The unique remaining content here is deleting the `# pragma: no cover` guards in `sk_norm` and `sandwiched_renyi_conditional_entropy`, and I'd push back on those:

- The `sk_norm` guards are `if problem.status != "optimal": raise ...`. Those are solver-status checks, exactly the kind of defensive guard that #1957 tracks for *wider* adoption, not removal. Deleting them lets a non-optimal SDP solve proceed with a garbage value.
- The `sandwiched_renyi` guards (`support_overlap <= 0`, `max_eig <= 0`, the finiteness/convergence check) are numerical safety nets on a non-convex optimizer. Removing the `max_eig <= 0` branch in particular changes behaviour on degenerate spectra, and dropping the final convergence check turns a failed optimization into a silently-returned bad value.

These being uncovered by tests is expected (they fire only on solver/numerical failure); it does not make them dead. I'd suggest closing this one, since its mergeable content is already in via #1963. If you'd like to pursue the `sandwiched_renyi` numerics, the `max_eig <= 0` path would be the thing to examine on its own merits rather than delete. Appreciate the contribution.

PR #1969, opened 2026-09-05
test: cover reduction-criterion propagation and three abstain guards in is_separable

09Catho:
## Description

Continues #1915. Adds four tests that reach lines and branches in `is_separable.py` that the test suite wasn't hitting yet, without touching `is_separable.py` itself.

## Changes

- Covers the reduction-criterion propagation line in `is_separable` by mocking `_reduction_ppt_criterion` on a state that already reaches section 8 (`_RHO_ENT_SYMM_3x3_REACHES_12B`, used by an existing test above it). A mock is needed here rather than a real violating state, because the comment right above that call notes that PPT already implies the reduction criterion holds, so no real PPT state can violate it once execution gets that far.
- Covers `_vidal_tarrach_ppt_criterion`'s guard-false branch directly, once with a mismatched spectrum length and once with `prod_dim_val <= 1`.
- Covers `_qubit_qudit_ppt_criteria`'s Johnston length guard and its empty-block guard together, by calling the helper directly with a zero-sized state and a nonzero `prod_dim_val`.
- Covers `_range_projector_product_overlap_3x3_rank4`'s loop-exhaustion fallback by forcing `max_iter=1` on a real rank-4 3x3 state.

Coverage of `toqito/state_props/is_separable.py` moves from 97.76% (475 statements, 5 missing, 240 branches, 11 partial) to 98.60% (475 statements, 4 missing, 240 branches, 6 partial).

The remaining misses look unreachable through any input without editing `is_separable.py`, which the issue asks to leave alone here, so I didn't attempt them:

- Lines 469, 479, and 481 guard internal-consistency invariants. Every branch of the `dim` normalization above them already guarantees a nonzero `dA` and `dB` whenever `state_len != 0`, so these `raise`s can't fire.
- Line 1290 guards `denom_ha` against being near zero, but `denom_ha = 1 - t + t^2` has a minimum of 0.75 over the real `t` values the loop iterates through, so it's the same situation.
- The `741->752` branch is inside `_terhal_2000_tile_witness`, which runs once at import time with a fixed seed and no caller-adjustable parameters, so its convergence loop can't be forced to run out from a test.
- The `1256->1262` branch checks `eig_a_real.size > 0` right after the enclosing `if a_block.size > 0 and b_block.size > 0 and c_block.size > 0:`. Since all three blocks share the same square dimension, a nonzero block size always means a nonzero eigenvalue count, so that inner guard can't fail on its own.

## Verification

```
uv run pytest toqito/state_props/tests/test_is_separable.py -q
162 passed, 3 skipped, 1 xfailed

uv run ruff check toqito/state_props/tests/test_is_separable.py
All checks passed!
```

## Checklist

- [x] Check the documentation build does not lead to any failures. (test-only change, no doc-facing symbols touched)

### DIFF
--- a/toqito/state_props/tests/test_is_separable.py
+++ b/toqito/state_props/tests/test_is_separable.py
@@ -26,6 +26,7 @@ from toqito.state_props.is_separable import (
     _range_projector_product_overlap_3x3_rank4,
     _reduction_ppt_criterion,
     _sorted_real_eigs_desc,
+    _vidal_tarrach_ppt_criterion,
 )
 from toqito.states import basis, bell, horodecki, isotropic, max_entangled, tile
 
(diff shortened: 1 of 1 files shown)


vprusso:
Verified: all four new tests fail when the branch they target is mutated, so they're real coverage, not just line-hits. Green, tests-only, doesn't touch is_separable.py. The unreachable-line reasoning matches what I found auditing this file earlier. Merging, thanks.

Issue #1955, opened 2026-09-04
refactor: de-duplicate min-error SDP builders and Gauss-Legendre quadrature


vprusso:
### Summary
Several near-identical SDP builders and one numerics helper are duplicated across modules.

### Duplicates
1. **Gauss-Legendre on [0,1]**, three copies: `state_metrics/measured_relative_entropy.py:139` (`_gauss_legendre_on_01`), `channel_metrics/channel_measured_relative_entropy.py:148` (`_gauss_legendre_on_01`, byte-identical), and `cones/operator_relative_entropy_epi_cone.py:150` (`_gauss_legendre`, same object via Golub-Welsch). Consolidate into one shared helper.
2. **Min-error state-discrimination SDP**: `state_opt/state_distinguishability.py:286,309` (`_min_error_primal`/`_min_error_dual`) vs `state_opt/state_exclusion.py:384,406` (same names). Near-identical PICOS programs differing only in `max`/`min` sense and the dual constraint direction.
3. **Min-error channel-discrimination SDP**: `channel_opt/channel_distinguishability.py:264,299` vs `channel_opt/channel_exclusion.py:214,245`. Same Choi-based SDP, differing only in objective sense and one constraint direction.
4. **Collins-Gisin index maps** duplicated within one file: `state_opt/bell_inequality_max.py:525-529` and `:1127-1139` define byte-identical `_cg_row_index`/`_cg_col_index`. Lift to module-level helpers.

### Fix
For 2 and 3, a single parameterized builder taking the objective sense (min vs max) and the dual direction. For 1, one shared `_gauss_legendre_01(m)`. For 4, two module-level helpers.

Refactor. Behaviour must be identical, so run the affected `state_opt` / `channel_opt` / `*_relative_entropy` tests.

vprusso:
Assessment before implementing: this is a real maintainability item, but the pieces carry more risk to correctness-critical code than the duplication costs, so it is worth doing carefully as a standalone effort rather than a quick refactor. Notes for whoever picks it up:

[...]

PR #1970, opened 2026-09-05
refactor: de-duplicate SDP builders and Gauss-Legendre quadrature

mukktinaadh:
## Summary

Addresses #1955 by de-duplicating two families of copied implementation:

1. **Gauss-Legendre quadrature** — three byte-for-byte copies of the same `[0,1]`
   quadrature rule (`_gauss_legendre_on_01` x2 in `state_metrics`/`channel_metrics`,
   and the Golub-Welsch variant `_gauss_legendre` in `cones`) now live in a single
   shared helper, `toqito._quadrature._gauss_legendre_01`. The cone's public-facing
   `_gauss_legendre` wrapper is kept (its tests exercise it) and now delegates.

2. **Minimum-error SDP builders** — the state and channel pairs each carried
   identical primal/dual builders for distinguishability and exclusion:
   - `state_distinguishability._min_error_primal/_min_error_dual` gained a
     `sense` argument (`"max"` = distinguishability, `"min"` = exclusion);
     `state_exclusion` now imports them instead of re-defining them.
   - `channel_distinguishability._bayesian_primal/_bayesian_dual` were renamed
     `_min_error_primal/_min_error_dual` and parameterized the same way;
     `channel_exclusion` now imports them.

Net: **-159 / +66 lines** across 8 files (7 modified + 1 new).

## Correctness notes

- The dual SDPs flip constraint direction *and* objective sense together so the
  dual remains the dual of the requested primal sense; return values are shaped
  per-sense (`[Y]` for distinguishability, constraint duals for exclusion) to
  preserve the existing public return contracts.
- The cone's Golub-Welsch `_gauss_legendre` and the `leggauss`-based version
  agree to machine precision (verified numerically; includes the `m == 1` case).

## Verification

- Full test suite: **3736 passed, 60 skipped, 2 xfailed**
- `ruff check` and `ruff format --check` clean
- Numeric equivalence checks of the refactored SDP builders against the
  pre-refactor implementations on random inputs: values agree to ~1e-9 or better
  (solver noise), for both primal and dual, both state and channel, both senses.

### DIFF
--- a/toqito/_quadrature.py
+++ b/toqito/_quadrature.py
@@ -1,0 +1,26 @@
+"""Shared numerical quadrature helpers."""
+
+import numpy as np
+
+
+def _gauss_legendre_01(m: int) -> tuple[np.ndarray, np.ndarray]:
+    r"""Compute m-point Gauss-Legendre quadrature nodes and weights on [0, 1].
+
+    The nodes and weights are obtained from the standard Gauss-Legendre rule on
+    :math:`[-1, 1]` (via ``numpy.polynomial.legendre.leggauss``) and mapped
+    affinely onto :math:`[0, 1]`.
+
+    Args:
+        m: Number of quadrature points. Must be at least 1.
+
+    Returns:
+        Tuple of nodes and weights on the interval :math:`[0, 1]`.
+
+    Raises:
+        ValueError: If ``m`` is less than 1.
+
+    """
+    if m < 1:
+        raise ValueError("m must be at least 1.")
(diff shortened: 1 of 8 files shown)


vprusso:
Verified the risky parts: state_distinguishability and state_exclusion give identical values to master on both the primal and dual paths (16 cases), so unifying the dual with the shared `primals=None` didn't change behavior. Channel builders and the quadrature swap check out too (205 tests pass across the touched modules). Good handling of the sense/constraint flip. Merging, thanks.

